Issue #37, opened 2023-03-01
Add A Code Linter to the Project


GoodGuyGregory:
As the project gets larger we will need to maintain consistency among developers. 

the team has decided to implement a linting tool.

I plan to work through this with the team and post a documentation trail here for future projects as well.

GoodGuyGregory:
This has been handled and the styling preferenced. We are willing to move forward with the code so far, @Jared-Krajewski @leekahung

PR #39, opened 2023-03-02
Adds Linter and Formatter 

GoodGuyGregory:
In Response to #37 this adds Eslint for JS formatting with a blank slate for prettier. This PR includes ES-Lint with the standard airbnb flavor for linting. Please Merge and discuss the rules as Prettier will need to be modified to everyone's liking


leekahung:
@GoodGuyGregory should we merge this into Development first before Master? I'm down with setting up prettier and lints.

GoodGuyGregory:
@leekahung as per standard now that we have both Master and Dev tracking real changes, you're correct first step is development then master for merging. I will be adding a few more files into the project before a merge is appropriate. We can chat more tonight in the Voice channel

PR #41, opened 2023-03-02
Updating README.md to replace build.yml to ci.yml

leekahung:
Noticed the badge broke on Master, so I've checked the README.md and updated it


k-mack (approved):
Thanks for catching that!

PR #46, opened 2023-03-09
#39 - Adds Base Configuration Changes for ESLint

GoodGuyGregory:
@leekahung This expands on what we discussed in the meeting this week. 

It should add base configurations. I recommend we go line by line with this one locally, if all of your changes are up to date in Development, 

I should have some time Friday if you're available on Discord

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,3 @@
+./docs
+./node_modules
+./src/index.html
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,37 +1,30 @@
-import { SessionProvider } from "@inrupt/solid-ui-react";
-import Login from "./components/Login/Login";
-import {
-  UploadDocumentForm,
-  FetchDocumentForm,
-  DeleteDocumentForm,
-} from "./components/Form";
-import CrossPodQuery from "./components/Form/CrossPodQuery";
-import CrossPodWrite from "./components/Form/CrossPodWrite";
+import { SessionProvider } from '@inrupt/solid-ui-react';
+import { React } from 'react';
+import Login from './components/Login/Login';
+import { UploadDocumentForm, FetchDocumentForm, DeleteDocumentForm } from './components/Form';
+import CrossPodQuery from './components/Form/CrossPodQuery';
+import CrossPodWrite from './components/Form/CrossPodWrite';
 
-const AppHeader = () => {
-  return (
-    <header>
-      <h2>Getting Started with PASS</h2>
-    </header>
-  );
-};
+const AppHeader = () => {return (
+	<header>
+		<h2>Getting Started with PASS</h2>
+	</header>
+)};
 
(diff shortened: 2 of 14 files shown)


.prettierignore
```diff
@@ -0,0 +1,3 @@
+./docs
+./node_modules
+./src/index.html
```
leekahung, line 3: I don't think Line 3 is necessary. The base index.html is located on the root directory: ./

src/App.jsx
```diff
@@ -1,37 +1,30 @@
...
+import CrossPodWrite from './components/Form/CrossPodWrite';
 
-const AppHeader = () => {
-  return (
-    <header>
-      <h2>Getting Started with PASS</h2>
-    </header>
-  );
-};
+const AppHeader = () => {return (
```
leekahung, line 8: Can review arrow function block statements for React in Friday's discussion, but essentially, components can be made without explicit returns if it's only return JSX Elements: https://stackoverflow.com/questions/53203257/making-components-with-curly-brackets-versus-parenthesis-when-using-functional-c

There's an example of what's shown in StackOverflow in Development's updated AppHeader without the explicit return.

src/components/Form/CrossPodWrite.jsx
```diff
@@ -1,95 +1,77 @@
-import { useState } from "react";
-import { useSession } from "@inrupt/solid-ui-react";
-import { useField } from "../../hooks";
-import StatusNotification from "./StatusNotification";
-import DocumentSelection from "./DocumentSelection";
+import { useState, React } from 'react';
```
leekahung, line 1: Should be: import React, { useState } from 'react';

src/components/Form/DocumentSelection.jsx
```diff
@@ -1,15 +1,15 @@
...
-        {docTypes.map((doc, index) => {
-          return <option key={index}>{doc}</option>;
-        })}
-      </select>
-    </>
-  );
+	return (
+		<select id={htmlId} name="document">
+			{docTypes.map((doc, index) => {
+				return <option key={index}>{doc}</option>;
```
leekahung, line 8: Found a workaround for key prop. Decided to just use the first word of multiworded docType, which can be obtained using: doc.split(' ')[0]. Can be discussed on Friday.

Check out Development's DocumentSelection.jsx for the workaround.

leekahung (changes requested):
These feedback will likely be discussed on our Friday discussion. Don't see much of a problem with lint rules being presented here. It's very similar to what I've been exploring. Just need to hash things out by line as mentioned on Discord on what rules to add or drop. After that, I think we're good to merge based on the changes to ESLint/Prettier config files and updates from reformatting using Prettier.

.prettierignore
GoodGuyGregory: removing now

PR #54, opened 2023-03-11
Disable Button when application runs

### LINKED ISSUE #53: Disable button while application is running to upload/fetch/querying file form Pod
To prevent extraneous request, it might be good to disable the button for specific form while it's running the request

Include action as part of statusReducer

None

**Additional context**
None

leekahung:
- Correcting JSDoc for statusReducer (which is suppose to be file property, not clearValue)
- Included processing property to useStatusNotificationObject
- Changed reducer to return only action of specific property when called
- Included SET_PROCESSING and CLEAR_PROCESSING to components and runNotification function

### DIFF
--- a/src/components/Form/CrossPodQueryForm.jsx
+++ b/src/components/Form/CrossPodQueryForm.jsx
@@ -18,6 +18,8 @@ const CrossPodQueryForm = () => {
 
   const handleCrossPodQuery = async (event) => {
     event.preventDefault();
+    dispatch({ type: 'SET_PROCESSING' });
+
     try {
       const documentUrl = await fetchDocuments(
         session,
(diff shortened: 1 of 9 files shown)


leekahung:
This will add function that disable button while application is processing request to upload/query/delete files from Solid; Will be enabled again after its processed

Aim of this is to reduce potential multiple of the same request from being sent to Solid

As a side, also fixed JSDoc for statusReducer

Jared-Krajewski:
not sure if this is working as intended, I Haven't had a chance to troubleshoot why but it still allows spam submissions when I cloned and ran it.

leekahung:
No, you are correct. At the moment, this feature is still a little buggy. Trying to figure out the positioning on where I set the button disabled state within runNotification and the handler functions in the Forms.

leekahung:
Did a few fixes and refactoring for runNotification and the Form handler functions. Think it should be working now.

Issue #20, opened 2023-02-08
Deploy React app to project's GitHub Page via GitHub Actions


k-mack:
**Is your feature request related to a problem? Please describe.**
Automated deployments will eliminate the overhead of manual deployments.

**Describe the solution you'd like**
When a branch is merged into Master, the React App on Master is built and deployed to GitHub Pages.

**Describe alternatives you've considered**
None.

**Additional context**
None.

k-mack:
Note to self. `refs/heads/Master`, not `ref/heads/Master`.

Jared-Krajewski:
fixed with <path: './dist'> change under upload frontend assets in ci.yml and adding <base: '/PASS/',> to vite.config.js

PR #58, opened 2023-03-20
minor changes to vite.config.js and ci.yml to fix GH-pages integration

Jared-Krajewski:
for issue #20 ..Tested on forked Repo and working with changes to 'path' : to './dist' under upload frontend assets of ci.yml and base: '/PASS/', added to vite.config.js. This should fix the deployment but it wont run until merged with Master.


leekahung (approved):
The pathing and configuration from Jared's fork seems to be in order. Think it'll work here as well.

Issue #55, opened 2023-03-15
General review/refactoring of codebase for typos/errors/bugs


leekahung:
**Is your feature request related to a problem? Please describe.**
A general review of the codebase for typos/errors/bugs (no new features will be added)

**Describe the solution you'd like**
Going over each file for typos/errors/bugs and modifying them if needed

**Describe alternatives you've considered**
None

**Additional context**
None

leekahung:
Branch has been merged into Development. Closing issue.

PR #56, opened 2023-03-15
Merging issue-55/general-code-review to Development; Includes guard rails for setDocAclPermission

leekahung:
This branch was built from #53 branch, so changes from the unmerged #53 can also be found in the commits/file changes here. Can be merged into Development after #53 and is reviewed.

This branch was initially created to review/refactor existing codebase without including additional features. Additionally, to keep track of general updates to existing code, a CHANGELOG.md is now included to this branch and hopefully into Development, then Master.

The idea of the changelog was discussed during the March 14, 2023 meeting. Overall, it'll be used to include a general description of all future changes to the codebase in this file (instead of just a commit description), including what it is (like a feature) and their pull request number (i.e. #56), so that everyone can be kept up-to-date. Feedback for what to include in CHANGELOG are welcomed.

However, one feature has been introduced in this PR. I've found out while writing out general instructions for "How to use PASS" ([Google Doc to Instructions](https://docs.google.com/document/d/1VtlBW7fkELDzIkKGdOZ-e12A2rmxB6I2yPFCh1vo820/edit?usp=sharing)), that users can accidentally revoke their own access to their document container using the Permission to Files section.

This is bad since it'll force users to have to manually delete their faulty ACL from their Pod before the container can be accessed again. Not only that, but that their original ACL file will be lost as a result. So the new feature is included as a guard rail that prevent users from doing that. That's the only feature included here, which is addressed in #57.

PR #59, opened 2023-03-20
Merge development into master with issue #55 updates

Jared-Krajewski:


leekahung (approved):
This is fine. These are the same changes I've done for #55 and #56.

PR #63, opened 2023-03-20
62 chromes file explorer crashing application

leekahung:
Refactoring handleFileChange in UploadDocumentForm and changing input name for upload-doctype from file to upload-doctype

PR #65, opened 2023-03-21
Development to master to fix bug causing chrome crash

Jared-Krajewski:
merge to fix crash in chrome due to extra "." in file types of input form.


leekahung (approved):
I believe the bug is found and correct for in PR #63.

PR #74, opened 2023-04-03
Issue 72/user list management, Issue 73/refactoring, Issue 76/application routing

leekahung:
Draft PR for Issue #72 (user list manager) and includes Issue #73 (refactoring session-core/session-helper/reducer, etc.) as part of the commits. Issue #76 (application routing) is also included here in preparation for MUI and multi-page support.

Improvements here includes the addition of a user list manager and login tracker (both of which could be subject to change as is) in preparation for a home screen. Functionality and components build here could be used for that.

The user list can only be accessed by the user themselves within Pod container Users and shows the data for a user's name and their Pod URL.

The login tracker only shows the time and date of their latest login and is placed within public, which only provides read access to other users.

Combined these two functionalities provides a list of users that could be seen from their main page the other user's name, the other user's  Pod URL, and their Last Active Date.

PR #75, opened 2023-04-04
Material UI

andycwilliams:
Material UI test


.parcel-cache/092db1189cc128d8
leekahung, line 1: .parcel-cache and dist directories can be removed from pull request.

For Parcel, we're no longer using it as our bundler. For dist, the directory will be built for deployment in our production build.

src/components/Login/Login.jsx
```diff
@@ -1,5 +1,16 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
leekahung: In Line 1, PR #74 will deal with linting rule by moving this into eslintrc.js, so the can be removed for now (see .eslintrc.js from branch issue-72/user-list-management if you wish to implement that rule to your branch)

leekahung (changes requested):
Minor review for now, will go more in-depth once #74 is merged to Development.

.parcel-cache/092db1189cc128d8
andycwilliams: These have been removed in the latest commit

Jared-Krajewski:
closing pull request due to new approach being used to implement MUI.

PR #84, opened 2023-04-14
Separating loading of UserList and User's last active date

leekahung:
Pull request involves the refactoring of the Users List by separating the loading of User List and user last active date in an asynchronous manner.

Have also included simple loading notifications for when the data is being fetched from Solid.


Jared-Krajewski (approved):
Changes made to match what was discussed in discord voice chat. Loads user list without last active date to speed up initial paint to window then adds in last active date when available. 🚀

PR #90, opened 2023-04-26
Fixing potential redirect bug

leekahung:
Using window.location.href as initial state.


Jared-Krajewski (approved):
Testing to resolve redirect issue that works locally but does not work on github pages. Only way to verify is to push to master.

PR #97, opened 2023-04-28
Cross-pod Upload/Update functionality

leekahung:
Updated existing functionality for file upload and update to handle cross pod file upload and updates.

Created new container called Documents which allow other users access to upload/update documents in it. At the moment, ACL permissions are set to full access for both the user and others (if permission has been set for it). New functionality setDocContainerAclPermission() allows for this and is similar to setDocAclPermission() in functionality which sets permissions to files.

New components are included to support these new functionality and CrossPodWriteForm is no longer just a placeholder.


Jared-Krajewski (approved):
discussed and tested changes in discord voice chat and reviewed code changes.

PR #101, opened 2023-04-29
Add inactivity notification popup that toggles after 3 minutes of no …

benzelcurry:
### Issue-88: Inactivity message

# Implementations/Features
- Created an inactivity message component in the form of a modal that triggers after 3 minutes.
- Only displays if a user is actively signed in
- Displays in the center of user's screen.
- Allows user to either close the modal and continue session or to log out.
- Uses Material UI for base <Modal /> and <Button /> components.
- Just has inline styling for now, but could probably be spruced up in the future by the design team.

# Notes
- Can provide a picture of what the notification looks like. Not sure how to add it to the PR without uploading it to the repo as a whole.

### DIFF
--- a/src/components/Notification/InactivityMessage.jsx
+++ b/src/components/Notification/InactivityMessage.jsx
@@ -1,0 +1,87 @@
+import React, { useState, useEffect } from 'react';
+import { useNavigate } from 'react-router-dom';
+import { Modal, Button } from '@mui/base';
+
+/**
+ * Inactivity Notification Component - Component that displays a popup modal
+ * after 3 minutes of inactivity, prompting the user to either logout or
+ * continue their session.
+ *
+ * @memberof Notification
+ * @name InactivityMessage
+ */
+
+const InactivityMessage = () => {
+  const navigate = useNavigate();
+
+  const [showPopup, setShowPopup] = useState(false);
+  const [activeUser, setActiveUser] = useState(false);
+
+  // Checks for active user by looking for a loggedIn key in local storage
+  useEffect(() => {
+    const activeCheck = localStorage.getItem('loggedIn');
+    if (activeCheck) setActiveUser(activeCheck);
+  }, []);
+
+  // Toggles the popup after three minutes of inactivity
+  useEffect(() => {
+    let timer = null;
+
+    const resetTimeout = () => {
+      if (timer) clearTimeout(timer);
+  
+      timer = setTimeout(() => {
+        setShowPopup(true);
+      }, 180000)
+    }
+  
+    const handleUserActivity = () => {
+      resetTimeout();
+    };
+  
+    resetTimeout();
+  
+    window.addEventListener('mousedown', handleUserActivity);
+    window.addEventListener('mousemove', handleUserActivity);    
+
+    return () => {
+      clearTimeout(timer);
+      window.removeEventListener('mousedown', handleUserActivity);
+      window.removeEventListener('mousemove', handleUserActivity);
+    }    
+  }, [])
+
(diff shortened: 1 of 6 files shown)


src/components/Notification/InactivityMessage.jsx
```diff
@@ -0,0 +1,87 @@
...
+               alignItems: 'center', borderRadius: '10px', position: 'fixed',
+               top: '50%', left: '50%', transform: 'translate(-50%, -50%)', zIndex: 99 }}
+    >
+      <div style={{ display: 'flex', flexDirection: 'column', alignItems: 'center', padding: '50px', border: '2px solid black', borderRadius: '10px' }}>
+        <p>You've been inactive for a few minutes now. Would you like to log out?</p>
+        <div style={{ display: 'flex', gap: '20px' }}>
+          <Button onClick={() => setShowPopup(false)} style={{ width: '100px', backgroundColor: 'white', borderColor: 'black', borderRadius: '5px', cursor: 'pointer' }}>
+            Continue Session
+          </Button>
+          <Button onClick={() => handleLogout()} variant='contained' style={{ width: '100px', backgroundColor: '#4287f5', color: 'white', borderColor: 'black', borderRadius: '5px', cursor: 'pointer' }}>
```
leekahung: I see that you're using the Button component from MUI for handling logout, but why no just wrap it inside LogoutButton from Inrupt's Solid React SDK? Something like:

```jsx
  <LogoutButton onLogout={handleLogout}>
    <button type="button" onClick={() => setShowConfirmation(false)} variant='contained' style={{ width: '100px', backgroundColor: '#4287f5', color: 'white', borderColor: 'black', borderRadius: '5px', cursor: 'pointer' }}>
      Logout
    </button>
  </LogoutButton>
```

src/components/Notification/InactivityMessage.jsx
```diff
@@ -0,0 +1,87 @@
...
+    localStorage.removeItem('redirectUrl');
+    localStorage.removeItem('restorePath');
+    localStorage.removeItem('issuerConfig:https://opencommons.net');
+    navigate('/');
+  };
+  
+
+  return (
+    <Modal open={showPopup && activeUser} 
+      style={{ display: 'flex', backgroundColor: 'white',
```
leekahung: I see that you're using style for the HTML elements, but I was wondering if there's a reason why not use the sx prop for styling MUI components?

src/components/Notification/InactivityMessage.jsx
```diff
@@ -0,0 +1,87 @@
...
+            Log Out
+          </Button>
+        </div>
+      </div>
+    </Modal>
+    )
+}
+
+
+export default InactivityMessage;
```
leekahung: I would also add a blank line at the end of the file so that it'll be consistent with the other files.

src/components/Notification/InactivityMessage.jsx
```diff
@@ -0,0 +1,87 @@
...
+      resetTimeout();
+    };
+  
+    resetTimeout();
+  
+    window.addEventListener('mousedown', handleUserActivity);
+    window.addEventListener('mousemove', handleUserActivity);    
+
+    return () => {
+      clearTimeout(timer);
```
leekahung, line 51: I think this would work just fine. How about adding "touchstart" as well to cover for mobile cases and touch screens? I think that'll be great!

leekahung (changes requested):
Great work! Just a few things I want to touch on for revision.

leekahung:
Also changing the merge to Development instead of Master

leekahung (approved):
Looks good to me.

PR #103, opened 2023-04-29
Replacing BrowserRouter with HashRouter

leekahung:
As discussed previously, BrowserRouter and GitHub Pages are not currently compatible. Have tried HashRouter last time, but not a large extent.

HashRouter works for GitHub Pages, but not with Inrupt's libraries because of the hash fragment as the redirectUrl. But this can be fixed by removing the hash fragment from the initial window.location.href call using .split("#") and using that as the redirectUrl instead. This way HashRouter will not clash with Inrupt's libraries.

I've tested this locally and it seems to work. Have also tried it in a local fork: [https://leekahung.github.io/PASS/#/PASS/](https://leekahung.github.io/PASS/#/PASS/)


Jared-Krajewski (approved):
discussed and tested in discord voice chat. Merging to fix Github pages refresh.

PR #106, opened 2023-05-03
Add some scripts and environment config to allow developers to easily use a local server.

timbot1789:
### Purpose:
This PR adds `community-solid-server` as a developer dependency to the project, and integrates it with the PASS application. The intent is to allow developers to run a solid server locally that they can test the PASS project against. That way they don't have to access `opencommons.net` to run tests. Although we can still use `opencommons.net` as an integration testing area, and developers can still target that as their server if they choose.

### Changes

- PASS now checks the runtime environment, instead of a hardcoded string, to get the pod URL.
- PASS checks a different env variable depending on the application `mode`
- `community-solid-server` is added as a dev dependency
- The script `dev:pod` is added to `package.json` to make starting the solid server easier

### Testing

1. Pull Branch
2. Run `npm install`
3. Run `npm run dev:pod`
4. Run `npm run dev`
5. Test that PASS points to localhost:3000, not opencommons.net. The application should work otherwise normally
6. Take down PASS server
7. Run `npm run build`
8. Check that PASS points to opencommons.net


.nvmrc
```diff
@@ -0,0 +1 @@
+16
```
timbot1789, line 1: Sets a default node version to use in the project. NVM makes different node versions easier to manage. Community Solid Server expects Node 16. It seems to run on Node 18 fine, but doesn't officially support it.

.env
```diff
@@ -0,0 +1,3 @@
+VITE_SOLID_IDENTITY_PROVIDER_PRODUCTION="https://opencommons.net"
```
timbot1789: It's generally considered bad practice to put `.env` files in source control, since they often contain sensitive information like private keys. But IMO it's fine in this case because otherwise pulling this branch would break PASS without dev action

.env
```diff
@@ -0,0 +1,3 @@
+VITE_SOLID_IDENTITY_PROVIDER_PRODUCTION="https://opencommons.net"
+
+VITE_SOLID_IDENTITY_PROVIDER_DEV="http://localhost:3000"
```
leekahung: Nice! Thanks for starting a branch for this issue and make a PR.

With regards to an alternative Solid Identity Provider or OIDC issue, I think we should use https://solidcommunity.net instead of localhost:3000.

This way PASS is set up to a valid Solid server.

package.json
```diff
@@ -42,6 +43,7 @@
     "last 3 edge versions"
   ],
   "devDependencies": {
+    "@solid/community-server": "^5.1.0",
```
leekahung, line 46: While I think we need to swap over from Node Solid Server to Community Solid Server, I'm not too sure about having it integrated with PASS directly.

I would need some opinion from @k-mack. Do you think this would be something that we want for PASS?

leekahung (changes requested):
See feedback.

package.json
timbot1789: I'm not sure I understand what you mean. Maybe I wasn't clear about what this does. This is just a dev tool to make development easier for engineers. Instead of having to test PASS on a remote server (something they don't have control over), this just makes it convenient for them to run their own local server, that they do have control over. It doesn't integrate with PASS directly.

I agree that PASS shouldn't have any specific integration with a particular type of server. It's a client application, and should be able to run against any valid SOLID server. It doesn't (and shouldn't) have knowledge of the specifics of the server implementation it's calling.

.env
timbot1789: The developer can certainly redirect PASS to point at a different Identity Provider. They need simply change these values to a different URL. I just wanted to include some sensible defaults here. And move the URLs to a more standard file. I actually think it's best to remove this `.env` file from source control entirely after distributing it.

I think it's a good idea to run against localhost at least while developing since `solidcommunity.net` can be taken down at any time. If that happens, our development will stall. If we are instead able to run a server on our own computer, each developer will be able to put up their own server and keep developing normally. I'm just trying to make that easier.

leekahung: Perhaps I'm a bit unclear as to the changes being made. But, I'm assuming it suppose to allow us to bypass login and allow us to go to different pages? Let me know if this is not correct.

Of course, we can't onto a Solid Pod without logging in, so I can ignore that for now. However, I can't seem to access /PASS/home/ and /PASS/forms/ using localhost:3000.

package.json
leekahung: Sorry @timbot1789, I think I might be a bit confused too on my end. Please correct me if I'm wrong on this.

You're trying to do now is to allow devs to bypass the existing Solid Pods on opencommons.net and solidcommunity.net by trying to create the Pod locally at localhost:3000, correct? So that way, we can have some mock server functionality to log in and work on the rest of the application or allows us to bypass log in completely and just move onto the other pages to work on them.

.env
leekahung: Ah, never mind. I understand now. It was a misunderstanding on my part. This would be alright with me as a fix/potential start to migrating to Community Solid Server, albeit locally for now.

package.json
leekahung: Never mind, it was a misunderstanding on my part.

leekahung (approved):
Got things running properly with the dev environment. Got a taste of Community Solid Server locally. I think It'll be good to have this as our dev environment. We just need to update CONTRIBUTING.md and README to let devs know of the changes which allows them to work locally.

.env
timbot1789: Updated this to point to `solidcommunity.net` by default.

Jared-Krajewski (approved):
very useful given the current state of our solid server and upcoming change to community solid server. Thanks for the updates!

PR #107, opened 2023-05-04
Update to Issue 43 - Confirmation Modal for logout

xscottxbrownx:
**Updating this confirmation popup:**
Screenshot 2023-05-04 at 11 40 14 AM
 - that still allowed user to focus and type on background elements, had some accessibility issues, etc
 
 
 **To this:**
Screenshot 2023-05-04 at 11 41 43 AM
- styled with Material UI (as per current direction - colors and text can be changed)
- added Material Icons to `package.json` as we will need for upcoming styling per UX team anyway, and used some here

Created as it's own component `LogoutModal.jsx`.

One future possibility is to refactor this to just be a base `Modal` component that we could reuse more often by passing in `Title`, `Text`, and any necessary `CallToAction`s (like buttons/links/etc.)

As of now, this component lives in the `Login` folder.  If we ever refactor to just be a modal component - move that to a new folder?

### DIFF
--- a/src/components/Login/Logout.jsx
+++ b/src/components/Login/Logout.jsx
@@ -1,5 +1,17 @@
+// React Imports
 import React, { useState } from 'react';
-import { useSession, LogoutButton } from '@inrupt/solid-ui-react';
+// Solid Imports
+import { useSession } from '@inrupt/solid-ui-react';
+// Material UI Imports
+import Box from '@mui/material/Box';
+import Button from '@mui/material/Button';
+import Link from '@mui/material/Link';
+import Typography from '@mui/material/Typography';
+import LogoutIcon from '@mui/icons-material/Logout';
+// Custom Component Imports
+import LogoutModal from './LogoutModal';
+
+
 
 /**
  * Logout Component - Component that generates Logout section for users to a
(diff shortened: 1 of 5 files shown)


src/components/Login/Logout.jsx
```diff
@@ -9,45 +21,59 @@ import { useSession, LogoutButton } from '@inrupt/solid-ui-react';
...
 const Logout = () => {
+
   const { session } = useSession();
-  const [showConfirmation, setShowConfirmation] = useState(false);
   localStorage.setItem('loggedIn', true);
+  // state for LogoutModal popup
+  const [showConfirmation, setShowConfirmation] = useState(false);
 
-  // Event handler for logging out of PASS and removing items from localStorage
+  // Event handler for logging out of SOLID POD and removing items from localStorage
```
leekahung, line 33: Yeah, I think this should be the case with Community Solid Server and we're finally logging out of the Solid Pod as well as PASS.

So I think this is fine, just need to confirm it on my part.

leekahung (approved):
This looks great. Just checked it out locally on my dev environment. I'm good with this being merged into Development

Jared-Krajewski (approved):

PR #108, opened 2023-05-06
Issue 109/Login styling updates (outdated)

andycwilliams:
Starting new branch off of Development. Adding styling only to Login page. Had no issues logging in afterwards.


scratch.js
```diff
@@ -0,0 +1,205 @@
+/* ----- NEW CLIENT DIALOG BOX -----
```
leekahung: I believe this file is not being used currently? I think can remove this file.

src/components/Login/Login.jsx
```diff
@@ -17,24 +22,45 @@ const Login = () => {
 
   return (
     <>
+      <CssBaseline />
```
leekahung: I believe CssBaseLine is something that should be used in the global index.jsx for MUI's CSS Reset, not here in individual components.

src/theme.js
```diff
@@ -0,0 +1,47 @@
+import { createTheme } from '@mui/material/styles';
+
+const theme = createTheme({
+  typography: {
+    fontFamily: [
```
leekahung: I am curious as to why there's so many font families in here? Can we not just select one, like "Roboto"

leekahung (changes requested):
Seems to be running fine, login seems to be working as intended.

See feedback for requested changes.

scratch.js
andycwilliams: Yes it can

src/theme.js
andycwilliams: They're used as backups in case the prior ones cannot be loaded, but yeah we can cut down to one or two

leekahung:
Oh, one more thing, please refrain from naming branches with "#". The "#" I gave as an example is meant to be a placeholder for the actual issue number. Like issue #108. So we should really be naming this branch "issue-108/login-styling-updates".

PR #111, opened 2023-05-06
Issue 109/login styling updates

### LINKED ISSUE #109: Login Styling Updates
Styling login page with MUI

andycwilliams:
Adding MUI styling for Login page


src/theme.js
```diff
@@ -0,0 +1,36 @@
+import { createTheme } from '@mui/material/styles';
+
+const theme = createTheme({
+  typography: {
+    fontFamily: ['-apple-system', 'BlinkMacSystemFont', '"Segoe UI"', 'Roboto'].join(',')
```
leekahung: Just want to make sure. Is "-apple-system", "BlinkMacSystemFont", and "Segoe UI" going to be used for styling fonts in PASS? If it is, I think it's fine. But if they're not, perhaps we can cut down on them to just "Roboto"

leekahung (changes requested):
See feedback for request changes.

src/theme.js
andycwilliams: That decision will probably be ultimately left to UX/UI. I think it is fine to leave as is, but can trim it down for now.

leekahung:
For commit "3c97388", personally, I'll rename the commit message with git rebase to "Refactoring CSSBaseline from Login.jsx to index.jsx" or something along that line so that it'll be more descriptive if someone decided to going to the commits to see what the changes are specifically.

src/theme.js
leekahung: Avoid resolving conversations that was started from others to avoid confusion. Let the person who started the conversation resolve it if they think it's sufficiently resolved.

Jared-Krajewski: Agree to remove extra fonts for now leaving just 'Roboto'. We can add future ones when the need arises. Otherwise I'm good with this merging to Development.

leekahung: Let's trim them down. Jared makes a good point. We'll cross that bridge when when the UX/UI team decides we actually need more fonts for PASS.

xscottxbrownx:
`Login.jsx`
line 34: the `htmlFor` is doing nothing because that `id` does not exist on anything. I believe you want it on <LoginButton>
line 38: `div` with a classname of "row" - no need for the class, and may want to change to <Box>

Honestly it doesn't really matter for this route/page. It needs a complete redesign. This is not what we would want or users would expect.  This is like the landing page of the whole site.  It should be the normal navbar on top (with a login button in it to right side), and like a big hero image and description of what PASS is and how it benefits the user.

--------------

Roboto font comments...  IF we are going to only have minimal fallbacks (which is fine) we need to have the Roboto font installed in the index.html
⬆️ This is part of the first steps of MUI that we didn't do (found here: https://mui.com/material-ui/getting-started/usage/)

Screenshot 2023-05-05 at 7 27 02 PM

src/components/Login/Login.jsx
```diff
@@ -18,23 +22,43 @@ const Login = () => {
...
+            flexDirection: 'column',
+            alignItems: 'center',
+            textAlign: 'center',
+            padding: '20px'
+          }}
+        >
+          <Paper elevation={2} sx={{ display: 'inline-block', mx: '2px', padding: '20px' }}>
+            <section id="login">
+              <div className="row">
+                <label id="labelLogin" htmlFor="btnLogin">
```
leekahung: Thanks @xscottxbrownx. I did miss those.

For Line 39, htmlFor was for when previously used when we're still using Inrupt's Solid JS client library for a customized button instead of Inrupt's Solid SDK's LoginButton component. We're no longer using that and it seems like we haven't removed it from before.

We should just remove the htmlFor="btnLogin". It no longer used in Login.jsx.

src/components/Login/Login.jsx
```diff
@@ -18,23 +22,43 @@ const Login = () => {
...
+            display: 'flex',
+            flexDirection: 'column',
+            alignItems: 'center',
+            textAlign: 'center',
+            padding: '20px'
+          }}
+        >
+          <Paper elevation={2} sx={{ display: 'inline-block', mx: '2px', padding: '20px' }}>
+            <section id="login">
+              <div className="row">
```
leekahung: The CSS class "row" in Line 38 is no longer being used for PASS; className="row" can be removed completely.

src/theme.js
```diff
@@ -0,0 +1,36 @@
+import { createTheme } from '@mui/material/styles';
+
+const theme = createTheme({
+  typography: {
+    fontFamily: ['Roboto', 'sans-serif'].join(',')
```
leekahung: If we really just want a fallback of "sans-serif" instead of the browser's default, in case "Roboto" breaks for any reason, I think we can just use fontFamily: "Roboto, san-serif".

This seems overly complicated? I think MUI's documentation also allows the syntax that I've just used for fontFamily. (see theme for fontFamily specifically)

Screen Shot 2023-05-05 at 19 51 23

leekahung (changes requested):
See requested changes.

src/components/Login/Login.jsx
andycwilliams: That entire div should be unneeded then?

leekahung (approved):
I'm okay with this.

xscottxbrownx (approved):
**Login.jsx:**

Line 50 can be deleted entirely. 

Not really an issue though, again this will completely need to be changed anyway.

andycwilliams:
> **Login.jsx:**
> 
> Line 50 can be deleted entirely.
> 
> Not really an issue though, again this will completely need to be changed anyway.

I believe this can be done in the future. I'm not sure it's worth a full commit to remove a single space.

andycwilliams:
Misclicked and closed by accident. Reopened.

PR #114, opened 2023-05-07
Fixing ESLint Pathing

leekahung:
It seems like ESLint wasn't pathed correctly. It does not follow the same syntax as Prettier for processing files.

However, this issue is now fixed with the following changes in package.json:

```
    "lint": "eslint \"src/**/*.js[x]\"",
    "lint:fix": "eslint --fix \"src/**/*.js[x]\"", 
```

Had ran Prettier again to make sure that it was working properly as well. It updated and formatted the new lines from @timbot1789 in session-helper from the recent PR after running npm run prettier:run after finding out it's not formatted correctly using npm run prettier:check.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,8 +9,8 @@
     "preview": "vite preview",
     "docs": "jsdoc --readme ./docs/README.md -c jsdoc.config.json",
     "docs:clear": "node clearDocs.js",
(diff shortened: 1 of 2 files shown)


Jared-Krajewski (approved):
good catch

PR #116, opened 2023-05-09
Issue 110/appheader styling updates

### LINKED ISSUE #110: AppHeader Styling Updates
Styling AppHeader with MUI

andycwilliams:
Adding styling to AppHeader. AppHeader is now consolidated with the previous NavBar and changes upon login.

The current layout is based on an in-progress mobile-first design developed by the UX team. As such, it is subject to change based on their results. It contains a search bar, menu, and notifications icon, each one responsive to screen size changes.

The Home and Forms tabs have been moved into the menu to reduce clutter. The remaining icons, such as My Account, are currently not functional.

The first commit (a8dec0c) styles only the existing elements present in the Development branch and can easily be reverted to.


xscottxbrownx:
I feel like this one may take a discussion (especially with Design Team.)

### I will talk about design here, and actual code tomorrow in new comment. 
There are things I like, and things I dislike.

---
**1. On mobile view we lose all branding in `Navbar`. There's no logo/text. And we are leaving `Search` and `Notifications`.**
   - I can understand `Notifications`, but `Search`? How often do you search a full website?
   - Personally I think it should be app `Name/Logo` on left and `Hamburger menu` on right
---
**2. This `Search`... is it searching the whole site?** 
   - **If so**, we should leave in `Navbar`.  
   - If it's only searching the screen we are on (the displayed route/component), then it shouldn't be in `Navbar`. It should be in the route/component below `Navbar`.
       - In this case, it's not a navigation.
---
**3. I like the search icon, animation when clicked to open... I dislike that it moves the app title.** 
   - I think that should stay in place and search fills the space allotted to it.
---
**4. `Add button` in `Hamburger menu` should not be in `Navbar`.**
   - This button should be it's own component that gets rendered/displayed in the route/page shown below the `Navbar`.
   - This button will either be adding cases/users/people, documents, or whatever other functionalities we come up with and belongs with those things.
   - It's not a navigation (search could be, IF it's searching the whole site - if searching just page, should only be in page and not `Navbar` either.)
   > This was an older iteration, but...
   > examples of `Search` and `Add` being in components they are doing those things:
   
Screenshot 2023-05-08 at 9 22 46 PM

---
**5. In `Hamburger menu` I think `My Account` should be under the navigation links.**
   - I feel it should be all the routes the user may want to go to, then `My Account`, then `Log Out`
   - And maybe `Log Out` in a different color too? Maybe the red, sort of like a "danger" - more intuitive that way?
   - Reason for this layout (Routes, Account, Logout) is what will be commonly used. Much more common to change pages/views/tabs than to go to your account settings. A lot of people don't log out of any website accounts before closing either.

leekahung:
Had a look at the design from the branch. I think we should have a discussion with the UX team first before implementing this, especially since this is "in-progress" as stated from the PR. It's likely not what they want implemented at the end.

To that end, I think we should just implement what you had previously on the MUI branch as our baseline styling. The previous iteration seems like a more sensible mobile-first design.

leekahung:
Had a look at the design from the branch. I think we should have a discussion with the UX team first before implementing this, especially since this is "in-progress" as stated from the PR. It's likely not what they want implemented at the end.

To that end, I think we should just implement what you had previously on the MUI branch as our baseline styling. The previous iteration seems like a more sensible mobile-first design.

andycwilliams:
I have reverted back to the original, minimum amount of styling.

src/components/AppHeader.jsx
```diff
@@ -12,24 +19,40 @@ import { useSession } from '@inrupt/solid-ui-react';
...
-      <h2>Getting Started with PASS</h2>
-      {session.info.isLoggedIn ? (
-        <>
-          <Link to="/PASS/home/">Home</Link>
-          <Link to="/PASS/forms/">Forms</Link>
-        </>
-      ) : null}
-    </header>
+    <>
+      <CssBaseline />
```
leekahung: We already have CssBaseline moved to index.jsx. I don't think we need another one here in Line 24. We can remove this without affecting the styling.

src/components/AppHeader.jsx
```diff
@@ -12,24 +19,40 @@ import { useSession } from '@inrupt/solid-ui-react';
...
+      <Box sx={{ flexGrow: 1 }}>
+        <AppBar position="static" color="primary">
+          <Toolbar>
+            {session.info.isLoggedIn ? (
+              <>
+                <Typography variant="h4" noWrap component="div">
+                  PASS
+                </Typography>
+                <ButtonGroup
```
leekahung: I do think we want to cluster the links, but I don't think we need a ButtonGroup.

src/components/AppHeader.jsx
```diff
@@ -12,24 +19,40 @@ import { useSession } from '@inrupt/solid-ui-react';
...
+                  <Link to="/PASS/home/">
+                    <Button>Home</Button>
+                  </Link>
+                  <Link to="/PASS/forms/">
+                    <Button>Forms</Button>
+                  </Link>
+                </ButtonGroup>
+              </>
+            ) : (
+              <Typography variant="h4" noWrap component="div">
```
leekahung: I believe "Getting started with PASS" was centered in the newer version? I think that's fine. Of course if the screen gets too small, we could always shortened it to just "PASS".

We need to consider using just PASS after login though. Having PASS centered on mobile would be great (because of the small screen), but it might not work for laptop/desktop screens, since it'll look out of place. So we can make this responsive to be left-justified on mobile instead of centered. That would be great.

leekahung (changes requested):
Hey Andy, just a bit of feedback to go over.

But I was wondering what happened to the button group on the right with the notification bell and profile menu? I think we can keep those two things for now. Notification bell could be a placeholder for inbox (or simply just an indicator, we'll see) and the profile menu could be used to access the user's Solid Pod.

As for the add, I think we can drop that for now or move it else where. Not sure what we'll use it for as part of navigation, but if we're planning to use that button to be adding something, I think we can move that button to that page rather than keeping it in the navbar.

src/components/AppHeader.jsx
leekahung: Oh, I want to add. If we're wondering how to fit the links for mobile, I think we can use a dropdown hamburger menu for that, similar to what is shown in the newer version. But it should be grouped with the other icons.

leekahung:
I also thought we're transitioning to Navbar and replacing the AppHeader after our recent discussions?

src/components/AppHeader.jsx
andycwilliams: > I also thought we're transitioning to Navbar and replacing the AppHeader after our recent discussions?

The code has already been consolidated. Do you mean changing the file structure and moving it into its own folder?

andycwilliams:
> I also thought we're transitioning to Navbar and replacing the AppHeader after our recent discussions?

The code has already been consolidated. Do you mean changing the file structure and moving it into its own folder?

src/components/AppHeader.jsx
andycwilliams: Done

leekahung:
> > I also thought we're transitioning to Navbar and replacing the AppHeader after our recent discussions?
> 
> The code has already been consolidated. Do you mean changing the file structure and moving it into its own folder?

We did have a Navbar directory with three components in there, so I thought they'll be used. But I don't see it anymore. Did you remove it?

andycwilliams:
> > > I also thought we're transitioning to Navbar and replacing the AppHeader after our recent discussions?
> > 
> > 
> > The code has already been consolidated. Do you mean changing the file structure and moving it into its own folder?
> 
> We did have a Navbar directory with three components in there, so I thought they'll be used. But I don't see it anymore. Did you remove it?

No. It was removed from I rebased from Development. I can add it back, however.

leekahung:
> > We did have a Navbar directory with three components in there, so I thought they'll be used. But I don't see it anymore. Did you remove it?
> 
> No. It was removed from I rebased from Development. I can add it back, however.

We should have rebased it such that you add Development under the MUI branch. That way we would prevent any features made in MUI to be removed.

src/components/NavBar/NavBar.jsx
```diff
@@ -0,0 +1,158 @@
...
+      <AppBar position="static" color="primary">
+        <Toolbar>
+          {session.info.isLoggedIn ? (
+            <>
+              <Typography variant="h4" noWrap component="div" mr="10px">
+                PASS
+              </Typography>
+              <>
+                <Link to="/PASS/home/">
+                  <Button
```
leekahung: For the buttons for Home (Line 60 to 66) and Forms (Lines 69 to 75), I don't think we need them. The Links work just fine.

Screen Shot 2023-05-09 at 14 35 54

src/components/Login/Login.jsx
```diff
@@ -6,7 +6,7 @@ import Container from '@mui/material/Container';
 import Paper from '@mui/material/Paper';
 import { useRedirectUrl } from '../../hooks';
 import { SOLID_IDENTITY_PROVIDER } from '../../utils';
-import AppHeader from '../AppHeader';
+import NavBar from '../NavBar/NavBar';
```
leekahung: I think we can refactor these components into their own index.js inside their directory. We could leave this for now for a future PR.

src/components/NavBar/NavBar.jsx
```diff
@@ -0,0 +1,169 @@
...
+                  edge="end"
+                  aria-label="account of current user"
+                  aria-controls={menuId}
+                  aria-haspopup="true"
+                  onClick={handleOpenMenu}
+                  color="inherit"
+                >
+                  <AccountCircle />
+                </IconButton>
+                <Menu
```
leekahung: Line 121 to Line 142 looks duplicated from Line 88 to 109. Keep one of them, but not both.

src/components/NavBar/NavBar.jsx
```diff
@@ -0,0 +1,169 @@
...
+/**
+ * NavBar Component - Component that generates NavBar section for PASS
+ *
+ * @memberof GlobalComponents
+ * @name NavBar
+ */
+
+const NavBar = () => {
+  const { session } = useSession();
+  const [anchorEl, setAnchorEl] = React.useState(null);
```
leekahung: Use `useState` instead of `React.useState` for consistency with other files. So import React, { useState } from "react' on Line 2.

This is true for Lines 30, 31, and 32.

src/components/NavBar/NavMenu.jsx
```diff
@@ -0,0 +1,57 @@
...
+      transformOrigin={{
+        vertical: 'top',
+        horizontal: 'right'
+      }}
+      open={openMenu}
+      onClose={handleMenuClose}
+      onClick={handleMenuClose}
+      sx={{ marginTop: '40px' }}
+    >
+      <MenuItem>
```
leekahung, line 41: The icon and Profile is off-center. I think we can do a simple update for that styling.

Screen Shot 2023-05-09 at 14 35 18

src/components/NavBar/NavBar.jsx
leekahung: Also I've just noticed, is there a way to remove that blue(?)-ish border from the header? If not, we should have it like with the Login component where they're simply using a thin box shadow.

src/components/NavBar/NavBar.jsx
```diff
@@ -0,0 +1,169 @@
...
+                  size="large"
+                  aria-label="show 17 new notifications"
+                  color="inherit"
+                  onClick={handleNotificationsMenu}
+                >
+                  <Badge badgeContent={17} color="error">
+                    <NotificationsIcon />
+                  </Badge>
+                </IconButton>
+                <Menu
```
leekahung: Compared to NavMenu which have the Menu below the icon, the Notification Menu covers the Bell icon.

Screen Shot 2023-05-09 at 14 40 31
Screen Shot 2023-05-09 at 14 35 18

Make it below their respective icon and make sure they're both leveled at the same distance below the icon (I think you can use NavMenu as a guide for positioning that).

leekahung: Also, while it looks alright for demonstrative purposes, I think we can turn off the badgeContent for now. We don't exactly have any notifications implemented currently, so this will prevent some confusion.

src/components/NavBar/NavBar.jsx
```diff
@@ -0,0 +1,169 @@
+// React Imports
+import React from 'react';
```
leekahung: See feedback for useState on Line 30 to 32 for explanation, but we should change Line 2 to:

`import React, { useState } from 'react'`;

leekahung (changes requested):
Alright, I think we're getting close. See feedback.

src/components/NavBar/NavBar.jsx
andycwilliams: If you mean removing the MUI Buttons, this is the result:


That border is from style.css lines 10-12:

`
header {
  border-bottom: #5795b9 solid;
  padding-left: 0.5rem;
}
`

Should they be deleted?

src/components/Login/Login.jsx
andycwilliams: Noted

src/components/NavBar/NavBar.jsx
xscottxbrownx: Found the blue `border`... took me a minute.

@andycwilliams looks like the `AppBar` MUI component renders a `<header>` element with a border-bottom and that blue color.  I could only get rid of it with an `sx` on the `AppBar` -> `sx={{borderBottom: 'none'}}` worked for me.
Take a look.   It's probably part of the default MUI theme would be my guess. I never noticed it before.

Good catch KaHung. random.

src/components/NavBar/NavMenu.jsx
xscottxbrownx: I think passing a prop to the `<IconButton>` on line 35 will fix. I believe it is `edge="start"`.

src/components/NavBar/NavBar.jsx
andycwilliams: @xscottxbrownx Have you looked at the `style.css` file? Lines 10-12 alter the header element. Removing the styling from that gets rid of the border. That isn't part of MUI.

We also should coordinate on how much of the `style.css` file we want to keep there and how much we want to try and add to the `theme.js` file.

xscottxbrownx: That `<header>` also has a left padding of `0.5rem` no clue why, but seems weird... maybe see what you think about getting rid of that too?
Screenshot 2023-05-09 at 5 39 06 PM

xscottxbrownx: > @xscottxbrownx Have you looked at the `style.css` file? Lines 10-12 alter the header element. Removing the styling from that gets rid of the border. That isn't part of MUI.
> 
> We also should coordinate on how much of the `style.css` file we want to keep there and how much we want to try and add to the `theme.js` file.

Oh perfect!  

I keep forgetting that file exists honestly.  We should probably wipe anything that we can from there I would think...  I'm for keep it all MUI if possible. Just have one source for styling.  But not sure if we'll run into any limitations or not. We'll find out.

andycwilliams: I haven't worked on the 'style.css' file but would be fine removing some of it.

Also, as I've been working on another component, the Upload Document one, the `.panel` styling clashes with MUI so far. So it would help to know if we want to keep that or if it's just a placeholder.

andycwilliams: I haven't tried it yet, but we can probably migrate everything from `style.css` to `theme.js`. Everything we decided on keeping, anyhow.

Some documentation:
https://mui.com/material-ui/react-typography/

Scott, maybe you could tackle that next? I'm still working on a couple of components.

xscottxbrownx: > If you mean removing the MUI Buttons, this is the result: 
> 
> That border is from style.css lines 10-12:
> 
> `header { border-bottom: #5795b9 solid; padding-left: 0.5rem; }`
> 
> Should they be deleted?

Personally I like the text links better, but maybe use the `Tabs` MUI component?  Especially colored tabs using the `theme.palette.tertiary.main color` ?? That's my thought/idea.  https://mui.com/material-ui/react-tabs/#colored-tab

You also had some div and styling to them previously though: 
Screenshot 2023-05-09 at 5 47 05 PM

src/components/Forms.jsx
```diff
@@ -28,7 +28,7 @@ const Forms = () => {
 
   return (
     <>
-      <AppHeader />
+      <NavBar />
       <Logout />
```
xscottxbrownx, line 30: Maybe separate PR for this later?
---

I think `<Logout />` should go away entirely. I believe the button from it should be moved into the `NavMenu` with the profile button. I think this is where logout/login typically lives.

But this would require using `session.info.isLoggedIn` in `NavMenu` (already made available by the `import { useSession } from '@inrupt/solid-ui-react';` in there - so you can conditional render `LogoutButton` or `LoginButton` with a ternary.  

The above also requires taking the `LoginButton` from `Login.jsx`

src/components/NavBar/NavBar.jsx
leekahung: > If you mean removing the MUI Buttons, this is the result:
> 

Can we not just place them in a Container, Box, or Grid component and set a gap between items? We have something similar before, but with a flex box.

> That border is from style.css lines 10-12:
> 
> `
> header {
>   border-bottom: #5795b9 solid;
>   padding-left: 0.5rem;
> }
> `
> 
> Should they be deleted?

Yes, I don't see a purpose to a bottom border.

src/components/NavBar/NavBar.jsx
```diff
@@ -0,0 +1,169 @@
...
+              <Box sx={{ flexGrow: 1 }} />
+              <Typography variant="h4" noWrap component="div">
+                Getting started with PASS
+              </Typography>
+              <Box sx={{ flexGrow: 1 }} />
+            </>
+          )}
+        </Toolbar>
+      </AppBar>
+      {openMenu ? (
```
xscottxbrownx: **Since KaHung discussed some specifics, I'm going to look at more basics...**

The `menu for the account settings/profile` got refactored into it's own component. 
But the `menu for the notifications` did not. 

Was there a reason why one and not the other?

---

Also the `menu for the account settings/profile` seems to rely on state to render **(line 156)**, where the `notifications menu` seems to solely rely on an `onClick` event. 

Is there a reason for the difference? Did I miss something here entirely? 

---

Maybe if `Navbar` items continue to grow we could discuss need (for maintainability) to have the links be one separate component, the notifications be another, and the profile be another? Not sure that's necessary now, but if menu keeps growing/links keep growing - maybe?

That way we could just create an array of the items, in their own components, and `.map()` over them to create the lists/menus.  Less repeated code maybe.

src/components/NavBar/NavBar.jsx
```diff
@@ -0,0 +1,169 @@
...
+        <NavMenu
```
xscottxbrownx, line 169: **Lines 156-164:**
Since the ternary is "all this" or "nothing",  could use `{openMenu && ( all the elements ) }`

Just another way to conditional render in React.  if the left is true, the right will render. If the left is false, the right will not render.

Doesn't make any difference though really.

src/components/NavBar/NavMenu.jsx
```diff
@@ -0,0 +1,57 @@
...
+        <IconButton
+          size="large"
+          aria-label="account of current user"
+          aria-controls="primary-search-account-menu"
+          aria-haspopup="true"
+          color="inherit"
+        >
+          <AccountCircle />
+        </IconButton>
+        <a
```
xscottxbrownx: - `<a>` should be a `<Link>` from MUI (not react router by accident.) 
- And I feel it should wrap the whole `<MenuItem>` or `<IconButton>` (whichever it is, based on docs and trial/error.)

this way user is redirected there whether they click the icon, the text, or just the row of the menu. Right now it only works if click the text directly.

xscottxbrownx (changes requested):
think I'd like to know more about the menus, and the NavMenu link before approving (on top of KaHung's feedback.)

src/components/NavBar/NavBar.jsx
andycwilliams: MUI has a specific tab component, which I'm currently trying to make work with the React Links. But I can try other types of components as well.

src/components/Forms.jsx
leekahung: Yeah, I think it's totally doable in a separate PR. We can move the Logout and LogoutModal inside NavMenu.

src/components/NavBar/NavBar.jsx
andycwilliams: The above points are based on code I used from MUI templates, so my assumption is that those are at least serviceable. If we want to look into optimizing it, perhaps that can be done in a future PR?

src/components/NavBar/NavBar.jsx
xscottxbrownx: > MUI has a specific tab component, which I'm currently trying to make work with the React Links. But I can try other types of components as well.

If you can get those to work, I think they looked the best out of the text link options.  And try the `theme.palette.tertiary.main` color - that golden yellow. Think it'll look good! and the header and footer would match better too.

I was looking at them earlier, is it something like:
```js
<Tabs all the attributes>
   <Tab label="link name">
      <Link all the stuff />
   </Tab>
</Tabs>
```
???

leekahung: Yes, could we get the Links in the Navbar to be of a different color than the default blue and purple?

For example, if we were to look at the following image for the Navbar, it's very difficult to read given our dark green background.

Screen Shot 2023-05-10 at 1 56 45

Either we keep things uniformed with the color white (e.g. PASS and the other white icons on the Navbar) or we use one of the theme colors which is bright, like the golden yellow Scott used for the footer.

src/components/NavBar/NavBar.jsx
leekahung: That's true. Given what we're rendering either the component or null given a condition, using && might be a better choice than a ternary. I recommend changing it to this.

src/components/NavBar/NavMenu.jsx
leekahung: I don't see any changes to this in the recent commits, so I'm not resolving this yet.

Find a way to center both the icon and the wording. It looks out of place.

src/components/NavBar/NavBar.jsx
leekahung, line 1: Three things I would like to note.

Firstly, the Navbar height is not fixed.

Screen Shot 2023-05-10 at 2 13 49
Screen Shot 2023-05-10 at 2 14 00

I don't see a good reason for this at this time, even for mobile.

Next, there is this extra white padding above and below the dropdown menu.

Screen Shot 2023-05-10 at 2 32 31

Could we get rid of that?

Lastly, I was thinking that maybe a hamburger menu icon would be better the 3 dot, but that's just my opinion for the dropdown menu icon for mobile.

src/components/NavBar/NavBar.jsx
```diff
@@ -0,0 +1,179 @@
...
+                <Box sx={{ display: { xs: 'flex', md: 'none' } }}>
+                  <IconButton
+                    size="large"
+                    aria-label="show more"
+                    aria-controls={mobileMenuId}
+                    aria-haspopup="true"
+                    onClick={handleOpenMobileMenu}
+                    color="inherit"
+                  >
+                    <MoreIcon />
```
leekahung: In Line 141, I think we should use the hamburger menu icon rather than the vertical more icon.

src/components/Login/Login.jsx
leekahung: Given that you've already added the index.js back in to the NavBar directory, I think you should import this properly as:

`import { NavBar } from './NavBar';`

src/components/NavBar/NavMenu.jsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import AccountCircle from '@mui/icons-material/AccountCircle';
+import IconButton from '@mui/material/IconButton';
+import Link from '@mui/material/Link';
+import Menu from '@mui/material/Menu';
+import MenuItem from '@mui/material/MenuItem';
+import { useSession } from '@inrupt/solid-ui-react';
+
```
leekahung, line 8: I think we can add JSDoc comments here for the NavMenu component like what we have for NavBar

src/components/NavBar/NavMenuMobile.jsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react';
+import AccountCircle from '@mui/icons-material/AccountCircle';
+import Badge from '@mui/material/Badge';
+import IconButton from '@mui/material/IconButton';
+import Link from '@mui/material/Link';
+import Menu from '@mui/material/Menu';
+import MenuItem from '@mui/material/MenuItem';
+import NotificationsIcon from '@mui/icons-material/Notifications';
+import { useSession } from '@inrupt/solid-ui-react';
+
```
leekahung, line 12: Same thing here as NavMenu, we can add JSDoc comments here for the NavMenuMobile component like what we have for NavBar

leekahung (changes requested):
See feedback for recent fixes, updates, and files. Just a few more changes I think and they're relatively minor.

src/components/NavBar/NavMenu.jsx
xscottxbrownx: I think passing a prop to the `<IconButton>` on line 35 will fix the off-center issue. I believe it is `edge="start"`.

src/components/NavBar/NavBar.jsx
leekahung: We could save the padding issue for a future PR.

But I'm sure there's a way in MUI to disable that media query for the MUI Toolbar. Else just force overwrite it with `sx={{ minHeight: '64px' }}` in Line 68.

src/components/NavBar/NavBar.jsx
```diff
@@ -0,0 +1,182 @@
...
+      <AppBar position="static" color="primary">
+        <Toolbar>
+          {session.info.isLoggedIn ? (
+            <>
+              <Typography variant="h4" noWrap component="div" mr="10px">
+                PASS
+              </Typography>
+              <>
+                <Box>
+                  <Tabs
```
leekahung, line 76: Seems like we have a minor issue with the Tabs.

https://github.com/codeforpdx/PASS/assets/14917816/9477ed52-c393-4cce-aa7f-c5bf133052af

At first, I thought the text colors were like that because we're only active on one page. However, as you see from the clip, the colors doesn't change when you go to a different page. Either keep the colors uniformed, or make it change color based on the page the user's using.

src/components/NavBar/NavMenu.jsx
leekahung: I saw the new changes to this. I think it looks okay now.

src/components/NavBar/NavBar.jsx
```diff
@@ -0,0 +1,182 @@
...
+
+  const handleOpenMobileMenu = (event) => {
+    setAnchorElMobile(event.currentTarget);
+    setOpenMobileMenu(true);
+  };
+
+  return (
+    <Box sx={{ flexGrow: 1 }}>
+      <AppBar position="static" color="primary">
+        <Toolbar>
```
leekahung: Force an sx for the Toolbar min-height, if you're not sure how to disable the default media query for Toolbar.

leekahung (changes requested):
Just some minor stuff left remaining, I think.

We can move one of the issues (the padding for the dropdown menus) to a future PR.

See feedback for other places.

src/components/NavBar/NavBar.jsx
andycwilliams: I haven't been able to determine what's causing that. If you click it a second time then it highlights properly. And it only occurs on the Forms tab, not Home.

It seems related to `to="/PASS/forms/"` on line 85.

src/components/NavBar/NavBar.jsx
leekahung: Fixed issue with minHeight using sx.

src/components/NavBar/NavBar.jsx
leekahung: Fixed this issue by having the NavBar on a higher level inside index.jsx. This would prevent any potential re-rendering of NavBar, which I think is likely the cause of this problem.

leekahung (approved):
I think we've covered most of what we could in this PR. Any remaining issue would be handled in a later PR.

src/components/NavBar/NavBar.jsx
leekahung: Updated Toolbar to include sx={{ minHeight: '64px' }}

src/components/NavBar/NavBar.jsx
xscottxbrownx: Complete

andycwilliams:
Is this ready to be closed?

xscottxbrownx (approved):
As discussed, any future adjustments will be done on subsequent PRs

Issue #115, opened 2023-05-08
Create footer component


xscottxbrownx:
**Create footer component to use across app**

xscottxbrownx:
PR #118 addresses this issue.

Jared-Krajewski:
resolved

PR #118, opened 2023-05-09
Issue 115/create footer component

xscottxbrownx:
**Description**
---
This PR implements a `<footer>` so we can start laying out pages/views.  UX designers have not done a footer design yet, so this is subject to change.

**Updates**
---
Screenshot 2023-05-08 at 8 28 09 PM

**Related Issue**
---
Closes Issue #115

**Future Thoughts**
---
- Would need PASS logo and tagline, and also the CodeForPDX logo (or rebranded logo) to complete the way it is now. 
- Left placeholder links for possible legal notices
- Will need to be refactored a little to stack vertically for mobile devices instead of horizontally (didn't tackle in this PR, as we need to lift an `isMobileView` state up and create a window resize event listener.)

### DIFF
--- a/src/components/Footer/Footer.jsx
+++ b/src/components/Footer/Footer.jsx
@@ -1,0 +1,113 @@
+// React Imports
+import React from 'react';
+// Material UI Imports
+import Box from '@mui/material/Box';
+import Button from '@mui/material/Button';
+import Container from '@mui/material/Container';
+import Grid from '@mui/material/Grid';
+import Link from '@mui/material/Link';
+import Typography from '@mui/material/Typography';
+import { useTheme } from '@mui/material/styles';
+// Material Icons Imports
+import TwitterIcon from '@mui/icons-material/Twitter';
+import FacebookIcon from '@mui/icons-material/Facebook';
+import InstagramIcon from '@mui/icons-material/Instagram';
+
+const Footer = () => {
+  const theme = useTheme();
+
+  return (
+    <Box
+      component="footer"
+      sx={{
+        position: 'sticky',
+        top: '100%',
+        textAlign: 'center',
+        bgcolor: `${theme.palette.primary.main}`
+      }}
+      py={5}
+    >
+      <Container maxWidth="lg">
+        <Grid container columnSpacing={0} my={2}>
+          <Grid item xs={7} sx={{borderRight: 2, borderColor: theme.palette.tertiary.main, pr: 12}}>
+            <Typography variant="h5" color={`${theme.palette.tertiary.main}`}>
+              Want to partner with PASS?
+            </Typography>
+            <Typography variant="body1" color="#fff">
+              If your organization is interested in partnering with PASS and would like to discuss
+              further, contact us below.
+            </Typography>
+            <Button variant="contained" color="secondary" sx={{ my: '1rem' }}>
+              Partnership Proposal
+            </Button>
+          </Grid>
+          <Grid item xs={2}>
+            <Typography color={`${theme.palette.tertiary.main}`}>PASS LOGO</Typography>
+            <Typography color="#fff" mb={5}>tagline</Typography>
+            <Typography color={`${theme.palette.tertiary.main}`}>Follow Us</Typography>
+            <Link href="https://twitter.com/" target="_blank" rel="noopener" color="#fff" mr={1}>
+              <TwitterIcon />
+            </Link>
+            <Link href="https://www.facebook.com/" target="_blank" rel="noopener" color="#fff">
+              <FacebookIcon />
+            </Link>
+            <Link
+              href="https://www.instagram.com/"
+              target="_blank"
+              rel="noopener"
+              color="#fff"
+              ml={1}
+            >
+              <InstagramIcon />
+            </Link>
+          </Grid>
+          <Grid item xs={2}>
+            <Typography color={`${theme.palette.tertiary.main}`}>Built By:</Typography>
+            <Link
+              href="https://www.codeforpdx.org/"
+              target="_blank"
+              rel="noopener"
+              underline="none"
+            >
+              <Typography variant="body2" color="#fff">
+                C4PDX LOGO
+              </Typography>
+            </Link>
+            <Typography variant="body2" color={`${theme.palette.tertiary.main}`} mb={5}>
+              ©{new Date().getFullYear()}{' '}
+              <Link
+                href="https://www.codeforpdx.org/"
+                target="_blank"
+                rel="noopener"
+                underline="none"
+                color={`${theme.palette.tertiary.main}`}
+              >
+                Code for PDX
+              </Link>
+            </Typography>
+            <Typography variant="body2">
+              <Link
+                href="https://www.codeforpdx.org/"
+                underline="none"
+                color="#fff"
+              >
+                Privacy Policy
+              </Link>
+            </Typography>
+            <Typography variant="body2">
+              <Link
+                href="https://www.codeforpdx.org/"
+                underline="none"
+                color="#fff"
+              >
+                Terms and Conditions
+              </Link>
+            </Typography>
+          </Grid>
+        </Grid>
+      </Container>
+    </Box>
+  );
+};
+
+export default Footer;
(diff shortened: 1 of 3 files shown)


xscottxbrownx:
@andycwilliams @leekahung @Jared-Krajewski 

Whenever you all are free to take a look at the code. We can also tweak the design, but not sure it's necessary as UX Design team will get to it at some point.

Jared-Krajewski:
Currently we have been merging into Development for single issue implementations. Usually after a few decent changes are made and some time before Demo nights we update master so it goes live on Github pages. Development holds the latest updates to base off of for working. So for now going to change this to base from Development. This is likely temporary until MVP

Jared-Krajewski (approved):
Changes look good. In the future is this going to sustain across all routes(after logged in)?

src/components/Footer/Footer.jsx
```diff
@@ -0,0 +1,113 @@
...
+import Container from '@mui/material/Container';
+import Grid from '@mui/material/Grid';
+import Link from '@mui/material/Link';
+import Typography from '@mui/material/Typography';
+import { useTheme } from '@mui/material/styles';
+// Material Icons Imports
+import TwitterIcon from '@mui/icons-material/Twitter';
+import FacebookIcon from '@mui/icons-material/Facebook';
+import InstagramIcon from '@mui/icons-material/Instagram';
+
```
leekahung, line 15: Something minor, we don't have to do it in this PR in particular if we do want to add them. It's already explicit enough being named Footer, but I was thinking on if we would like to add some JSDoc to the file to be consistent. So something along the lines of:

```
/**
 * Footer Component - Footer Component for PASS
 * 
 * @memberof <something>
 * @name Footer
 */
```

leekahung (approved):
I have a small comment on the PR, but overall, this looks good to me as well. I'm okay with merging this into Development.

xscottxbrownx:
> Currently we have been merging into Development for single issue implementations. Usually after a few decent changes are made and some time before Demo nights we update master so it goes live on Github pages. Development holds the latest updates to base off of for working. So for now going to change this to base from Development. This is likely temporary until MVP

I’m guessing it showed master? I could’ve sworn I changed it to Development. Thought it was the first thing I did.   I think Andy said the same thing on my last PR, so clearly I’m missing something.  I will get with one of you on my next PR and share screen and figure it out!

xscottxbrownx:
> Changes look good. In the future is this going to sustain across all routes(after logged in)?

I was thinking yes.  I was thinking all routes would be this:

`<Navbar />`
`{route component}`
`<Footer />`

src/components/Footer/Footer.jsx
xscottxbrownx: Oh yes. Sorry! First time working with JSdoc.  I will add, and also read into it more/make sure to add it to my development routine.

xscottxbrownx: Added

andycwilliams:
I would like to consult with Hugh before dealing with any legal or branding matters. Especially since the Code For PDX name is going to change.

So this only applies to the desktop version? There is no mobile footer in place?

leekahung:
> I would like to consult with Hugh before dealing with any legal or branding matters. Especially since the Code For PDX name is going to change.
> 
> So this only applies to the desktop version? There is no mobile footer in place?

I think it's fine for now since it's just a placeholder with text.

As for making it responsive, let's do it in a new PR.

leekahung (approved):
I think I'm good with the current changes.

xscottxbrownx:
> I would like to consult with Hugh before dealing with any legal or branding matters. Especially since the Code For PDX name is going to change.

Just to be clear, these are just placeholders. I have zero intention on actually writing these policies, that is for a legal team.
But in an app that holds/uses private user data, government docs including things like social security, driver's license, and the like - these WILL absolutely be necessary. Not an option.

So, I put in some placeholders to be replaced when the real things are ready. We just put in the logo images (of course with the C4PDX rebranded name/logo) or put in the href for the links. 

This allows us to see what the footer will actually look like, and saves us time in the future.  If we don't have these, we will just end up redoing the whole footer later and defeat the purpose of attempting now.

---

> So this only applies to the desktop version? There is no mobile footer in place?

Yep. I cannot code the conditional render for a proper mobile version until the Navbar is complete.   

It requires lifting up state into the parent component (App.jsx) as it will effect both the Navbar and the Footer and every other route component.   But the Navbar is currently being worked on and I didn't want to create conflicts.

I envisioned it would be stack rather than horizontal like:
Screenshot 2023-05-08 at 11 22 54 AM

and pushed together more as screen size decreases

BUT all of this can be critiqued and adjusted as we see fit. Just a first crack at the design.  Something to give us a component to work with, so I can go in and get the layout properly for React Router - just need the Navbar and Footer to do so.

leekahung:
> > Changes look good. In the future is this going to sustain across all routes(after logged in)?
> 
> I was thinking yes. I was thinking all routes would be this:
> 
> `<Navbar />` `{route component}` `<Footer />`

Yeah, I think we can create a general layout like that for our future pages.

leekahung:
For the vertical stacking for mobile, I think you and Andy could discuss what would be a good breakpoint for that transition so that it'll be smooth for both Navbar and Footer. I was thinking around 768px (which would target everything below a tablet)?

xscottxbrownx:
> For the vertical stacking for mobile, I think you and Andy could discuss what would be a good breakpoint for that transition so that it'll be smooth for both Navbar and Footer. I was thinking around 768px (which would target everything below a tablet)?

Sure, we just need a state - something like:
> const [mobileView, setMobileView] = useState();

and then set an event listener for window resizing (and a check onMount of window size) to determine this state.

I could do this only in the Footer, but it seemed silly to make a state like this in every component, when we can just make it in App.jsx and pass it to the components that need it.

leekahung:
> I could do this only in the Footer, but it seemed silly to make a state like this in every component, when we can just make it in App.jsx and pass it to the components that need it.

Yeah, I agree. Doing it for every component would probably be exessive. But I was thinking that since Navbar and Footer would work in tandem with one another in a general layout, it should at the very least apply to them both at the same time?

xscottxbrownx:
> > I could do this only in the Footer, but it seemed silly to make a state like this in every component, when we can just make it in App.jsx and pass it to the components that need it.
> 
> Yeah, I agree. Doing it for every component would probably be exessive. But I was thinking that since Navbar and Footer would work in tandem with one another in a general layout, it should at the very least apply to them both at the same time?

That's exactly what I'm trying to portray, it would be applied to all components at the same time (based on the mobileView state (or whatever it's named.)

So in every component will be conditional rendering something like (can be done in lots of ways):

> Footer.jsx for example:
```js
const Footer = ({mobileView}) => {
  const theme = useTheme();
  let footerElements;

// determine footerElements based on screen size
  if (mobileView) {
    footerElements = <all the code for mobile>;
  } else {
    footerElements = <all the current code (code for desktop)>;
  }

// component rendering
  return (
    <Box
      component="footer"
      sx={{
        position: 'sticky',
        top: '100%',
        textAlign: 'center',
        bgcolor: `${theme.palette.primary.main}`
      }}
      py={5}
    >
      <Container maxWidth="lg">
         {footerElements}
      </Container>
    </Box>
  );
};

export default Footer;
```

andycwilliams:
For the mobile version, I think it could be shorter than the desktop version, to avoid taking up too much of the screen. Some information or links could be moved to a separate tab, as with other apps. Especially since the text may end up being small and buttons difficult to press.

As for breakpoints, MUI has information:
https://mui.com/material-ui/customization/breakpoints/

leekahung:
> For the mobile version, I think it could be shorter than the desktop version, to avoid taking up too much of the screen. Some information or links could be moved to a separate tab, as with other apps. Especially since the text may end up being small and buttons difficult to press.

If font-size will become a problem, I think we can tweak what we have in themes, see https://mui.com/material-ui/customization/typography/

> As for breakpoints, MUI has information: https://mui.com/material-ui/customization/breakpoints/

We can use MUI's default breakpoints.

xscottxbrownx:
@andycwilliams @leekahung good thoughts on mobile view, but I suggest we save it for another PR or discuss elsewhere.

As stated, cannot do it on this PR without merging PR #116 and creating a new homepage layout using `<Outlet />` and `outletContext hook` from React Router.  So, this is at a minimum x2-3 PR's out. 

Just busying this PR's comments section with item that cannot be addressed here. 😆

leekahung:
> @andycwilliams @leekahung good thoughts on mobile view, but I suggest we save it for another PR or discuss elsewhere.

Yes, I have commented on that earlier about getting a new PR for those changes. I'll stop digressing.

Issue #119, opened 2023-05-09
Logout Issue: Logging back in on refresh despite explicit logout


leekahung:
**Describe the bug**
Accepting to logout from LogoutModal doesn't clear localStorage as intended from handleLogout.

https://user-images.githubusercontent.com/14917816/237038222-83026448-8bcb-4ece-8af5-b58e203628ac.mov

Was thinking that the handleLogout function wasn't running properly because LogoutButton is a special component for a logout button from Inrupt's Solid React SDK. If called with children, it perhaps needs to move its onClick property to the children.

**To Reproduce**
1. Log into PASS
2. Click on Log Out
3. Click "YES" to logout
4. See error (if I were to refresh here, I'll log me back in as if I never logged out)

https://user-images.githubusercontent.com/14917816/237038879-d5ebdcdf-026d-452d-876a-1aa88fb28205.mov

**Expected behavior**
Expected to clear localStorage to prevent logging back in on refreshing if the user wish to fully logout.

**Proposed Solution**
Thinking of moving handleLogout from LogoutButton to children.

leekahung:
Issue resolved in PR #120. Closing issue.

PR #120, opened 2023-05-09
Issue 119/logout issue

leekahung:
Fixed part of the bug related issue #119 by moving handleLogout function to the children which is an MUI button.

However, this does not remove "loggedIn" stored in the localStorage and is still set to true. When refreshed, App.jsx picks it up and think it's part of a session restore.

So to prevent that, setItem for "loggedIn" is also refactored out from Logout.jsx. It's now part of the useEffect which fetches data from Solid and is triggered when session.info.isLoggedIn is true. This fixes the bug completely.

https://user-images.githubusercontent.com/14917816/237043062-085e2000-49df-4603-8fd6-551422435d21.mov

https://user-images.githubusercontent.com/14917816/237043553-34bbdd65-2c4c-4974-9af8-a2fea4ad4d04.mov

However, you notice, it has not completely clear localStorage as intended by handleLogout from the clips above.

For issueConfig, this is due to the fact it's still using opencommon.net which is hard-coded. I've fixed this using SOLID_IDENTITY_PROVIDER from session for to dynamically pass in the provider.

As for redirectUrl, I've corrected this by including an additional condition to the useRedirectUrl custom hook to trigger only if "loggedIn" exist in localStorage and is set to true.

Now it completely removes all things related only to PASS from localStorage as intended by handleLogout.

https://user-images.githubusercontent.com/14917816/237049299-3e53592b-ad2f-4528-aa0f-fbdecc39ced2.mov

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -78,6 +78,7 @@ const App = () => {
     }
 
     if (session.info.isLoggedIn) {
+      localStorage.setItem('loggedIn', true);
       fetchData();
     }
   }, [session.info.isLoggedIn]);
--- a/src/components/Login/Logout.jsx
+++ b/src/components/Login/Logout.jsx
@@ -10,6 +10,7 @@ import Typography from '@mui/material/Typography';
 import LogoutIcon from '@mui/icons-material/Logout';
 // Custom Component Imports
 import LogoutModal from './LogoutModal';
+import { SOLID_IDENTITY_PROVIDER } from '../../utils';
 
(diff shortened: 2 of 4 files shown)


xscottxbrownx:
Good catch!!! 

This LOOKS great, but I'll let @Jared-Krajewski look at code review to be sure.

Jared-Krajewski (approved):
reviewed in discord voice chat. Good catch!

Issue #121, opened 2023-05-09
Prettier not configured completely


leekahung:
**Describe the bug**
Not a bug per-say with the code, but Prettier doesn't seem to be catching all files in src/ as expected. Was a bit suspect last time when fixing ESLint, but didn't notice it completely then.

**To Reproduce**
1. Run `npm run prettier:check`
2. See error (Some files in src/ seems to be checked for formatting, but not all files)

**Expected behavior**
All files should be processed.

**Proposed Solution**
Add "quotes" around the file name. Like Prettier --write 'src/**/*.js' like what we've done with ESLint.

leekahung:
Issue resolve in PR #122. Closing issue.

PR #122, opened 2023-05-09
Fixing Prettier pathing

leekahung:
Fixing the configuration for Prettier as stated in Issue #121. The pathing previously set does not seem to be correct. I'm expecting this is due to the lack of quote for the file names. In fact, it does seem to use the same syntax as ESLint, something I've thought it doesn't when fixing the pathing for ESLint. Seems to be a miss on my part.

Have given a proper check this time for Prettier and tested it to see if it truly process all .js and .jsx files in src/. After running the tests (see screenshot below), I'm certain they're pathed correctly now.

Screen Shot 2023-05-09 at 2 31 14 (2)

The only file it missed is the types.d.ts, but it's not a .js or a .jsx file. So Prettier is running as intended from the script written in package.json. We could consider updating the script to include .d.ts, but it's only a one-liner at this time and it's used only to help with typing from JSDoc. So I think the fix should be sufficient at this time. Could reconsider in the future.

Again, I'm only fixing the package.json scripts for Prettier in this PR (and making the formatting for ESLint look nicer with single quotes instead of escaped double quotes). Given the number of branches already proposed for merging at the moment, I think we can run ESLint and Prettier after those are included in Development to prevent potential merge conflicts.

I have also tried formatting Prettier exactly like ESLint for 'src/**/*.js[x]'. Unfortunately, when running `npm run prettier:run`, it misses the js files completely and only captured the jsx files (see screenshot below).

Screen Shot 2023-05-09 at 2 57 13 (2)

So I've kept it as is with Prettier calling two file paths recursively instead of one like ESLint.

The extra --write flag was also removed as it seems unnecessary to the script.


Jared-Krajewski (approved):
reviewed in discord voice chat

PR #127, opened 2023-05-10
A few changes to eslint config

timbot1789:
The current eslint config uses a ruleset based off AirBnb's standard distribution, but disables a few rules that will be useful to the application longterm. This PR re-enables a few of them.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,28 +1,37 @@
 module.exports = {
   env: {
     browser: true,
-    commonjs: true,
     es2021: true
   },
-  extends: ['plugin:react/recommended', 'airbnb', 'prettier', 'plugin:jsdoc/recommended'],
   parserOptions: {
+    ecmaVersion: 'latest',
+    sourceType: 'module',
     ecmaFeatures: {
       jsx: true
-    },
-    ecmaVersion: 14
+    }
   },
-  plugins: ['prettier', 'jsdoc'],
+  plugins: [
+    'react',
+    'prettier',
+    'jsdoc'
+  ],
+  extends: [
+    'plugin:react/recommended',
+    'airbnb',
+    'prettier',
+    'plugin:jsdoc/recommended'
+  ],
   rules: {
     'react/function-component-definition': ['error', { namedComponents: 'arrow-function' }],
-    'react/jsx-props-no-spreading': 'off',
-    'react/no-unescaped-entities': 'off',
     'react/prop-types': 'off',
-    'no-console': 'off',
+    'react/jsx-props-no-spreading': 'off',
     'jsx-a11y/label-has-associated-control': ['error', { assert: 'either' }],
-    'arrow-body-style': ['error', 'as-needed'],
     'no-param-reassign': ['error', { props: false }],
-    'jsdoc/valid-types': 'off',
-    'no-alert': 'off',
-    'import/no-extraneous-dependencies': 'off'
+    'no-alert': 'off'
+  },
+  settings: {
+    'jsdoc': {
+      'mode': 'typescript'
+    }
   }
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -25,13 +25,11 @@ module.exports = {
     'react/function-component-definition': ['error', { namedComponents: 'arrow-function' }],
     'react/prop-types': 'off',
     'react/jsx-props-no-spreading': 'off',
-    'jsx-a11y/label-has-associated-control': ['error', { assert: 'either' }],
-    'no-param-reassign': ['error', { props: false }],
     'no-alert': 'off'
   },
   settings: {
-    'jsdoc': {
-      'mode': 'typescript'
+    jsdoc: {
+      mode: 'typescript'
     }
   }
 };
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -32,11 +32,9 @@ const App = () => {
     }
 
     if (restore && localStorage.getItem('loggedIn')) {
-      console.log('restoring session');
       session.login({
         oidcIssuer: SOLID_IDENTITY_PROVIDER,
-        redirectUrl,
-        onError: console.error
+        redirectUrl
       });
     }
   }, [restore]);
--- a/src/components/Form/CheckAclPermsDocContainerForm.jsx
+++ b/src/components/Form/CheckAclPermsDocContainerForm.jsx
@@ -100,7 +100,9 @@ const CheckAclPermsDocContainerForm = () => {
     >
       <form onSubmit={handleAclPermission} autoComplete="off">
         <div style={formRowStyle}>
-          <label htmlFor="set-acl-to">Check permissions to username's Documents container: </label>
+          <label htmlFor="set-acl-to">
+            Check permissions to username&apos;s Documents container:{' '}
+          </label>
           <br />
           <br />
           <input id="set-acl-to" size="60" name="setAclTo" {...user} placeholder={selectedUser} />
--- a/src/components/Form/CrossPodQueryForm.jsx
+++ b/src/components/Form/CrossPodQueryForm.jsx
@@ -74,6 +74,7 @@ const CrossPodQueryForm = () => {
     margin: '20px 0'
   };
 
+  /* eslint-disable jsx-a11y/label-has-associated-control */
   return (
     <FormSection
       title="Cross Pod Search"
@@ -104,6 +105,7 @@ const CrossPodQueryForm = () => {
       </form>
     </FormSection>
   );
+  /* eslint-enable jsx-a11y/label-has-associated-control */
 };
 
 export default CrossPodQueryForm;
--- a/src/components/Form/FetchDocumentForm.jsx
+++ b/src/components/Form/FetchDocumentForm.jsx
@@ -53,6 +53,7 @@ const FetchDocumentForm = () => {
     margin: '20px 0'
   };
 
+  /* eslint-disable jsx-a11y/label-has-associated-control */
   return (
     <FormSection
       title="Search Document"
@@ -71,6 +72,7 @@ const FetchDocumentForm = () => {
       </form>
     </FormSection>
   );
+  /* eslint-enable jsx-a11y/label-has-associated-control */
 };
 
 export default FetchDocumentForm;
--- a/src/components/Form/ManageUsers.jsx
+++ b/src/components/Form/ManageUsers.jsx
@@ -94,6 +94,7 @@ const ManageUsers = () => {
     margin: '20px 0'
   };
 
+  /* eslint-disable jsx-a11y/label-has-associated-control */
   return (
     <FormSection
       title="Manage Users"
@@ -128,6 +129,7 @@ const ManageUsers = () => {
       </form>
     </FormSection>
   );
+  /* eslint-enable jsx-a11y/label-has-associated-control */
 };
 
 export default ManageUsers;
--- a/src/components/Form/ManageUsers.jsx
+++ b/src/components/Form/ManageUsers.jsx
@@ -103,12 +103,12 @@ const ManageUsers = () => {
     >
       <form onSubmit={handleAddUser} style={formRowStyle} autoComplete="off">
         <div>
-          <label htmlFor="add-user-given-name">User's first/given name: </label>
+          <label htmlFor="add-user-given-name">First/given name: </label>
           <input id="add-user-given-name" name="addUserGivenName" {...userGivenName} />{' '}
         </div>
         <br />
         <div>
-          <label htmlFor="add-user-last-name">User's last/family name: </label>
+          <label htmlFor="add-user-last-name">Last/family name: </label>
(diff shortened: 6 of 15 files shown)


.eslintrc.js
```diff
@@ -1,28 +1,27 @@
...
-    'react/prop-types': 'off',
-    'no-console': 'off',
-    'jsx-a11y/label-has-associated-control': ['error', { assert: 'either' }],
-    'arrow-body-style': ['error', 'as-needed'],
-    'no-param-reassign': ['error', { props: false }],
-    'jsdoc/valid-types': 'off',
-    'no-alert': 'off',
-    'import/no-extraneous-dependencies': 'off'
-  }
+    'react/prop-types': 0,
```
timbot1789: Proptypes is a typechecking system for react components, not unlike JSDoc. We don't have it installed in the project. Adding it would be a significant PR, and I don't know if it's actually better than well-designed JSDocs

.eslintrc.js
```diff
@@ -1,28 +1,27 @@
...
+    'react/jsx-props-no-spreading': 0,
```
timbot1789: This rule is generally good to have, since it disallows random props being passed to react components. It can also be used to easily bypass proptypes. However, without proptypes a lot of its advantages are lost. And it will require a significant refactor

.eslintrc.js
```diff
@@ -1,28 +1,27 @@
...
+    'jsx-a11y/label-has-associated-control': 0,
```
timbot1789: This I think is a good rule to have. A11y is an accessibility compliance spec. It checks that the application works well with screenreaders. Currently it fails on a few React forms. I'd recommend a separate PR to enable this.

.eslintrc.js
```diff
@@ -1,28 +1,27 @@
...
+    'no-param-reassign': 0,
```
timbot1789: This is also a valuable thing to have, but I'm working on a separate PR to fix the offending section of code.

.eslintrc.js
```diff
@@ -1,28 +1,27 @@
...
+    'no-alert': 0,
```
timbot1789: Modern UX design generally prefers not to use the browser provided alert system, especially since such alerts are often used for easy dev testing. They can also lead to annoying behavior if several are fired at once. Most UX designers prefer to use application-specific modals.

However, this is an issue for UX, and given the current state of the project, it's probably best to use the built-in alert systems for now.

.eslintrc.js
```diff
@@ -1,28 +1,27 @@
...
+    'comma-dangle': 0
```
timbot1789: This is a weird rule in Airbnb's distribution and I don't like it. It also interferes with Prettier. We need to disable either this rule, or the inverse rule in Prettier.

.eslintrc.js
```diff
@@ -1,28 +1,28 @@
...
-    'no-alert': 'off',
-    'import/no-extraneous-dependencies': 'off'
-  }
+    'react/prop-types': 0,
+    'react/jsx-props-no-spreading': 0,
+    'jsx-a11y/label-has-associated-control': 0,
+    'no-param-reassign': 0,
+    'no-alert': 0,
+    'comma-dangle': 0,
+    'react/function-component-definition': 0
```
timbot1789: This requires all react function components to follow a specific style. It also disallows arrow functions. Airbnb does this because the names of arrow functions are inferred, which can make debugging more difficult. But that's not a common opinion, and not one I think is worth enforcing.

src/App.jsx
```diff
@@ -32,11 +32,9 @@ const App = () => {
     }
 
     if (restore && localStorage.getItem('loggedIn')) {
-      console.log('restoring session');
       session.login({
         oidcIssuer: SOLID_IDENTITY_PROVIDER,
-        redirectUrl,
-        onError: console.error
```
timbot1789, line 39: We shouldn't leave console statements in production code. It pollutes the browser console, slows it down, and may accidentally expose implementation details that we want to keep hidden.

src/components/Form/CheckAclPermsDocContainerForm.jsx
```diff
@@ -100,7 +100,7 @@ const CheckAclPermsDocContainerForm = () => {
     >
       <form onSubmit={handleAclPermission} autoComplete="off">
         <div style={formRowStyle}>
-          <label htmlFor="set-acl-to">Check permissions to username's Documents container: </label>
+          <label htmlFor="set-acl-to">Check permissions to username&apos;s Documents container: </label>
```
timbot1789: The linter recommends escaping apostrophe characters like this.

src/components/Form/ManageUsers.jsx
```diff
@@ -37,7 +37,7 @@ const ManageUsers = () => {
     };
 
     if (!userObject.username) {
-      runNotification(`Operation failed. Reason: No username provided`, 5, state, dispatch);
+      runNotification('Operation failed. Reason: No username provided', 5, state, dispatch);
```
timbot1789: use single quotes if it's a string literal. Only use backticks if the string is interpolated.

benzelcurry (approved):
All of these sound like great changes to me.

.eslintrc.js
```diff
@@ -1,28 +1,28 @@
...
-    },
-    ecmaVersion: 14
+    ecmaVersion: 'latest',
+    sourceType: 'module',
   },
-  plugins: ['prettier', 'jsdoc'],
+  plugins: [
+    'react',
+  ],
   rules: {
```
timbot1789, line 19: Just to clarify here: everything under the rules section are turned off. They are rules included in the airbnb distribution, and this code disables them.

.eslintrc.js
leekahung, line 1: I think we can keep plugins for JSDoc even if we switched to Proptypes for type-checking. It'll help lint for potential errors on JSDoc comments.

src/App.jsx
leekahung: Sound reasonable. I think we can drop this.

src/components/Form/FetchDocumentForm.jsx
```diff
@@ -63,7 +63,7 @@ const FetchDocumentForm = () => {
       <form onSubmit={handleGetDocumentSubmission} autoComplete="off">
         <div style={formRowStyle}>
           <label htmlFor="search-doctype">Select document type to search: </label>
-          <DocumentSelection htmlId="search-doctype" />{' '}
+          <DocumentSelection htmlId="search-doctype" />
```
leekahung: We might leave this to the UI team since it's more related to styling. But without the spacing, the `<select>` element and the `button` element is squashed together. (see below)

Screen Shot 2023-05-11 at 0 58 59

Perhaps we can leave a TODO comment or something related related to DocumentSelection and button in the forms or temporarily leave an `&nbsp;` if not the {' '}

src/components/Form/ManageUsers.jsx
leekahung: Good catch. We can just revert these back to strings instead of string literals.

leekahung (changes requested):
Looks pretty good to me for the most part. Just a few comments. See feedback.

.eslintrc.js
```diff
@@ -1,28 +1,33 @@
 module.exports = {
   env: {
     browser: true,
-    commonjs: true,
```
timbot1789, line 4: We're not using commonjs in this project, best to remove it.

.eslintrc.js
```diff
@@ -1,28 +1,33 @@
...
-    'arrow-body-style': ['error', 'as-needed'],
     'no-param-reassign': ['error', { props: false }],
-    'jsdoc/valid-types': 'off',
-    'no-alert': 'off',
-    'import/no-extraneous-dependencies': 'off'
+    'no-alert': 'off'
+  },
+  settings: {
+    'jsdoc': {
+      'mode': 'typescript'
```
timbot1789: this setting lets us reactivate the `jsdoc/valid-types` rule

.eslintrc.js
timbot1789: Also worth noting, the airbnb version of this is more strict than the default rule. It requires that the label have BOTH an `htmlFor` attribute, and that the control be nested inside the label: https://github.com/airbnb/javascript/issues/1885

src/components/Form/CrossPodQueryForm.jsx
```diff
@@ -74,6 +74,7 @@ const CrossPodQueryForm = () => {
     margin: '20px 0'
   };
 
+  /* eslint-disable jsx-a11y/label-has-associated-control */
```
timbot1789, line 77: Added these disable notes to code sections that break the rule, but require a larger refactor than we can do in this PR. We should follow up in separate PRs and remove these comments.

src/components/Form/CheckAclPermsDocContainerForm.jsx
leekahung: While I'm not really a fan of using HTML entities for special characters (I think we've previously turn off linting rules for that), I'm not sure of a work around for that.

Think this would have to do.

src/components/Form/CrossPodQueryForm.jsx
leekahung: Yeah for this rule we'll probably have to refactor most of the labels with htmlFor.

In the case of label/DocumentSection however I think we can look more into the case where the label and input components are custom components. (https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/label-has-associated-control.md#case-my-label-and-input-components-are-custom-components)

If it's applicable, I think we could try this.

leekahung (changes requested):
Just one more feedback.

src/components/Form/CheckAclPermsDocContainerForm.jsx
timbot1789: Yeah I have mixed feelings about it too. Although I get what they're going for. If you have these regularly (especially > and <) eventually one of them will break something. Most companies avoid this rule because they display strings like this in a separate document that's then imported as a JS constant. They do this so they can more easily translate the string. But we shouldn't worry about implementing something like that in PASS right now.

I can disable that rule again, or try to add an option to allow apostrophes.

src/components/Form/FetchDocumentForm.jsx
timbot1789: I removed these updates, altough really UX should be using CSS for this kind of spacing.

src/components/Form/CrossPodQueryForm.jsx
timbot1789: Corrrect me if I'm wrong, but I don't think that's why the custom components are failing. Most of the labels already have `htmlFor` on them correctly (brownie points to UX for that).

You can see the airbnb config that we're usinghere:
https://github.com/airbnb/javascript/blob/7982931ba745c0f57ba6934fc7e6cc43901dac76/packages/eslint-config-airbnb/rules/react-a11y.js#L123

You can see that they set the `depth` to the max value of 25, so I think that makes eslint search custom components down to a depth level of 25 to find the base `label` and `input` elements.

I think the reason that these are failing is the `assert: both` statement. That means labels need to _both_ have a proper `htmlFor` attribute, and be a parent of their respective `input` elements. Not just one or the other.

AirBnb says this is because that's the only system screenreaders universally support, and I'm inclined to agree. Screenreaders are notorious for not implementing accessibility standards correctly, or implementing them differently. Having a single, somewhat pedantic, style that works for all screenreaders I think is a good move.

src/components/Form/FetchDocumentForm.jsx
benzelcurry: > I removed these updates, altough really UX should be using CSS for this kind of spacing.

I agree. I was a bit confused when I saw manual spacing being added like this, although I know that some Prettier configs will result in these kinds of things being entered into the code when you run the Prettier check. Not sure if that's what happened here or if it was manually entered.

timbot1789: So one of the things I accidentally broke with Ka Hung's eslint file was the prettier configuration. I fixed it now, and these weird curly spaces are gone.

timbot1789: Hmm seems like if there's leftover space at the end of a line in jsx, prettier will escape it like this, since it has syntactic significance in JSX. We probably don't actually want prettier doing this, and instead probably want the jsx whitespace removed. Not sure of the best way to fix it though, since the code is already in development. Maybe something for a followup PR

.eslintrc.js
timbot1789: Yeah my original updates to your file accidentally broke the JSDoc and Prettier integrations. I fixed those now.

src/components/Form/CheckAclPermsDocContainerForm.jsx
leekahung: I think we can leave this rule on.

leekahung (approved):
Changes and review looks good to me. I think we can merge.

PR #137, opened 2023-05-16
Allow users to add cryptographically secure signatures to documents

timbot1789:
## Currently

A case manager has no way to communicate that they have reviewed and approved a certain document upload. This is desirable, as we want to be able to confirm that a certain file is valid and contains all necessary information.

## Changes

Adds a collection of backend utilities through which users can sign and validate documents. Currently, for testing, this system is tied to document upload through a checkbox in the upload form. If the box is checked, the system will generate a signature document from an `RSA-PSS` private key generated by the system for the logged-in user, and stored in the logged-in user's pod.

The current system is as follows:
  - If the user wishes to validate a document, the system will fetch the user's private key stored at `${user_profile}/PASS_Credentials/private_key.ttl` to use to sign the document
    - If no private key is found, PASS will generate a key pair, and store both the private and public key in `PASS_Credentials`
    - The private key is stored in the `private_key.ttl` file, and an `acl` is generated that only gives the user access to the key
    - The public key is stored in `public_key.ttl`, and an `acl` is generated giving public read access to the file
    - PASS tries again to fetch the private key
  -  PASS then uses the private key to generate a `signature.ttl` file with the following format:
    - `accountablePerson` property points to the agent who generated the signature. This is needed to later fetch the public key
    - `url` property points to the related document
    -   `sha256` property points to the signature stored in a Base64 encoded string
  - The `signature.ttl` file is then stored in the same container as the `document.ttl`

There is also a `validateSignature` function in `credentials_helper` that can be used to validate if a signature is authentic, however it is not hooked up to any flow in PASS yet.
   

## Considerations

- This uses the signature algorithm `RSA-PSS` hashed with `SHA-256.` I believe this setup is still secure.
- This uses the browser `WebCrypto` API to generate the keys and signatures, and perform signature validation. All settings for generation are taken from the MDN WebCrypto examples. All major browsers support the WebCrypto API.
- The testable code here is backend only. There's a small integration with the frontend to make testing easier, but an actual frontend integration will be a different PR.
- The code is agnostic to cross-pod vs same-pod uploads. It can be used in either use case.
- The signature is generated for the whole `document.ttl` file. Currently, if we store multiple documents in the dataset represented by the `document.ttl` file, the user currently will need to sign every document, or none of them.

## Limitations
- The current structure places a user's private key on the network a lot. Currently, we generate the private key client side, send it over the network to a server, then fetch it from the server whenever we want to sign a document. This is inherently less secure, and I'm not sure we want to do this long-term. Ideally, we'd generate a private key for the user, have them save it manually, then send a request to their pod whenever we want the user to sign a document. The user would then respond with the document signature (not private key) which we could store. This, however, would require having a server-side signing utility the user has access to.
- The file structure of where things are saved could possibly be improved/reduced


package.json
```diff
@@ -67,6 +67,7 @@
     "eslint-plugin-react-hooks": "^4.6.0",
     "jsdoc": "^4.0.0",
     "jsdoc-tsimport-plugin": "^1.0.5",
+    "jsdom": "^22.0.0",
```
timbot1789, line 71: By default, `vitest` runs tests in a node environment. However, PASS runs in a browser environment. JSDom gives a light-weight browser implementation that can be used in testing. It's not perfect, but it's better than nothing.

src/utils/credentials-helper.js
```diff
@@ -0,0 +1,244 @@
...
+import { Session } from '@inrupt/solid-client-authn-browser';
+
+import { createDocAclForUser } from './session-helper';
+
+import { RDF_PREDICATES } from '../constants';
+
+const generateRsaKeyStrings = async () => {
+  const { privateKey, publicKey } = await window.crypto.subtle.generateKey(
+    {
+      name: 'RSA-PSS',
```
timbot1789, line 21: `RSA-PSS` seems to be the easiest to use signing algorithm available through the `WebCrypto` API. I'm not against using a different algorithm if someone knows more than me.

src/utils/credentials-helper.js
```diff
@@ -0,0 +1,244 @@
...
+  getStringNoLocale
+} from '@inrupt/solid-client';
+import { Session } from '@inrupt/solid-client-authn-browser';
+
+import { createDocAclForUser } from './session-helper';
+
+import { RDF_PREDICATES } from '../constants';
+
+const generateRsaKeyStrings = async () => {
+  const { privateKey, publicKey } = await window.crypto.subtle.generateKey(
```
timbot1789, line 19: The `WebCrypto` API adds a cryptography library usable in the browser:
https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API

src/components/Form/UploadDocumentForm.jsx
```diff
@@ -125,6 +125,15 @@ const UploadDocumentForm = () => {
       defaultMessage="To be uploaded..."
     >
       <form onSubmit={handleFormSubmission} autoComplete="off">
+        <label htmlFor="verify-checkbox">
```
timbot1789, line 128: This is just a default for testing / demonstration. I'm ok with removing it if we have a better idea for the UI.

src/utils/session-helper.js
```diff
@@ -289,8 +293,12 @@ const createFileChecksum = async (fileObject) => {
  * @param {fileObjectType} fileObject - Object containing information about file
  * from form submission (see {@link fileObjectType})
  * @param {string} documentUrl - url of uploaded document or resource
+<<<<<<< HEAD
```
leekahung: Found lines related to merge conflicts from the recent rebasing.

timbot1789: fixed

src/utils/session-core.js
```diff
@@ -181,9 +189,13 @@ export const uploadDocument = async (session, uploadType, fileObject, otherPodUs
   // Place file into Pod container and generate new ttl file for container
   await placeFileInContainer(session, fileObject, containerUrl);
   const newTtlFile = await createResourceTtlFile(fileObject, documentUrl);
+  const signingKey = verifyDocument ? await getUserSigningKey(session) : null;
```
leekahung, line 192: Have tested out the verify file feature. It seems to work out nicely as noted from what's stated from the PR with the private/public keys and the signature.ttl.

Was wondering if we should relegate the initial generation of the keys to when people log in for the first time like what we're doing with the Users container in App.jsx.

I've tried uploading for the first time without PASS Credentials and it was quite slow since it had to generate the keys first before creating a signature for the file for uploading. It eventually finished, but I though the process hanged.

Subsequent uploads after having the keys were much faster. So that's all good.

leekahung (changes requested):
I think it looks good overall. Had tried out the verify file on upload option for upload document and the keys and signature were generated as stated in the PR. I just had one thing to note in this PR. See feedback.

src/utils/session-core.js
timbot1789: I haven't thought too much about the UX for this feature yet. I'm not sure how I feel about doing this on login though. I think the biggest challenge with the signing UX is distinguishing between users who need credentials (case managers who sign things) and users who don't (clients who are uploading their own documents to their own pods). One of the drawbacks of this design is it adds PASS-specific files to a user's pod, and I'd like to avoid adding those files if the user doesn't need them.

I think this is difficult to do on login because login doesn't give us a way to identify what type of user this person is. It just lets us access their pod. We'd need to do a separate fetch to find a document that tells us that.

We could have a separate button for generating credentials on a user profile management page. Then if they attempt to verify a document without credentials, we show a popup that says "Your profile has no associated credentials. Please generate them before signing" and stop verifying the document. That would make it clearer to the user where the slowdowns are occurring.

leekahung: Considering that it only occurs once for when the user first uploads without their keys, I think this would do for now.

We could revisit this again in the future. I'll leave that discretion to you.

src/utils/credentials-helper.js
```diff
@@ -0,0 +1,244 @@
...
+  return key;
+};
+
+/**
+ * Serializes a given Solid DataSet into TTL
+ * Note: Should not be used for networking. Use inrupt's libraries for that
+ *
+ * @memberof utils
+ * @function serializeDataSet
+ * @param {import('@inrupt/solid-client').SolidDataset} dataSet - Inrupt Dataset you want to serialize
```
leekahung, line 144: For types in `@param` we could use imported `@typedef` (see Lines 35 to 45 in session-core.js).

After that you could use just `SolidDataset` as a type.

The type `{Session}` is a good example of this in Line 58 of session-core, the imported `@typedef` could be found in Lines 35 to 37.

src/utils/credentials-helper.js
```diff
@@ -0,0 +1,244 @@
...
+
+/**
+ * Generates a cryptographic signature for a given document
+ *
+ * @memberof utils
+ * @function generateRsaSignature
+ * @param {CryptoKey} key - The RSA-PSS private key to sign the document with
+ * @param {string} data - The data to generate the signature for
+ * @returns {Promise} Promise that resolves to a string containing a Base64 encoding of the signature
+ */
```
leekahung, line 176: Not much of a problem since JSDoc are still just comments at the end of the day, so disregard if you think it's not needed.

For the most part, we have a blank space between the JSDoc comment and the function, would like to keep that consistent.

leekahung (approved):
Have gone through some of the replies and added a few more comments to consider, but not necessary changes. I'm okay with this merging to Development.

PR #131, opened 2023-05-15
Issue 130/search document styling updates

andycwilliams:
This branch is unfinished. Feedback needed.

You'll notice that I also changed how the styling for all forms works. It removes the `.panel` styling from the CSS file. It's pretty light but more consistent with the new MUI theming (with the exception of the `Logout` button, which will likely be moved to the menu anyway). The rest of the CSS file styling still needs to be transferred over to `theme.js`.

The `Get Document` button should work properly.

Switching to the new select option, however, is not as smooth. You can see that there is the previous select option as well as the new MUI version. The MUI version is not yet connected to the actual functionality and so is right now acting as a demonstration.

Before continuing, I wanted to know how we want to approach this. Do we want to rework the way `DocumentSelection.jsx` (and thus `docTypes.js`) works? For instance, MUI has its own method of iterating through options. 

By developing a plan here, styling the remaining components will be simpler.

PR #139, opened 2023-05-17
WIP: Development barcode related to Issue  #79

### LINKED ISSUE #79: Decode license barcodes for verification
Scans for and decodes barcodes from input files to obtain data for verification purposes. Eventually to be implemented into file upload with creation of .ttl file containing the data.

Working barcode scanner from input files.

various open source barcode scanners. chose zxing library because it seemed to be the most popular and well supported.

**Additional context**
this is mainly for proof of concept and is done in a single HTML file(barcode_scanner.html) with the code in a script tag. Working on further organizing the data into a more readable content

tomit4:
:construction: This branch addresses the feature request for integrating a barcode scanner specifically for when the user scans a Driver's License. Using the @zxing/browser library, a properly scanned driver's license will return all essential information from that document and load it into PASS and the solid server. Both the scanned barcode image and the turtle file record this information.

__TODO__:

- [ ] Properly document the recently added code so that it can be easily read from the IDE using jsdoc.
- [ ] Document how to appropriately use the barcode scanner feature as well as what the returned information corresponds to when, say, the Driver's License is scanned and info read at the DMV.
- [ ] Appropriately parse out the essentially needed information for the front end (possibly in a new component/page).


src/assets/barcode.gif
leekahung, line 1: I think we can call this .gif as "demo_barcode.gif".

Also, considering we should allow this type of image format, I think you could update the accept criteria for acceptable upload files in Line 148 of UploadDocumentForm.jsx. We can also change Line 184 in CrossPodWriteForm.jsx while we're at it.

src/constants/rdf_predicates.js
leekahung, line 1: Yeah, we'll probably want to refactor this soon, haha.

src/utils/session-helper.js
```diff
@@ -12,8 +12,9 @@ import {
   saveSolidDatasetAt
 } from '@inrupt/solid-client';
 import sha256 from 'crypto-js/sha256';
+import getDriversLicenseData from './barcode-scan';
 import { RDF_PREDICATES } from '../constants';
-
+// see sandbox/barcode_scanner/rdf_prediates.txt
```
leekahung: Think we can leave this out for the merge since we don't a sandbox/barcode_scanner directory.

src/utils/session-helper.js
```diff
@@ -272,6 +273,48 @@ const createFileChecksum = async (fileObject) => {
...
+  const dlData = await getDriversLicenseData(fileObject.file);
+  return buildThing(createThing({ name: 'document' }))
+    .addDatetime(RDF_PREDICATES.uploadDate, new Date())
+    .addStringNoLocale(RDF_PREDICATES.additionalType, dlData.DCA)
+    .addStringNoLocale(RDF_PREDICATES.conditionsOfAccess, dlData.DCB)
+    .addStringNoLocale(RDF_PREDICATES.expires, dlData.DBA)
+    .addStringNoLocale(RDF_PREDICATES.givenName, dlData.DCS)
+    .addStringNoLocale(RDF_PREDICATES.alternateName, dlData.DAC)
+    .addStringNoLocale(RDF_PREDICATES.familyName, dlData.DAD)
+    .addStringNoLocale(RDF_PREDICATES.dateIssued, dlData.DBD)
```
leekahung: For the dates, is it possible to change the format from MMDDYYYY to a JS Date object or Date string?

Screen Shot 2023-05-17 at 17 20 15

Also I see there are two predicates using identifier. Are there more of a difference between the two. Perhaps we could use a different identifier?

src/utils/session-helper.js
```diff
@@ -282,9 +325,14 @@ const createFileChecksum = async (fileObject) => {
  * @param {string} documentUrl - url of uploaded document or resource
  * @returns {object}
  */
+
 export const createResourceTtlFile = async (fileObject, documentUrl) => {
   const checksum = await createFileChecksum(fileObject);
 
+  if (fileObject.type === 'Drivers License') {
```
leekahung: One of the new PR's going to change the type to "Driver's License". See PR #131 session-helper.js

leekahung (changes requested):
Looking pretty good! Only a few comments. See feedback.

src/utils/session-helper.js
tomit4: This should be possible given JS's native Date() object. I'll look into making this adjustment.

src/utils/session-helper.js
tomit4: alright for sures, easy enough.

src/utils/session-helper.js
tomit4: Lol, yeah this references something totally different. I should have paid closer attention to my git diff. I'll take care of this.

src/assets/barcode.gif
tomit4: very easy change. no problem.

src/constants/rdf_predicates.js
tomit4: This I have less of an intuitive sense of the best way of going about this. Any insights would be helpful.

leekahung: The only thing I can think of at the moment, is something along the line of using `https://schema.org/` as a basis for refactoring. But we can leave this for now in this PR.

src/utils/session-helper.js
tomit4: Okay, this should now be set to now return dates in our preferred format.

src/utils/session-helper.js
```diff
@@ -272,6 +274,48 @@ const createFileChecksum = async (fileObject) => {
...
+ * @param {string} documentUrl - url of uploaded document or resource
+ */
+
+const createDriversLicenseTtlFile = async (fileObject, documentUrl, checksum) => {
+  const dlData = await getDriversLicenseData(fileObject.file);
+  return buildThing(createThing({ name: 'document' }))
+    .addDatetime(RDF_PREDICATES.uploadDate, new Date())
+    .addStringNoLocale(RDF_PREDICATES.additionalType, dlData.DCA)
+    .addStringNoLocale(RDF_PREDICATES.conditionsOfAccess, dlData.DCB)
+    .addStringNoLocale(RDF_PREDICATES.expires, `${formattedDate(dlData.DBA)}`)
```
leekahung: Minor thing, I think we should use .addDate instead of .addStringNoLocale for the formatted dates without time information.

I'm good having them either way.

src/utils/session-helper.js
```diff
@@ -272,6 +274,48 @@ const createFileChecksum = async (fileObject) => {
...
+    .addStringNoLocale(RDF_PREDICATES.givenName, dlData.DCS)
+    .addStringNoLocale(RDF_PREDICATES.alternateName, dlData.DAC)
+    .addStringNoLocale(RDF_PREDICATES.familyName, dlData.DAD)
+    .addStringNoLocale(RDF_PREDICATES.dateIssued, `${formattedDate(dlData.DBD)}`)
+    .addStringNoLocale(RDF_PREDICATES.dateOfBirth, `${formattedDate(dlData.DBB)}`)
+    .addStringNoLocale(RDF_PREDICATES.gender, dlData.DBC)
+    .addStringNoLocale(RDF_PREDICATES.Eye, dlData.DAY)
+    .addStringNoLocale(RDF_PREDICATES.height, dlData.DAU)
```
leekahung: For height, I believe this is in inches. The extra 0 in front is probably for the extremely rare case of people taller than 8ft 4in, but I'm not sure.

In either case, I think we can use .addInteger instead of .addStringNoLocale

leekahung (changes requested):
Just a few more things. See feedback.

For those in session-helper.js, they're more or less recommendations rather than request changes.

leekahung:
Also forgot to mention since #131 has yet to merge into Development, you should change Line 163 in session-core.js to:
```
case: "Driver's License":
```
and Line 13 in form-helper.js to:
```
const docTypes = ['Bank Statement', 'Passport', "Driver's License"];
```

Else, the new function will look for the wrong route and not run properly.

src/utils/session-helper.js
tomit4: Makes total sense. I'll push these changes as well. Thanks for catching these.

src/utils/session-helper.js
tomit4: This seemed like a simple request, but I'm getting that 'Operation not ...' error message ending in '[object Object]'. I've tried coercing the dlData.DAU to a Number(), and even JSON.stringify() just to see if either of those would work with addInteger, but apparently not... I'm not sure.

src/utils/session-helper.js
tomit4: On a second pass look at this. I'm getting a strange message back from solid server when I attempt to load the driver's license image when using the .addDate method instead of the .addStringNoLocale method. It basically says it can't retrieve the data [Object object]. Same as with the height. I've checked the types of the passed data against @inrupt's documentation, this should work as it still expects a string, but perhaps the @inrupt expects the format of the date to be different? That also could be the same as their .addInteger function though that one makes less sense as I tried to coerce those values to Number() and it still didn't work. I'm a bit at a loss.

I'll have to take a closer look at @inrupt's documentation (though it is sparse imho) to get a better understanding.

leekahung: That is certainly strange. At the very least, I was expecting .addInteger to work because it was one of the examples shown in their documentation: https://docs.inrupt.com/developer-tools/javascript/client-libraries/tutorial/read-write-data/#get-the-thing-to-modify

Might have to do a little testing on the side

src/utils/session-helper.js
leekahung: Same feedback as above with .addDate, seems strange that it's not working. I'll do a little testing myself to see what might be the problem.

src/utils/session-helper.js
leekahung: I think I found the issue. It's because .addDate requires you to pass in a Date object, but what we get from `formattedDate` is just the parsed string. To make sure that it works, we need to update the value using:

```
new Date(`${formattedDate(dlData.<3-letter-code>)}`)
```

What we get from the TTL file looks like this:
Screen Shot 2023-05-18 at 2 24 36

Here's the full TTL file.
Screen Shot 2023-05-18 at 2 23 57

The Z at the end of the Dates are part of the Date Object specification for ISO 8601

src/utils/session-helper.js
leekahung: For .addInteger, I seem to get it working with just using Number(dlData.DAU).

Screen Shot 2023-05-18 at 2 24 36

src/utils/session-helper.js
tomit4: Nice find. I'll make those changes on my end and push if it works for me.

src/utils/session-helper.js
tomit4: dang...I swear I tried that, and am not sure....I'll test it again, maybe that was coming from the addDate() then...

src/utils/session-helper.js
tomit4: Oh, do you mind providing a link to the docs where you found this? I'm just curious.

src/utils/session-helper.js
leekahung, line 1: Just want to make sure we catch all the lines with the old 'Drivers License'. I believe you'll update Line 333 in a following commit.

Other lines to update includes Line 163 in session-helper.js and Line 13 in form-helper.js. That should be all of them.

src/utils/session-helper.js
tomit4: Alright, I tested both, and have pushed these new changes. Thanks for pulling the hard work on that one. As I said in the other Date issue, if you could point me to the docs so I can review them for possible issues in the future. Thanks so much!!!

src/utils/session-helper.js
leekahung: For the Date Object specification here's the link: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#date_time_string_format

For .addDate, I've just doubled check the type .addDate requires for their parameters:

Screen Shot 2023-05-18 at 2 38 08

leekahung (changes requested):
See feedback for "Driver's License"

src/utils/session-core.js
```diff
@@ -697,10 +697,10 @@ export const sendMessageTTL = async (session, messageObject, otherPodUsername) =
 
   const newTtlFile = buildThing(createThing({ name: 'document' }))
     .addDatetime('https://schema.org/uploadDate', date)
```
leekahung: Think we got an RDF_PREDICATE for uploadDate, `RDF_PREDICATE.uploadDate`, we're using it in Line 288 for the createDriversLicenseTtlFile function

leekahung (changes requested):
Those should be the final checks I believe. After these changes, I think I'll be ready to approve this. Great work!

src/utils/session-helper.js
tomit4: Got it. That should be all set now.

tomit4: Thanks for providing that code snippet...yet another reason to change to vs code. That didn't come up for me. I'll consider trying a different LSP. And sorry, I should have looked harder at the official @inrupt docs also, they also specify this as well, I just misread it as string. I think I need to slow down when reading documentation in general. Again, thank you.

src/utils/session-core.js
tomit4: Yes, is this not correct? I included the base meta data that is part of the generic createResourceTtlFile() function as well just to have the name, endDate, serialNumber, description, and url as well.

leekahung: No, this is the correct url for that schema. It's just that since we're using RDF_PREDICATES for our schema, I was thinking we could just use that. Whether or not you decide to change this in a new commit, I'm still fine approving what we have here.

src/utils/session-helper.js
leekahung: No worries, we got the problems fixed. 💪

leekahung (approved):
Alright, it looks like everything is more or less in order. I approve this.

src/utils/session-core.js
tomit4: Ah yes, it appears I missed that one when making the changes to that section of the code. Thanks for catching that!

tomit4: I'll be heading to bed in a bit. Nice to see another night owl like myself. If you think this is ready, feel free to pull it in, but if so, I'll make a separate branch/PR for documentation. Otherwise I'll simply start working on the docs on this branch. This will start either tomorrow or on Friday depending on how my other work is going. :peace: And again, thanks for everything!

leekahung:
Resolved merge conflicts with Development.

tomit4:
Oh, I have to admit I'm not familiar with this aspect of development. Merge requests are generally handled by my mentor on the other project. Do you prefer a Commit a merge?

tomit4:
And my apologies, writing prettier --write showed I missed a semicolon.

leekahung:
> Do you prefer a Commit a merge?

Oh, I think you were referring to commit 70ba9fd. That was to deal with merge conflicts that existed between Development and the branch linked to this PR. A merge as a commit could also mean the branch is being rebased or updated if the two are out of sync.

tomit4:
Ah okay, yes I see. Very simple. Looks like Jared will need to review it. I get it. Thank you! Alright, heading to bed. I'll ttyl.

leekahung:
Fixing merge conflicts.

leekahung:
@Jared-Krajewski After this PR (PR #139), I'm thinking of merging Development into Master. That way, the latest changes in Development will be reflected in the GitHub Pages.

Issue #138, opened 2023-05-16
Move Logout/Login into Navbar


xscottxbrownx:
Move Logout and Login functionality into the Navbar component

PR #141, opened 2023-05-17
Issue 138/move logout login into navbar

xscottxbrownx:
**Description**
---
This PR moves the ability to login and logout into the `Navbar`.  
It also deletes the first logout div that was at the top of each route (no longer needed as it's always accessible in `Navbar`.)

**Updates**
---

https://github.com/codeforpdx/PASS/assets/71395931/32d853f7-e949-4ae2-ba56-3c93b329b948

**Related Issue**
---
Closes Issue #138 

**Future Thoughts**
---
- Can refactor the 2 menu components (`NavMenu` and `NavMenuMobile`) into 1. There's only one item difference between the 2 codebases. 
- Left placeholder links for possible `settings`.
- Move route links (`Home`, `Forms`, `Inbox`) into a menu for mobile view as well? On small screens maybe make the link be `Go to... ˅` and when clicked it will show the current route links (`Home`, `Forms`, `Inbox`)?


murphynd:
I think we need a follow-up issue for what the home screen should be when not logged in and for removing the mobile file and making the nav bar one. otherwise looks good.

murphynd (approved):
Approval for the login/logout change.

andycwilliams:
> Approval for the login/logout change.

Please allow the rest of us enough time to review the changes before merging @murphynd

andycwilliams:
I still have comments on this PR.

I like moving the Logout button.

However, I'm not sure it's necessary to move the Login to the Navbar. It seems to me like it would be slightly more inconvenient, especially for those on mobile. There may not be anything else on the landing screen before logging in anyway.

As for moving the route links for a mobile menu, I think it depends on how many tabs we decide we want to implement. If there are only a couple, then it seems unnecessary to add an extra click. We could also possibly move the Inbox into the existing menu. Since we currently have a notifications icon already, we might want to combine those somehow.

Much of this boils down to what the UX team decides, so you're right it's something to consider for the future.

PR #145, opened 2023-05-17
Refactor form submission

timbot1789:
Small refactor to form upload so cross pod uploads and self uploads hit the same code path + have unit tests.

Also fixes an issue with the notifications where they wouldn't display in the proper sequence

### DIFF
--- a/src/components/Form/helpers/FormSubmissionHelper.js
+++ b/src/components/Form/helpers/FormSubmissionHelper.js
@@ -1,0 +1,50 @@
+import { uploadDocument, updateDocument, runNotification } from '../../../utils';
+
+const makeHandleFormSubmission =
+  (uploadType, state, dispatch, session, clearInputFields) =>
+  async (event, crossPodUsername = '') => {
+    event.preventDefault();
+    dispatch({ type: 'SET_PROCESSING' });
+
+    if (!state.file) {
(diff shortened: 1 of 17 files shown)


src/components/Form/helpers/FormSubmissionHelper.test.js
leekahung, line 1: I think we can move helper to utils or some other location. This way we could keep our components separate functionality.

src/components/Form/helpers/FormSubmissionHelper.js
```diff
@@ -0,0 +1,50 @@
+import { uploadDocument, updateDocument, runNotification } from '../../../utils';
+
+const makeHandleFormSubmission =
```
leekahung: Add some basic JSDoc about the function is about and add basic types to the parameters

leekahung (changes requested):
Made a few comments on the PR. See feedback.

src/components/Form/helpers/FormSubmissionHelper.js
timbot1789: fixed

src/components/Form/helpers/FormSubmissionHelper.test.js
timbot1789: fixed

leekahung (approved):
Saw the revisions made. Looks good to me.

PR #148, opened 2023-05-17
Issue 140/render ttl messages

### LINKED ISSUE #132: Create/Regenerate Solid inbox on Login
At the moment, we're using Node Solid Server which has a native inbox set to each new account. However,  I believe Community Solid Server has phased that out. When we finally move onto Community Solid Server, we would need a way to generate a dedicated inbox for users of PASS.

A solution to this is to create a new function that builds an inbox container for the user's Pod on login and triggers only if container doesn't exist. At the moment, we could set the ACL exactly as it is in Node Solid Server, but we can tailor this in the future if needed.

### LINKED ISSUE #140: Pulling messages from Solid inbox and rendering them on PASS
At the moment we've completed the initial step of sending messages as TTLs to other people's pods via their inbox (see PR #128). The next step would be to use that data to render messages in the PASS as oppose to using Solid to read the message.

Create new components within /PASS/inbox/ to render information from message TTLs on Solid. Functions would need to be created for pulling data from Solid to PASS.

leekahung:
This PR will target the rendering of inbox messages from Solid to PASS.

New functionality has been created for pulling TTL files from the user's Solid inbox and passing it to PASS in JSON format. Like with loading users from the user list, it takes time to load the data from Solid, so a loading text is used. Could potentially be replaced by a loading animation like with loading user list.

https://github.com/codeforpdx/PASS/assets/14917816/ec11b6d6-b494-4b68-8ad2-ac0f0a299e5e

New components are created to render the messages.

Screen Shot 2023-05-25 at 0 05 42

The message is displayed/closed by clicking on the box.

https://github.com/codeforpdx/PASS/assets/14917816/4c7f596c-1be4-4260-9376-03ac17b0e01a

Other additions to this PR include updates to NewMessage.jsx to render as a Modal instead of a static form when clicking on New Message in /PASS/inbox/. This will likely be changed in a styling update, but will do for now.

Screen Shot 2023-05-25 at 0 06 39

## Future Considerations

- An outbox to store messages sent by users (See #182)
- Create components for said outbox
- Implement design improvements for inbox/outbox and message components


src/App.jsx
```diff
@@ -41,14 +46,20 @@ const App = () => {
...
+  const initialUserList = [];
+  const [userList, setUserList] = useState(initialUserList);
   const [loadingUsers, setLoadingUsers] = useState(false);
   const [loadingActive, setLoadingActive] = useState(false);
 
   const selectedUserObject = useMemo(() => ({ selectedUser, setSelectedUser }), [selectedUser]);
   const userListObject = useMemo(() => ({ userList, setUserList }), [userList]);
 
+  /** @type {inboxListObject[]} */
+  const initialInboxList = [];
```
xscottxbrownx, line 68: I don't think this variable `initialInboxList` is being used anywhere else (maybe I missed it though.)

If it isn't, I don't think we need it. Just make line 51: `const [userList, setUserList] = useState([]);`

src/components/Inbox/NewMessage.jsx
```diff
@@ -42,6 +46,7 @@ const NewMessage = () => {
     } else {
       try {
         await sendMessageTTL(session, message);
+
```
xscottxbrownx, line 51: **Just a thought...**

she we try to match these "message fields" more with email fields?

Like `Subject`, `To`, and `Content`/`Body`...   Actually I think `Message` works just fine too for the last one.
> well, they all "work", I was just wondering if users would be more familiar/comfortable with different terminology?

---

I think we can cut down the error messages though.  I think we should take out the "value for", so just: `Please enter a Message Title`, etc.   The "value for" makes perfect sense, but is a little more developer speak than typical user speak.

src/contexts/index.js
```diff
@@ -1,5 +1,6 @@
 import SelectUserContext from './selectUserContext';
 import UserListContext from './userListContext';
+import InboxMessageContext from './inboxMessageContext';
 
 /**
  * The contexts moedule contains React Contexts that helps with user management
```
xscottxbrownx: update "module" typo

src/utils/session-core.js
```diff
@@ -699,6 +703,67 @@ export const updateUserActivity = async (session) => {
...
+        const messageThing = messageTTLThing.find((thing) => thing.url.includes('#message'));
+        const message = getStringNoLocale(messageThing, RDF_PREDICATES.message);
+        const title = getStringNoLocale(messageThing, RDF_PREDICATES.title);
+        const uploadDate = getDatetime(messageThing, RDF_PREDICATES.uploadDate);
+
+        // Get data related to #sender
+        const senderThing = messageTTLThing.find((thing) => thing.url.includes('#sender'));
+        const sender = getStringNoLocale(senderThing, RDF_PREDICATES.sender);
+
+        // Get data related to #sender
```
xscottxbrownx: I think `#sender` should be `#recipient`

xscottxbrownx: I'll leave the functionality of this file to be reviewed by somebody more familiar with it (session-core.js) 

I'm getting a little lost in here 😅

I'm unsure about the `catch`es specifically.

xscottxbrownx (changes requested):
In general, the other thing I see consistently in here is the use of styled components instead of the MUI components.  But looks good as a base to start from. 

We can refactor these to be MUI components within the theming pretty quickly.

src/contexts/index.js
leekahung: Fixed.

src/App.jsx
leekahung: It's there for typing purposes since we're not using TypeScript.

With how it's initialized here, we'll get:
Screen Shot 2023-05-19 at 14 37 59

Without it, we'll get:
Screen Shot 2023-05-19 at 14 36 35
With never[] as the type.

src/components/Inbox/NewMessage.jsx
leekahung: I think we could switch to MUI's validation fields to catch errors. Think it'll let us cut back on them.

src/utils/session-core.js
leekahung: The first try/catch tries to see if they could find a user with the username being sent from PASS. If the user exists on the Solid Identity Provider (solidcommunity.net in this case), it'll set recipientName to the user's actual name, else it'll crash out with an error message that state the username does not exist.

Second try/catch tries to see if they could send a message to both the sender's and the recipient's inbox. If the inboxes exist, it'll create a TTL file in both inboxes. If either one of those inboxes does not exist, it'll crash out with an error message that states that at least one of the inboxes does not exist.

xscottxbrownx: @benzelcurry @leekahung thoughts on my first comment here? line 749...

src/components/Inbox/NewMessage.jsx
xscottxbrownx: @benzelcurry @leekahung thoughts on my question here?  Changing `Title` --> `Subject`, `Recipient Username` --> `To`, and maybe leaving `Message`?

Just thinking from a user experience standpoint trying to emulate e-mail as much as possible. the user will be more familiar with it, and that's essentially what this inbox/messaging feature is doing.

src/utils/session-core.js
leekahung: Oh! I see it now. Good catch! You're right,  it supposed to be recipient, not sender on Line 749

leekahung: Corrected this from `#sender` to `#recipient`

src/components/Inbox/NewMessage.jsx
leekahung: I think changing `Title` to `Subject` and `Recipient Username` to `To` makes sense to emulate that email like experience.

leekahung: Keeping message as message, but had done the change stated above.

xscottxbrownx: > Keeping message as message, but had done the change stated above.

Do you have to (for SOLID), or want to, keep the message state as the original keys? 
```js
const [message, setMessage] = useState({
    recipientUsername: '',
    title: '',
    message: ''
  });
  ```
  **Or do you want to change all the code to match new names?**
  
  which would involve change on lines:
  23, 24, 42, 44, 53, 54, 58, 84, and 93
  
  and in `MessagePreview.jsx`, line 25

leekahung: I think we can keep what we have for Solid.  The names match up for the schemas used for the ttl file.

leekahung (commented):
Would likely need to optimize/stylize these more in future issues/PRs, but I believe this should complete this portion of the message system.

Issue #143, opened 2023-05-17
Create dropdown menu for route links when in mobile view (small screens)


xscottxbrownx:
**Current Idea**
---
Move route links (`Home`, `Forms`, `Inbox`) into a menu when viewed from small screen device.

 `Go to... ˅` and when clicked, it will show the current route links (`Home`, `Forms`, `Inbox`.)
 
**MUI Component Idea**
---
https://mui.com/material-ui/react-menu/#customization

xscottxbrownx:
Completed via PR #154

Issue #142, opened 2023-05-17
Refactor NavMenu and NavMenuMobile into only NavMenu


xscottxbrownx:
Currently we have 2 separate components that are part of Navbar:  `NavMenu` and `NavMenuMobile`

They only differ by one element.

**1)** We can do a little conditional rendering on that one element inside `NavMenu` and get rid of `NavMenuMobile`. 
**2)** Will need to refactor `Navbar` slightly to accommodate those changes and continue to work as intended.

xscottxbrownx:
Completed via PR #154

PR #154, opened 2023-05-17
Issue 143/routes dropdown menu

xscottxbrownx:
**Description**
---
This PR creates a dropdown menu for the route links (`Home`, `Forms`, `Inbox`) in the `Navbar` if on a small screen.  
It also adds a backdrop so the menu is easily focused and seen.
Added this same backdrop to the other existing `Navbar` menu (`Profile`, `Settings`, `Logout`.)

**Updates**
---
https://github.com/codeforpdx/PASS/assets/71395931/c0f646fe-172f-4ada-a6d5-989687d41778

**Related Issue**
---
Closes Issue #143 

**Future Thoughts**
---
- Can refactor the 2 menu components (`NavMenu` and `NavMenuMobile`) into 1. There's only one item difference between the 2 codebases. 
- Can refactor `Navbar.jsx` to maybe list everything rendered for mobile navbar then everything rendered for desktop navbar. Currently it lists the elements in the order they display onscreen, so it alternates. 
- Can most likely refactor the states and the functions of opening and closing the menus. I'm sure they can be condensed to just 1 or 2 and reused.


andycwilliams:
It looks nice. Although depending on how many tabs we decide to go with. If we stick with only a couple, this may not be necessary. Having two dropdown menus could end up being a bit unappealing.

We can take a look at those refactoring suggestions.

The UX designs I've seen don't have a large number of tabs. So far I see just My Clients and Documents. Though, of crouse, that can change.

Simplifying as much as possible is best, imo. We want to take as much effort off of the user as possible.

Also, I'm not sure if this is just me, but the screen recordings always freeze after one second. I have to constantly restart it. It's like this no matter who's posting the recording.

src/components/NavBar/NavBar.jsx
leekahung, line 1: Before going through the code, there are styling issue related to either breakpoints or the scrollbar. Think we get have this issue resolve after a bit of digging.

Screen Shot 2023-05-17 at 19 25 07

Screen Shot 2023-05-17 at 19 27 23

Screen Shot 2023-05-17 at 19 25 40

src/components/NavBar/NavBar.jsx
```diff
@@ -90,16 +108,76 @@ const NavBar = () => {
             {session.info.isLoggedIn ? (
               <>
                 <Box>
+                  {/* ------------START----------- */}
```
leekahung: Potential refactor/pull this out into its own component.

src/components/NavBar/NavMenu.jsx
```diff
@@ -43,7 +43,7 @@ const NavMenu = ({ menuId, openMenu, setOpenMenu, anchorEl, setAnchorEl, setShow
       open={openMenu}
       onClose={handleMenuClose}
       onClick={handleMenuClose}
-      sx={{ mt: 5 }}
+      sx={{ mt: 5, backgroundColor: 'rgba(1, 121, 105, 0.2)' }}
     >
       <MenuList>
```
leekahung, line 49: Perhaps combine NavMenu with NavMenuMobile so that it's one component. They're essentially the same. This will be in line with issue #142

leekahung (changes requested):
Gone over some changes in the code and in the application. See feedback.

src/components/NavBar/NavBar.jsx
xscottxbrownx: Complete -> NavbarLinks.jsx

src/components/NavBar/NavMenu.jsx
xscottxbrownx: Complete

xscottxbrownx:
> It looks nice. Although depending on how many tabs we decide to go with. If we stick with only a couple, this may not be necessary. Having two dropdown menus could end up being a bit unappealing.
> 
> We can take a look at those refactoring suggestions.
> 
> The UX designs I've seen don't have a large number of tabs. So far I see just My Clients and Documents. Though, of crouse, that can change.
> 
> Simplifying as much as possible is best, imo. We want to take as much effort off of the user as possible.
> 
> Also, I'm not sure if this is just me, but the screen recordings always freeze after one second. I have to constantly restart it. It's like this no matter who's posting the recording.

Bummer about videos for you, I haven't had any trouble yet.

I did refactor the NavMenu and NavMenuMobile all into the NavMenu component.

Can you take a look to see if you can replicate what Ka Hung is finding with the Navbar and Scrollbar issue?  Most likely only to occur on the FORMS page.   I can't seem to duplicate.  But I'm on mac and seems my scrollbars are small and on top of the content, rather than creating padding to fit them.

src/components/NavBar/NavbarLinks.jsx
leekahung, line 1: This is looking nice! I do like that we're turning the indicator line back on.

One thing I've spotted while testing out the feature.

https://github.com/codeforpdx/PASS/assets/14917816/38ecc47a-cfc4-4299-bae5-828cc50a5339

Most likely, this is due to the fact that the two navigation are decoupled and doesn't share the same routing state.

leekahung (changes requested):
See feedback for additional request change.

src/components/NavBar/NavbarLinks.jsx
xscottxbrownx: I believe this to be fixed now.  Changed it from reading a state to using the current url path location from react router (like the mobile route menu works.)

src/components/NavBar/NavBar.jsx
xscottxbrownx: This is now fixed on small screens. 

**There are some exceptions though**

On `Home` if super narrow screen there will be issue due to the `usersList` component not being mobile responsive yet:
Screenshot 2023-05-19 at 12 33 39 PM

Same thing with the status section of all the forms:
Screenshot 2023-05-19 at 12 34 26 PM

And `Inbox` due to the message button:
Screenshot 2023-05-19 at 12 35 11 PM


We'll need to change a lot of components to be mobile responsive. All the MVP of this project was built by devs on laptops/computers so horizontal screens.

xscottxbrownx: To give proper credit @murphynd found the biggest culprit which was the `<Layout>` component.  Moved it into `App.jsx` per her direction and fixed most of this issue. Rest seem to be component sizes that don't fit in screen size.
🎉🥳🎉

src/components/NavBar/NavbarLinks.jsx
leekahung: Yeah, this definitely solves that issue, great work!

I think this is likely due to how Tabs are handled, but I was wondering if this delay in indicator moving be fixed?

https://github.com/codeforpdx/PASS/assets/14917816/28a79d66-e115-4659-a849-ce3be184d9c2

xscottxbrownx: Nothing easy...

There's this action prop that can be passed, looks like may be able to tell the indicator to update - but I don't believe that will do anything for the speed:
Screenshot 2023-05-19 at 2 42 12 PM


There's also the MUI class of `MuiTabs-indicator` that may be able to adjust some transition speed?  The issue is I can't find anything on what styles are already applied to this class (so I know how to adjust it.)
Screenshot 2023-05-19 at 2 42 33 PM

Screenshot 2023-05-19 at 2 42 24 PM

murphynd (changes requested):
Testing automation

src/components/NavBar/NavbarLinks.jsx
leekahung: While I do like having the line indicator, we could either drop it until a solution is found or leave it as is. It does go to the correct section, it just scrolls to it instead of staying or jumping directly to it.

leekahung, line 1: I've noticed that an initial Tab is not set by default.

https://github.com/codeforpdx/PASS/assets/14917816/dd20b8ba-0b16-47d9-a6b4-4733b8467c55

As a result of the tab not initializing, MUI seems to be complaining.

Screen Shot 2023-05-19 at 16 42 58

I would have to click it in order to get it initialized.

https://github.com/codeforpdx/PASS/assets/14917816/48bf83d0-e4e1-45c5-8226-de32465523cd

After that, MUI seems happy again and no longer give me the error messages.

Screen Shot 2023-05-19 at 16 46 38

leekahung (changes requested):
Didn't expect so many quirks with the Tabs and MenuItems. Unfortunately, I've noticed one more. See feedback.

src/components/NavBar/NavbarLinks.jsx
benzelcurry: Just went through and gave this a look over and this was the main thing I noticed as well.

benzelcurry:
I think the dropdown menus look great. The only issue I really came across was the one that Ka Hung just recently mentioned.

My main concern is the same as Andy's. I believe that having two dropdown menus is a bit unappealing; hopefully we'll have a small enough amount of tabs that this isn't something which will be required. I do like that you made it so the navigation dropdown explicitly states "Navigate to..." but I do think that if we could avoid making this a dropdown at all, that would be ideal.

benzelcurry (approved):
Tentative approval pending the changes that Ka Hung already requested.

src/components/NavBar/NavbarLinks.jsx
xscottxbrownx: Hmmmm... okay it's because when you click login for some reason the `useLocation().pathname,slice(6)` = 'home/'

when every other time it's just 'home' no ending '/'

Screenshot 2023-05-19 at 5 45 30 PM

compared to:

Screenshot 2023-05-19 at 5 45 44 PM
Screenshot 2023-05-19 at 5 45 50 PM

xscottxbrownx: Could also completely change it to be something more like this ->


https://github.com/codeforpdx/PASS/assets/71395931/3899af7b-ebca-4177-a0e9-d07e58847e3a

Of course that doesn't solve the tabs situation, but if you all aren't liking the double dropdowns right next to each other.  And I agree ultimately maybe it won't be needed at all depending on how many routes we end up with.

xscottxbrownx: My personal vote is leave as is, and can be it's own issue to see if there is a fix using one of the two options I posted above.

leekahung (approved):
Alright, the changes looks good to me. Of course, the sliding indicator line is still a bit quirky, but it's relatively minor at the moment and the end result is working as intended (line jumps to inbox if refreshed from inbox)

murphynd (approved):
Looks good. The scope of this ticket has been addressed. I am also interested in UX/UI or us devs looking at some bars that handle this amount of info and making another issue to redo the layout.

Issue #144, opened 2023-05-17
Enhance design and UX of InactivityNotification


benzelcurry:
### Enhancement

Improve the design and UX of InactivityNotification by blurring and preventing interaction with the background of the website while the modal is open, along with modifying the button color to match the rest of the website.

xscottxbrownx:
@benzelcurry , while working on this issue...

If you want to look at `LogoutModal.jsx` for the MUI component inspiration,

OR 

if you find a better design/implementation/component, we can refactor that one to match.

xscottxbrownx:
Completed with PR #166

PR #166, opened 2023-05-19
Issue 144/enhance design and ux of inactivity notification

benzelcurry:
# Description 

This pull request enhances the InactivityNotification component by blurring and preventing interaction with everything in PASS besides the notification itself until the notification has been closed. Additionally, the color of the logout button within the notification was adjusted to match the new color scheme of PASS.

# Screenshots 
image

# Relevant Issue
#144  

# Future Considerations
* N/A


src/components/Notification/InactivityMessage.jsx
```diff
@@ -4,8 +4,7 @@ import React, { useState, useEffect } from 'react';
 import { LogoutButton } from '@inrupt/solid-ui-react';
 // Styling Imports
 import styled from 'styled-components';
-// Material UI Base Imports; TODO: Update imports to @mui/material when re-styling
-import Modal from '@mui/base/Modal';
+// MUI base UI imports; TODO: Update imports to @mui/material when re-styling
```
leekahung: Use this phrasing instead "// Material UI Base Imports; TODO: Update imports to @mui/material when re-styling"

src/components/Notification/InactivityMessage.jsx
```diff
@@ -69,21 +68,34 @@ const InactivityMessage = () => {
...
-        <ButtonsContainer>
-          <StyledButton onClick={() => setShowPopup(false)}>Continue Session</StyledButton>
-          <LogoutButton>
-            <StyledLogoutButton onClick={handleLogout}>Log Out</StyledLogoutButton>
-          </LogoutButton>
-        </ButtonsContainer>
-      </StyledContainer>
-    </StyledModal>
-  );
+  return showPopup && activeUser ? (
```
leekahung: This is great already. I was wondering if there's a simpler alternative to this using just &&.

leekahung (changes requested):
See feedback. This is good. Think I could add this to Development before merging to Master.

xscottxbrownx:
**This looks good to merge.**  
There are a few things I will list here to think about or address when we go back and refactor this into MUI component.

---

Screenshot 2023-05-19 at 9 43 21 PM

This whole section could be refactored to be more DRY, but not sure it should be. It would actually be more lines of code and honestly less readable. But something like this:
```js
const eventTypes = ['mousedown', 'mousemove', 'touchstart'];

eventTypes.forEach(eventType => {
  window.addEventListener(eventType, handleUserActivity);
});

return () => {
  clearTimeout(timer);
  eventTypes.forEach(eventType => {
    window.removeEventListener(eventType, handleUserActivity);
  });
};

const handleLogout = () => {
  const keysToRemove = [
    'loggedIn',
    'redirectUrl',
    'restorePath',
    `issuerConfig:${SOLID_IDENTITY_PROVIDER}`,
    `issuerConfig:${SOLID_IDENTITY_PROVIDER.slice(0, -1)}`
  ];

  keysToRemove.forEach(key => {
    localStorage.removeItem(key);
  });
};
```
---

I've been playing with a green overlay (you can see on the navbar menu's overlays.) Not sure if we should do that, but if we decided we like it - would need to mimic here and on the logout confirmation modal overlay as well.

---

I also like the idea of icons in the buttons.  Not sure if it should be the logout icon from Navbar (probably), or a generic X or checkmark like the LogoutConfirmationModal?

leekahung:
> I also like the idea of icons in the buttons. Not sure if it should be the logout icon from Navbar (probably), or a generic X or checkmark like the LogoutConfirmationModal?

I do like that idea. Think we can do that in a UI update.

leekahung (approved):
Looks good as is now. Could look more into the icons and other UI changes in a future PR.

PR #157, opened 2023-05-19
Merging Development into Master  (Updating from v0.0.3 to v0.0.4)

leekahung:
This merge will incorporate the following changes listed from CHANGELOG.md as v0.0.4 in Master. It's been a while since our last update with version 0.0.3 (April 25th, 2023), but lots of new features been added. See below.

# Visual Changes

Login:
Screen Shot 2023-05-22 at 12 10 50 (2)

Home:
Screen Shot 2023-05-22 at 12 11 07 (2)

Logout:
Screen Shot 2023-05-22 at 12 11 14 (2)
Screen Shot 2023-05-22 at 12 12 01 (2)

Forms:
Screen Shot 2023-05-22 at 12 11 26 (2)

Inbox:
Screen Shot 2023-05-22 at 12 11 33 (2)
Screen Shot 2023-05-22 at 12 11 34 (2)

# Direct updates in PR #157:

- Fixed small bug for state.verifyFile. The dispatch for "TOGGLE_VERIFY_FILE" works as intended for toggling the check box for state.verifyFile, but state.verifyFile does not reset after form submission (i.e. state.verifyFile remains true, even though checkbox gets cleared). Thus the toggling after the first upload gives the opposite value for state.verifyFile. Created "CLEAR_VERIFY_FILE" to make sure state.verifyFile is always reset to false after submission.
- Fixed routing for cross-pod uploads with new "Driver's License" as fileObject.type. For cross-pod upload, the url will include the apostrophe with the URL. To remove it during cross-pod uploads, we've included a replace function to remove it.
- Fixed "\r" from appearing at the end of OIDUrl. Not sure how it ended up there, it never occurred previously during development and earlier during the review process. We thought it might be due to me adding blank lines at the end of .env files, but it shouldn't have affected it. So while this remains a mystery, SOLID_IDENTITY_PROVIDER is working again as intended.
- Fixed typo in "theme.js" where two of the hex color codes where mistyped with two hashes instead of just one.

# CHANGELOG (v0.0.4 - May 21, 2023)

## Features

- Upload Driver's License into Solid through a barcode scanner (#139)
- Included cryptographically secure signatures to documents (#137)
- Included new Inbox for PASS to allow users to send messages as TTL files (#128)
- Included checksum for uploaded files on Solid Pods via PASS (#123)
- Major UI updates with Material UI, including NavBar, Footer, Login, LogoutModal, etc. (#107, #111, #116, #118, #141, #154, #166)
- Created new Footer component for PASS (#118)
[...]

PR #167, opened 2023-05-20
allow user to choose identity provider on signin

### LINKED ISSUE #168: Update Users feature to use WebIds instead of username
Currently, the Users feature works using by tying usernames (timbot, jfelt, hunter2) to User information created by the case manager (first name, last name, etc). There is also logic to use these usernames to access the user's pod and interact with them. However, usernames are not what are used in Solid to identify pods owned by an individual; Solid uses WebIds (https://timbot.solidcommunity.net/profile/card#me, https://www.timsBlog/solidPod/card#me) to identify individuals.

The code for the users feature currently attempts to take the username and interpolate it into a WebId when it tries to access a user's pod. However, this only works if the user's real WebId is in the format the code expects. There's no requirement for WebId formats beyond that they be a valid URI. If the WebId is actually a different format, the code will fail. 

We can avoid a lot of these issues by storing the user's WebId instead of a username.

The WebId is harder to input than a simple username, however this can be mitigated with some clever UI.
We can auto-add the protocol (`https://`) so the case manager doesn't need to put that in. Additionally, we could provide templates for the WebId formats of the most popular OIDC providers (the current template works in probably 90% of cases, but we'd like the system to work in 100% of cases)

**Additional context**
Currently, we're also storing some user information in the user system. However, the WebId is intended to point to a public`profile` document. It's possible that instead of storing a large amount of info on each individual user, PASS could fetch the user's personal profile and display it to the case manager. This could reduce the amount of information we'd need to store in the PASS system.

timbot1789:
Adds an input field to PASS login that allows users to sign into a different identity provider if they so desire. By default, it loads with the `SOLID_IDENTITY_PROVIDER` configured in the enviroment. Also updates the login refresh logic to handle the new provider.

image

image


src/components/NavBar/NavBar.jsx
```diff
@@ -61,7 +57,7 @@ const NavBar = () => {
 
   // Event handler for logging out of SOLID POD and removing items from localStorage
   const handleLogout = () => {
-    removeKeys();
+    localStorage.clear();
```
timbot1789, line 60: Luckily for us, this `removeKeys` function is built into browsers, so we don't need to build it ourselves.
`localStorage.clear()` removes ALL data we've put in localStorage, which will work fine as long as we use `localStorage` just for info that lets us preserve login between sessions, which we're doing now. If we need to store info across different active logins we have a few options:

1. Store it clientside in [indexeddb](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API)
2. Store it serverside in a separate config file that we fetch on login.
3. Brainstorm a few other things too

These all have different advantages and tradeoffs.

leekahung: I had thought about a full clear before with localStorage.clear() before, but I've headed towards just clearing those generated by PASS.

leekahung:
Not exactly the same as PR #169/Issue #168, but related. Linking issue #168 to this PR.

leekahung:
This is a really small PR, but I'm going to hold this for now. Will approve after PR #157 is completed.

package.json
```diff
@@ -52,6 +52,8 @@
   ],
   "devDependencies": {
     "@solid/community-server": "^5.1.0",
+    "@testing-library/react": "^14.0.0",
```
timbot1789, line 56: Add some libraries to test behavior of react components. React Testing Library is the one recommended by React MUI. React Testing Library runs code on the DOM level. I.E. it renders everything down to the DOM document sent to users, completely removing react, and making tests interact directly with HTML elements.

The other UI testing library I've used is enzyme, which keeps things at the react component level. Tests can then interact more easily with the React components themselves.

IMO, React Testing Library takes a better approach, since it forces the tests closer to what the user will actually see.

package.json
```diff
@@ -52,6 +52,8 @@
...
+    "@testing-library/user-event": "^14.4.3",
```
timbot1789, line 57: A recommended utility library on top of React Testing Library. Provides a lot of functions that simulate how a user actually interacts with a system. E.G. `user.type(input,'hello')` simulates the user mousing over the `input` element, clicking on it, then typing h-e-l-l-o on their keyboard.

test/components/NavBar/OidcLoginComponent.test.jsx
```diff
@@ -0,0 +1,32 @@
...
+});
+
+it('renders correctly', () => {
+  const { container } = render(<OidcLoginComponent />);
+  expect(container).toMatchSnapshot();
+});
+
+it('sets OIDC provider on login', async () => {
+  vi.mock('@inrupt/solid-ui-react', () => ({
+    LoginButton: ({ children }) => <div>{children}</div>
```
timbot1789, line 19: Mock the Inrupt Login button so it doesn't try to fetch a session that doesn't exist. We don't care to test Inrupt's code here.

test/components/NavBar/OidcLoginComponent.test.jsx
```diff
@@ -0,0 +1,32 @@
+import { render, cleanup } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React from 'react';
+import { expect, it, vi, afterEach } from 'vitest';
+import OidcLoginComponent from '../../../src/components/NavBar/OidcLoginComponent';
+
+afterEach(() => {
+  vi.clearAllMocks();
+  cleanup();
```
timbot1789, line 9: `@testing-library/react` adds all the `render` calls to a single shared document object. Without this call, renders between tests leak, which is bad.

src/components/NavBar/OidcLoginComponent.jsx
```diff
@@ -0,0 +1,54 @@
...
+import { useRedirectUrl } from '../../hooks';
+
+const OidcLoginComponent = () => {
+  const [oidcIssuer, setOidcIssuer] = useState(SOLID_IDENTITY_PROVIDER);
+  const redirectUrl = useRedirectUrl();
+
+  return (
+    <>
+      <Box sx={{ flexGrow: 1 }} />
+      <Box sx={{ flexGrow: 1 }} />
```
leekahung: I believe these Boxes are here to help keep the components for TextField and the LoginButton on the right.

I think 1 Box should be enough since they're already using flexGrow: 1. The first box should fill out the space.

src/components/NavBar/OidcLoginComponent.jsx
```diff
@@ -0,0 +1,54 @@
...
+      <TextField
```
leekahung, line 18: TextField is off centered vertically, but that's a minor thing. Think we can update that in a future PR for styling.

leekahung (changes requested):
Think this looks pretty good. Just a small change. See feedback.

src/components/NavBar/OidcLoginComponent.jsx
timbot1789: Fixed

src/components/NavBar/OidcLoginComponent.jsx
timbot1789: Fixed

leekahung (approved):
Alright, this looks good to me.

PR #169, opened 2023-05-20
Issue 168/convert users to webids

### LINKED ISSUE #168: Update Users feature to use WebIds instead of username

timbot1789:
Draft PR. Not ready for review

A reworking of our backend for user list and user activity to get users to work with both WebIds and Community Solid Server, as well as improve the performance of them.

This PR does not include any UI changes except the addition of a WebId field in the Manage Users form, and a few updates to text to make it clearer. The manage users form will be revamped in the next PR of this sequence.

Major changes:

- Users all now have an additional field, `webId`, containing their OIDC webId. This information was originally (and inaccurately) stored in a `podUrl` field
- The `podUrl` field now stores the user's preferred pod url, not their web Id
- The functionality behind user list fetching and user activity fetching has been fully rewritten
  - User List functionality has been moved from `session-core` to `model-helpers/UserList.js`
  - User Activity functionality has been moved from `session-core` and `session-helper` to `model-helpers/User.js`
  - Management of the User List has been moved to the User List Context. All edits to it pass through there. The User List Context currently provides 2 hooks that allow components to interact with the list: `addUser` and `removeUser`.
  - All logic around loading users and the user list has been moved to the user list context. The loading state of the user list is exposed from that context as well.
- The `activity.ttl.acl` is updated to give public read-only access to the file. This is to keep it functional on community solid server, which does not provide a `public` folder out of the box.


src/components/Form/ManageUsers.jsx
```diff
@@ -24,118 +22,169 @@ import FormSection from './FormSection';
...
 
-    if (!userObject.username) {
-      runNotification(`Operation failed. Reason: No username provided`, 5, state, dispatch);
-      setTimeout(() => {
-        dispatch({ type: 'CLEAR_PROCESSING' });
-      }, 3000);
-      return;
-    }
-
-    if (!userObject.givenName) {
```
timbot1789, line 53: I think it's best to remove these object validations. There are jokes about how difficult it is to validate that a person's name has the correct format. Often the best thing to do is to take down whatever name they give you (regardless of its structure) and then assign them a unique ID that the system uses to separate them from others.

Also if we need robust data validations in the system, we'll probably need to set up a more object-oriented approach than this anyway. E.g. we have validations functions on the `userObject` that we call into. E.G.

```
if (!userObject.isValid()) {
  runNotification(`Operation Failed: ${userObject.getValidationErrors}`)
}
```

src/components/Form/ManageUsers.jsx
```diff
@@ -24,118 +22,169 @@ import FormSection from './FormSection';
...
+    dispatch
+  );
+
+  clearProcessing(dispatch);
+
+  await listUsers;
+};
+
+const renderWebId = (username) => {
+  const template = ['https://', '.solidcommunity.net/profile/card#me'];
```
timbot1789: Want to make this template structure more robust. We can have a few common formats providers use that users can select, so we can reliably generate the correct Web Id using just the username.

src/components/Form/ManageUsers.jsx
```diff
@@ -24,118 +22,169 @@ import FormSection from './FormSection';
...
-        dispatch({ type: 'CLEAR_PROCESSING' });
-      }, 3000);
-      return;
-    }
-
-    let listUsers = await addUserToPod(session, userObject);
-    listUsers = await getUserListActivity(session, listUsers);
+    const listUsers = await sendUserToSolid(userObject, dispatch, session, state);
 
     setUserList(listUsers);
```
leekahung: We could probably pass in setUserList as a parameter to the new sendUserToSolid function in Line 50 and set the state for userList in there.

So In Line 76, we'll have `setUserList(listUsers)` instead of `await listUsers` and we can change Line 104 directly to `await sendUserToSolid(userObject, dispatch, session, state, setUserList);`. This seems to solve the issue of users not appear after being added on Solid.

Changing those 3 lines seems to fix the issue.

src/components/Form/ManageUsers.jsx
```diff
@@ -1,14 +1,12 @@
...
-import {
-  runNotification,
-  addUserToPod,
-  getUserListActivity,
-  SOLID_IDENTITY_PROVIDER
-} from '../../utils';
+import { runNotification, addUserToPod, getUserListActivity } from '../../utils';
+
 // Custom Hook Imports
 import { useStatusNotification, useField } from '../../hooks';
```
leekahung, line 10: I think it's time we can abandon the old useField custom hook with the useState hook directly now that we're transitioning to MUI. The source of the MUI uncontrolled input errors seems to be coming from there.

src/components/Form/ManageUsers.jsx
```diff
@@ -24,118 +22,169 @@ import FormSection from './FormSection';
...
+
+const renderWebId = (username) => {
+  const template = ['https://', '.solidcommunity.net/profile/card#me'];
+  return `${template[0]}${username}${template[1]}`;
+};
+
 const ManageUsers = () => {
   const { session } = useSession();
   const { state, dispatch } = useStatusNotification();
   const { clearValue: clearUserGivenName, ...userGivenName } = useField('text');
```
leekahung, line 33: As mentioned before, the MUI errors seems to stem from these useField custom hooks. We can replace with with more direct useState hooks for givenName and familyName like with username and webId.

We could also add handler functions for those to pass direct into the TextFields for first-name-form and last-name-form with onChange and set their value to givenName and familyName, respectively. That strategy seems to work great, when I tested it out. (see video)

https://github.com/codeforpdx/PASS/assets/14917816/c37f59b4-c24d-4df8-b17f-e31abbcf63be

src/components/Form/ManageUsers.jsx
```diff
@@ -24,118 +22,169 @@ import FormSection from './FormSection';
...
   const { clearValue: clearUserFamilyName, ...userFamilyName } = useField('text');
-  const { clearValue: clearUsername, ...username } = useField('text');
+  const [username, setUsername] = useState();
```
leekahung: Missing an initial state inside useState(). An empty string `''` should do.

src/components/Form/ManageUsers.jsx
leekahung: I'm not sure at the moment, what we could do to replace the name.

src/components/Form/ManageUsers.jsx
```diff
@@ -24,118 +22,169 @@ import FormSection from './FormSection';
...
       state={state}
       statusType="Status"
       defaultMessage="To be added..."
     >
       <form onSubmit={handleAddUser} style={formRowStyle} autoComplete="off">
-        <div>
-          <label htmlFor="add-user-given-name">First/given name: </label>
-          <input id="add-user-given-name" name="addUserGivenName" {...userGivenName} />{' '}
-        </div>
+        <TextField
```
leekahung: We could wrap these TextFields with FormControl components. This could also mitigate MUI errors for uncontrolled inputs. FormControl is helpful since we could then wrap TextFields with FormHelperText components. They'll look like this at the bottom below the TextField:

Screen Shot 2023-05-20 at 17 04 54

You could replace that text on the bottom with something else.

leekahung (changes requested):
Hey Tim, this is looking really nice at the moment. I love the fact that you could really customize the Solid Provider if users wish to unlock the webId.

~~But I think there could be a better way with the Provider, I think. We could simply let users edit that instead of the full webId since I'm half-expecting them to already be using the correct username when filling out the initial form.~~

For displaying the full webId, I think we can use MUI's FormHelperText: https://mui.com/material-ui/api/form-helper-text/

The text on the bottom appears if you wrap both TextField and FormHelperText around a FormControl Component:

Screen Shot 2023-05-20 at 17 04 54

Updates: Could certainly move in the direction we're heading now with these changes. I had the wrong interpretation of them. Otherwise, I think the helper texts could still be nice.

timbot1789:
> Hey Tim, this is looking really nice at the moment. I love the fact that you could really customize the Solid Provider if users wish to unlock the webId.
> 
> But I think there could be a better way with the Provider, I think. We could simply let users edit that instead of the full webId since I'm half-expecting them to already be using the correct username when filling out the initial form.
> 
> For displaying the full webId, I think we can use MUI's FormHelperText: https://mui.com/material-ui/api/form-helper-text/
> 
> The text on the bottom appears if you wrap both TextField and FormHelperText around a FormControl Component:
> 
> Screen Shot 2023-05-20 at 17 04 54

I'm not sure what you mean by this:

> I'm half-expecting them to already be using the correct username when filling out the initial form.

The issue with usernames isn't just that a user can have a different provider. It's that different providers have different WebId formats. There's no requirement for a webId to follow any specific format, other than it being a URL.

e.g. If a user with username `tim` signs up for a pod at solidcommunity.net, they get a webIds like `tim.solidcommunity.net/profile/card#me` however, if `tim` creates a pod on the community solid server build we have in the repo, they get the format `localhost:3000/tim/profile/card#me`. By default, any provider using community solid server will have that format. a site pods.com will have the format `pods.com/tim/profile/card#me`

To put it another way, we can't use the Users feature with the dev server we have in the project repo because the dev pod generates WebIds in a different format. Simply switching out the provider doesn't fix that. That's what actually motivated my desire to change this. I'm just trying to make a fix that's robust.

src/components/Form/ManageUsers.jsx
timbot1789: Yeah that sounds like a good idea

src/components/Form/ManageUsers.jsx
timbot1789: Cool! I can remove it from the component then.

src/components/Form/ManageUsers.jsx
timbot1789: TextField actually already uses `FormControl` [components internally](https://mui.com/material-ui/api/text-field/#advanced-configuration). That helper text can be added to a TextField using the `helperText` parameter ([link to MUI demo's code](https://github.com/mui/material-ui/blob/0c58832b6e182f782e73b118ed9681055db3e249/docs/data/material/components/text-fields/FormPropsTextFields.tsx#LL55C33-L55C33). There may be some value in adding a `FormControl` into the `FormSection` just so we don't need to think about it, or adding helper text to the top of this whole form, but we can leave that for another PR.

timbot1789: I also think those errors were being thrown because we weren't setting the `value` and `onChange` properties, like you noted, not because of the FormControl

leekahung:
> The issue with usernames isn't just that a user can have a different provider. It's that different providers have different WebId formats. There's no requirement for a webId to follow any specific format, other than it being a URL.
> 
> e.g. If a user with username `tim` signs up for a pod at solidcommunity.net, they get a webIds like `tim.solidcommunity.net/profile/card#me` however, if `tim` creates a pod on the community solid server build we have in the repo, they get the format `localhost:3000/tim/profile/card#me`. By default, any provider using community solid server will have that format. a site pods.com will have the format `pods.com/tim/profile/card#me`
> 
> To put it another way, we can't use the Users feature with the dev server we have in the project repo because the dev pod generates WebIds in a different format. Simply switching out the provider doesn't fix that.

I guess I was under the wrong impression of how webIds are generated since that's how opencommons.net and solidcommunity.net does their webIds.

I thought it was a kind of standardization of webIds, but you're right. I don't really see anything about formatting in the latest specification. (https://solid.github.io/webid-profile/)

I'm certainly fine with what we have now then since it will be difficult to tell how the different servers would provide their webIds.

src/components/Form/ManageUsers.jsx
leekahung: Ah okay. I think I know what you're referring to.

I just read a bit more about it in the MUI documentation about controlled vs uncontrolled. https://mui.com/material-ui/react-text-field/#uncontrolled-vs-controlled

src/components/Form/ManageUsers.jsx
leekahung: I think we could try the format that was given as a default from Community Solid Server, so we'll get urls with `https://<servername>.com/`, "/profile/card#me".

src/components/Form/ManageUsers.jsx
```diff
@@ -24,118 +22,168 @@ import FormSection from './FormSection';
...
+          InputProps={{
+            endAdornment: (
+              <InputAdornment position="end">
+                <WebIdEditButton
+                  lockStatus={userEditingWebId}
+                  setLockStatus={setUserEditingWebId}
+                />
+              </InputAdornment>
+            ),
+            style: { width: `${webId.length > 40 ? `${webId.length * 9}px` : '300px'}` },
```
timbot1789: Resize WebId input field to always show full webId

src/components/Form/ManageUsers.jsx
timbot1789: fixed

src/components/Form/ManageUsers.jsx
timbot1789: I'm not exactly sure what you mean in your last paragraph, but replacing the `useField` with `useState` removed the errors and explicitly controlling the TextField components fixed the error.

src/components/Form/ManageUsers.jsx
timbot1789: Before we merge this, I want to add a few more templates (like the one you mentioned). By default the system will select the one associated with the `SOLID_IDENTITY_PROVIDER` configured in the env. The user can select a different template from a dropdown menu (pending) if they want. We can even configure the the dropdown menu to save an individual user's preferred template if it's different from the environment.

We could also add a Solid provider field (like you suggested) to make generating weird WebIds even easier.

Ideally, this PR won't cause any actual UX changes. A case manager will still create a user in the same way:  add first name, last name, username, then click `Add User`. PASS can generate the WebId in the background. The WebId field should only be a fallback if the Pass system can't figure out what the actual WebId should be.

leekahung (commented):
Alright, I think this is fine as is. There are some cosmetic things I'm thinking about like whether we should keep ManageUsers as a static form or as a Modal, but that could be saved for later.

Let me know if you think the draft is ready.

leekahung (commented):
My mistake. It was not suppose to be a request change, it was just meant to be a comment.

timbot1789:
> Alright, I think this is fine as is. There are some cosmetic things I'm thinking about like whether we should keep ManageUsers as a static form or as a Modal, but that could be saved for later.
> 
> Let me know if you think the draft is ready.

I still need to add unit tests to it. Once I do that it should be ready.

src/App.jsx
```diff
@@ -65,15 +68,19 @@ const App = () => {
      * @function fetchData
      */
     async function fetchData() {
-      await generateUsersList(session);
-      await generateActivityTTL(session);
-      await updateUserActivity(session);
-      await createDocumentContainer(session);
+      setLoadingUsers(true);
+      setLoadingActive(true);
+      let podUrl = (await getPodUrlAll(session.info.webId, { fetch: session.fetch }))[0];
```
timbot1789, line 64: Inrupt's `getPodUrlAll` function looks for pods declared on a user's profile in the `pim:storage` field. If there are no pods declared there (the field is not required) this function returns an empty array.

In that case we make an educated guess about where the pod should be.

src/App.jsx
```diff
@@ -94,44 +101,46 @@ const App = () => {
...
-              path="/PASS/forms"
-              element={session.info.isLoggedIn ? <Forms /> : <Navigate to="/PASS/" />}
-            />
-            <Route
-              path="/PASS/inbox"
-              element={session.info.isLoggedIn ? <Inbox /> : <Navigate to="/PASS/" />}
-            />
-            <Route path="*" element={<Navigate to="/PASS/" />} />
-          </Routes>
+          <SignedInPodContext.Provider value={signedInPodMemo}>
```
timbot1789: This doesn't seem to be working correctly, but we should be able to store the pod of the signed in user in a context so other components can access it.

src/components/Users/SolidSignup.jsx
```diff
@@ -0,0 +1,46 @@
...
+          borderRadius: '8px'
+        }}
+      />
+      <Box sx={{ marginRight: '32px' }} />
+      <Button
+        variant="contained"
+        color="primary"
+        size="large"
+        aria-label="Login Button"
+        onClick={() => {}}
```
timbot1789: This button isn't functioning yet. It can be implemented in a separate PR.

The hope here is to navigate to a separate site that hosts pods, sign the user up there, and then navigate back to the PASS. The OIDC provider should send along enough information to populate most if not all of the user entry.

src/components/Users/UserSection.jsx
```diff
@@ -17,21 +17,19 @@ const UserSection = ({ loadingUsers, loadingActive }) => {
...
-          <div style={{ marginBottom: '20px', textAlign: 'center' }}>Loading users list...</div>
-        </section>
-      )}
-    </>
-  );
+  return  <>
+        <ManageUsers />
+
+        {loadingUsers ? (
+          <section className="panel">
```
timbot1789: The logic here was inverted. We should display the loading screen when `loadingUsers` is true

src/utils/user-list-helpers.js
```diff
@@ -0,0 +1,190 @@
+import {
```
timbot1789: These functions were moved out of `session-core` and largely rewritten

src/utils/user-list-helpers.js
```diff
@@ -0,0 +1,190 @@
...
+ */
+
+/**
+ * @typedef {import("../typedefs").userListObject} userListObject
+ */
+
+const createUsersList = async (session, usersListUrl) => {
+  const newUsersList = buildThing(createThing({ name: 'userlist' }))
+    .addStringNoLocale(RDF_PREDICATES.name, 'Users List')
+    .addStringNoLocale(RDF_PREDICATES.description, 'A list of users')
```
timbot1789: There was originally a URL parameter pointing back to itself here. Deleted because it served no function.

leekahung:
The draft is getting pretty large. Would like to know the progress.

src/components/Users/ManageUsers.jsx
```diff
@@ -0,0 +1,165 @@
+// React Imports
+import React, { useContext, useState } from 'react';
```
timbot1789: Not gonna add unit tests for this form, since I expect we'll be changing it shortly to use automated signup.

src/components/Users/ManualWebIdForm.jsx
```diff
@@ -0,0 +1,77 @@
+import React, { useState } from 'react';
```
timbot1789: Not gonna add unit tests for this because we may be deleting it shortly. Keeping for demo purposes for now.

src/components/Users/ManageUsers.jsx
```diff
@@ -0,0 +1,165 @@
...
+
+  return (
+    <FormSection
+      title="Add New User"
+      state={state}
+      statusType="Status"
+      defaultMessage="To be added..."
+    >
+      <form onSubmit={handleAddUser} style={formRowStyle} autoComplete="off">
+        {import.meta.env.VITE_EXPERIMENTAL ? (
```
timbot1789: Following discussions in dev meetings, it seems best to rewrite the user signup form to automatically create a pod in a certain, preconfigured provider. That's what this button's intended purpose is, but it's not functioning yet. Leaving it in the codebase to make demos and discussion easier.

src/components/Users/UsersList.jsx
```diff
@@ -59,45 +64,41 @@ const UsersList = ({ loadingActive }) => {
           <tr>
             <th>First Name</th>
             <th>Last Name</th>
-            <th>Pod URL</th>
+            <th>Web Id</th>
```
timbot1789, line 58: This field was labeled Pod URL, but was actually displaying their Web ID. Updating label to reflect the data being displayed.

src/utils/network/session-core.js
```diff
@@ -356,9 +353,13 @@ export const deleteDocumentContainer = async (session, documentUrl) => {
  * documents being uploaded by authorized users
  */
 
-export const createDocumentContainer = async (session) => {
-  const userContainerUrl = getContainerUrl(session, 'Documents', 'self-fetch');
-  await createContainerAt(userContainerUrl, { fetch: session.fetch });
+export const createDocumentContainer = async (session, podUrl) => {
+  const userContainerUrl = `${podUrl}Documents/`;
+  try {
+    await createContainerAt(userContainerUrl, { fetch: session.fetch });
```
timbot1789: This line can throw an error if the container already exists (although this behavior seems to vary from server to server; node solid server seems to return 200 if the container already exists). So let's handle the error appropriately.

src/utils/user-list-helpers.js
```diff
@@ -0,0 +1,259 @@
...
+  } catch {
+    activityDataset = createSolidDataset();
+    await updateAndSaveActivity();
+    const accessObject = {
+      read: true,
+      append: true,
+      write: true,
+      control: true
+    };
+    const publicAccessObject = {
```
timbot1789: We probably want to keep `activity.ttl` hidden to just people with PASS affiliations, not the general public. However, we don't have a PASS organization set up that we can use for permissions, and setting such a thing up is out of scope of this PR. So giving the public read-only access seems fine for now.

src/utils/network/session-helper.js
```diff
@@ -216,7 +217,8 @@ export const setupAcl = (resourceWithAcl, webId, accessObject) => {
  * @param {URL} documentUrl - Url link to document container
  * @param {string} generateType - A string for "create" or "update"
  * @param {URL} webId - The webId of the user permissions are set to
- * @param {Access} accessObject - Access Object to use when creating the file
+ * @param {Access} accessObject - Access Object to set user permissions
+ * @param {Access} publicAccessObject - Access Object to set public permissions
```
timbot1789: Allow ACL setup to add public access objects as well. It may be worth refactoring ACL management in a later PR so we have better utilities for handling user vs public vs organization permissions.

src/utils/user-list-helpers.js
```diff
@@ -0,0 +1,259 @@
...
+  await setDocAclForUser(session, usersListUrl);
+  return usersListDataset;
+};
+
+const parseUserObjectFromThing = (userThing) => {
+  const person = getStringNoLocale(userThing, RDF_PREDICATES.Person);
+  const givenName = getStringNoLocale(userThing, RDF_PREDICATES.givenName);
+  const familyName = getStringNoLocale(userThing, RDF_PREDICATES.familyName);
+  const webId = getUrl(userThing, RDF_PREDICATES.identifier);
+  const podUrl = getUrl(userThing, RDF_PREDICATES.URL);
```
timbot1789: Add Web ID and Pod Url separately here. Solid allows a user to have a web id at one URL, and a pod at a totally different url.

src/utils/network/session-core.js
```diff
@@ -398,269 +399,6 @@ export const createDocumentContainer = async (session) => {
   }
 };
 
-/*
-  User List Section
```
timbot1789, line 402: Moved user list and user activity helpers into a separate file. Don't want `session-core` to become an unwieldy monolith

src/App.jsx
```diff
@@ -73,15 +79,18 @@ const App = () => {
      * @function fetchData
      */
     async function fetchData() {
-      await generateUsersList(session);
-      await generateActivityTTL(session);
```
leekahung, line 77: This function is needed for generating/updating the active.ttl used for Last Active Date. Should be kept.

src/App.jsx
```diff
@@ -73,15 +79,18 @@ const App = () => {
      * @function fetchData
      */
     async function fetchData() {
-      await generateUsersList(session);
```
leekahung, line 76: This function is needed to create our usersList. Should be kept. Perhaps refactor generateUserList instead with the getPodUrlAll function inside it.

src/App.jsx
```diff
@@ -73,15 +79,18 @@ const App = () => {
...
+      podUrl = podUrl || session.info.webId.split('profile')[0];
+      setSignedInPod(podUrl);
+      await updateUserActivity(session, podUrl);
+      await createDocumentContainer(session, podUrl);
       await createOutbox(session);
       try {
-        let listUsers = await getUsersFromPod(session);
-        setUserList(listUsers);
+        setLoadingUsers(true);
+        setLoadingActive(true);
```
leekahung: These states were already set to true when initialized.

src/components/Users/ManageUsers.jsx
leekahung, line 1: Could consider generating a new index.js in `src/components/Users/` for handling imports/exports in Users.

src/routes/UserSection.jsx
```diff
@@ -8,7 +8,7 @@ import LinearProgress from '@mui/material/LinearProgress';
 import Paper from '@mui/material/Paper';
 import Typography from '@mui/material/Typography';
 // Component Imports
-import { ManageUsers } from '../components/Form';
+import ManageUsers from '../components/Users/ManageUsers';
```
leekahung: Think we can `import { ManageUsers, UsersList } from "../components/Users";` after the index.js is setup.

src/utils/index.js
```diff
@@ -9,21 +9,14 @@
...
-import runNotification from './frontend/notification-helper';
+import { runNotification, clearProcessing } from './frontend/notification-helper';
 import {
   uploadDocument,
   updateDocument,
   getDocuments,
   deleteDocumentContainer,
   deleteDocumentFile,
   setDocAclPermission,
-  generateUsersList,
```
leekahung, line 20: Should keep this for the usersList.

src/utils/index.js
```diff
@@ -9,21 +9,14 @@
...
-  addUserToPod,
-  getUsersFromPod,
-  deleteUserFromPod,
-  generateActivityTTL,
```
leekahung, line 24: Should keep this for generating initial active.ttl file.

src/utils/network/session-core.js
leekahung: I agree we need to refactor this out, but I think we should do this in a separate PR.

src/utils/network/session-core.js
```diff
@@ -809,8 +547,11 @@ export const sendMessageTTL = async (session, messageObject) => {
 
 export const createOutbox = async (session) => {
   const outboxContainerUrl = getContainerUrl(session, 'Outbox', 'self-fetch');
-  await createContainerAt(outboxContainerUrl, { fetch: session.fetch });
+  try {
```
leekahung: The try/catch approach in PR #201 should be enough to fix this function.

src/utils/user-list-helpers.js
leekahung: Think we should limit the scope of this PR to the conversion of users to webids. Can refactor in a separate PR.

leekahung (changes requested):
The refactoring for the conversion from users to webids is pretty significant already, I think we can dedicate a new PR for the refactoring session-core (like moving all the significant refactoring from here and move them to a new branch from Development with a new PR).

As for the usersList, I think we can incorporate getPodUrlAll to that. However would like to know further as to how the new system would work, like how it'll keep track of webIds and the Last Active Date without active.ttl.

src/utils/network/session-core.js
leekahung: Think this is the strategy for the refactoring in PR #201. Could leave the refactoring to there.

src/utils/user-list-helpers.js
leekahung: Yeah, we should make active.ttl private in the future with a PASS specific container with read access for all PASS users. For now, we'll just leave that in public.

src/utils/network/session-helper.js
leekahung: Can use the same line for Line 221 for the updated JSDoc in setupAcl. Could also wrap `publicAccessObject` in square brackets to indicate it's an optional parameter, `[publicAccessObject]`. Same with accessObject (now that I see it).

src/utils/network/session-core.js
timbot1789: Removed this

src/routes/UserSection.jsx
timbot1789: Reversed this change

src/App.jsx
timbot1789: updated

src/utils/user-list-helpers.js
timbot1789: I've reworked this PR so it is only a backend update to users so they work properly with both webIds and Community Solid Server. Since we are still discussing the user registration flow, I can put that (and the UI updates) in a separate PR.

src/App.jsx
timbot1789: This functionality has been moved to `updateUserActivity` in `model-helpers/User.js`

src/App.jsx
timbot1789: This functionality has been moved to `loadUserList` in `model-helpers/UserList.js`

src/components/Users/ManageUsers.jsx
timbot1789: Removed this change from this PR.

src/utils/index.js
timbot1789: This functionality has been moved to `loadUserList` in `model-helpers/UserList.js`

src/utils/index.js
timbot1789: This functionality has been moved to `updateUserActivity` in `model-helpers/User.js

src/components/Form/ManageUsers.jsx
```diff
@@ -1,14 +1,11 @@
...
 // Utility Imports
-import {
-  runNotification,
-  addUserToPod,
-  getUserListActivity,
-  SOLID_IDENTITY_PROVIDER
-} from '../../utils';
+import { runNotification, SOLID_IDENTITY_PROVIDER } from '../../utils';
+import { createUser } from '../../model-helpers/User';
+
```
leekahung, line 8: Extra space in Line 8

src/model-helpers/User.js
```diff
@@ -0,0 +1,173 @@
...
+};
+
+/**
+ * Fetches the last time a user was active in pass
+ *
+ * @memberof User
+ * @function loadUserActivity
+ * @param {URL} podUrl - Url of user's pod to fetch from
+ * @param {Session} session - Solid's Session Object {@link Session}
+ * @returns {string} last time the user was active in PASS
```
leekahung: Mentioned this during our discussion, JSDoc could take in types similar to TypeScript, so for Promises with string we can have `{Promise<string>}` in Line 84

src/model-helpers/User.js
```diff
@@ -0,0 +1,173 @@
...
+};
+
+/**
+ * Converts an inrupt Thing taken from the users list into a user object
+ *
+ * @memberof User
+ * @function parseUserFromThing
+ * @param {import('@inrupt/solid-client').Thing} userThing - the Thing to build the user from
+ * @param {Session} session - Solid's Session Object {@link Session}
+ * @returns {object} user object
```
leekahung, line 138: This could be `{Promise<object>}` or `{Promise<userListObject>}` once we've updated the type definition for userListObject in typedefs.js.

Mostly a note to self: Might need to update the naming for Types too though in a future PRs, like use `UserListObject` instead with PascalCase to make it more TypeScript-like.

src/model-helpers/User.js
```diff
@@ -0,0 +1,173 @@
...
+};
+
+/**
+ * Creates a user object from a provided form submission
+ *
+ * @memberof User
+ * @function createUser
+ * @param {Session} session - Solid's Session Object {@link Session}
+ * @param {object} userSubmission - an object from a form submission containing the user creation data
+ * @returns {Promise} Promise - Updates last active time of user to lastActive.ttl
```
leekahung: We can use `{Promise<object>}` to update this JSDoc.

Think we could also update our typedefs for userListObject in typedefs.js so that it'll include podUrl.

Screen Shot 2023-06-02 at 15 19 40

If we do that, we can use `{Promise<userListObject>}` instead. We should probably update the naming for types too with `UserListObject` instead, but think we can do that in another PR.

leekahung (approved):
Nothing too big with regards to comments/feedback, think it looks good. Should be ready for merging.

PR #178, opened 2023-05-23
Issue 173/delete document styling updates

### LINKED ISSUE #173: Delete Document Styling Updates
Adding MUI styling to the Delete Document. Will implement styling done with Search Document (#131) for consistency.

andycwilliams:
Applying styling to the Delete Document component. Nearly identical to the styling done on #131. 

Note how the document selection remains unstyled, just as with Search Document. As it is an imported function that affects multiple components, it will be changed once those have had their initial styling. This is to limit the possibility of errors and breaking functionality.


src/components/Form/DeleteDocumentForm.jsx
```diff
@@ -50,30 +55,35 @@ const DeleteDocumentForm = () => {
...
-        <div style={formRowStyle}>
-          <label htmlFor="delete-doctype">Select document type to delete: </label>
-          <DocumentSelection htmlId="delete-doctype" />{' '}
-          <button disabled={state.processing} type="submit">
-            Delete Document
-          </button>
-        </div>
-      </form>
+      <Box sx={{ minWidth: 120 }}>
+        <form onSubmit={handleDeleteDocument} autoComplete="off">
```
leekahung: Now that I think about it, I don't think there are text field for this form or for FetchDocumentForm, they're already pre-populated.

Think we can safely remove the autoComplete attributes without affecting anything.

leekahung (changes requested):
Nothing too big, just one more thing. See feedback.

timbot1789 (commented):
Why are the buttons stretched to the max width?

src/components/Form/DeleteDocumentForm.jsx
andycwilliams: Removed.

src/components/Form/FetchDocumentForm.jsx
```diff
@@ -67,11 +67,11 @@ const FetchDocumentForm = () => {
       defaultMessage="To be searched..."
     >
       <Box sx={{ minWidth: 120 }}>
-        <form onSubmit={handleGetDocumentSubmission} autoComplete="off">
+        <form onSubmit={handleGetDocumentSubmission}>
           <InputLabel id="search-doctype">
             <i>Select Document Type</i>
           </InputLabel>
-          <FormControl fullWidth variant="standard">
+          <FormControl  variant="standard">
```
leekahung: Prettier caught the extra space here. do a `npm run prettier:run` to fix it.

src/components/Form/FetchDocumentForm.jsx
leekahung, line 1: Like what we did with the form containers, I think we can center the form elements if we're going to shrink the buttons to an appropriate size.

Screen Shot 2023-05-24 at 0 25 57

We can do the same thing for the "Delete Document". Think it'll turn out nicely.

leekahung (dismissed):
See feedback for request changes.

src/components/Form/FetchDocumentForm.jsx
andycwilliams: When I run that I get this error:
```
[error] No files matching the pattern were found: "'src/**/*.js'".
[error] No files matching the pattern were found: "'src/**/*.jsx'".
[error] No files matching the pattern were found: "'test/**/*.js'".
[error] No files matching the pattern were found: "'test/**/*.jsx'".
```

andycwilliams:
Tweaked the styling so that the form elements are smaller and centered. There is still a large amount of empty space but we're still waiting on guidance from the UX team on how to format the desktop version.

Also, I'm not sure how we will end up implementing the status updates at the bottom of the forms. Will they remain there, appear in a pop-up, something else?

xscottxbrownx:
> Tweaked the styling so that the form elements are smaller and centered. There is still a large amount of empty space but we're still waiting on guidance from the UX team on how to format the desktop version.
> 
> Also, I'm not sure how we will end up implementing the status updates at the bottom of the forms. Will they remain there, appear in a pop-up, something else?
> 
> 

I was thinking about these the other day...

my approach without any input from others would've been to have a spinning loader show while the function was running, once the function is complete it will show the result.  But that (once function is complete it will show result) would depend a little on what function/form it is and how they are displayed on the page.

Something that needs a discussion for sure.

src/components/Form/DeleteDocumentForm.jsx
```diff
@@ -50,30 +55,29 @@ const DeleteDocumentForm = () => {
...
-          </button>
-        </div>
-      </form>
+      <Box display="flex" justifyContent="center" sx={{ minWidth: 120 }}>
+        <form onSubmit={handleDeleteDocument}>
+          <InputLabel id="delete-doctype">
+            <i>Select Document Type to Delete:</i>
+          </InputLabel>
+          <FormControl variant="standard" fullWidth>
+            <DocumentSelection htmlId="search-doctype" />
```
xscottxbrownx: **line 71** - `htmlID` should be "delete-doctype" to match the `id` of the label to link the two together.

xscottxbrownx (changes requested):
just the tiny change for the label and it's matching input, then looks good to me.

src/components/Form/DeleteDocumentForm.jsx
leekahung: Good catch!

andycwilliams:
> I was thinking about these the other day...
> 
> my approach without any input from others would've been to have a spinning loader show while the function was running, once the function is complete it will show the result. But that (once function is complete it will show result) would depend a little on what function/form it is and how they are displayed on the page.
> 
> Something that needs a discussion for sure.

I've tested using a loading animation for the UsersList since it takes a moment longer to load than everything else, and I think it looks good.

We need more feedback before changing it, but a quick pop-up animation or alert makes sense to me. At least for some of these.

src/components/Form/FetchDocumentForm.jsx
leekahung: Not sure what happened, but this is what I get when I ran your latest changes. Seems to catch them just fine.

Screen Shot 2023-05-24 at 23 09 27

leekahung:
> 

This is starting to look more modal-like, which is good if we're going to transition to it. One thing we could do too, now that we're talking about the status notification is that we can center this too so that they're all lined up. The animation/message could replace what we have now for the purple status text eventually.

leekahung (commented):
Minor stylistic comments. See feedback.

src/components/Form/FetchDocumentForm.jsx
leekahung: Still thinking about what might have gone wrong on your end with Prettier, but the space problem is gone, so I'm fine for now.

andycwilliams:
Alright, so I've centered the elements more and refactored a bit of styling from the CSS file into the MUI theme. Plus I did a bit of formatting with the forms, mostly keeping in mind the modal approach. It ought to be easier to go from modal to desktop than vice versa.





Another change is adding loading animations. That wasn't a necessity for this PR, but I think this is a decent start. I added one to the Users List as well. By no means do we have to keep that. I just think it looks nice and wanted to share it.



I also tried to shorten the status messages a bit because they really stretch out the forms otherwise.

In addition, a lot of the smaller changes such as spacing, minimum width, text styling, etc. can be easily tweaked. Primarily I wanted to get more files converted to MUI, so a few more have been changed than only Delete Document.

There's certainly refactoring to do and redundant code to trim, but with this commit I'm looking for feedback on whether this is the direction we want to go.

src/components/Form/FetchDocumentForm.jsx
andycwilliams: Doesn't seem to be a real issue, but I wanted to provide more information on this in case it's useful to anyone.

Running `npm run prettier:run` yields the same error.

However, when I run `npm run prettier:check`, I get this: 

```
Checking formatting...
[error] No files matching the pattern were found: "'src/**/*.js'".
[error] No files matching the pattern were found: "'src/**/*.jsx'".
[error] No files matching the pattern were found: "'test/**/*.js'".
[error] No files matching the pattern were found: "'test/**/*.jsx'".
All matched files use Prettier code style!
```

The only explanation in my mind is that since I have Prettier installed on my editor anyway, the script checks for that and does not execute properly.

src/components/Notification/StatusMessage.jsx
leekahung, line 1: I certainly like the loading animations, we probably would need to tweak the timing for StatusMessage or when we should use the loading animation for StatusMessage. For something like uploading documents or searching documents, this works perfectly. The examples you showed in your last comment are good examples of when to use them as well. Would like to keep them if we can, however, there are a few edge cases.

For that reason, I would like to hold on just this one thing.

For example, for the selection of users, the "select user" interaction with the button should be instantaneous since it's all within PASS via React Context and not between PASS and Solid (which takes a significantly longer amount of time). So a loading animation would be unnecessary. But because we're using StatusMessage to indicate that, the loading animation is unfortunately triggered. Think we could refactor this to use an alternative message scheme (but that's for another PR).

Another case where we might not need the animation is when an operation errors out.

Screen Shot 2023-05-26 at 18 11 48

While it makes sense to disable the button to prevent redundant actions being performed while the function is running, it shouldn't still still be a "loading" screen if we know the operation failed. We can perhaps show an alternative icon/animation.

I certainly want these animations, but I think should do this in another PR where it can be better flushed out.

leekahung (commented):
See feedback comment.

src/components/Notification/StatusMessage.jsx
leekahung: Planning to approve without the loading animation with StatusMessage/StatusNotification at this time.

src/components/Notification/StatusMessage.jsx
```diff
@@ -18,19 +21,28 @@ import React from 'react';
...
+        display: 'flex',
+        flexDirection: 'column',
+        justifyContent: 'center',
+        alignItems: 'center'
+      }}
+    >
       {notification}
-    </dd>
+      <br />
+      <CircularProgress />
```
leekahung: I think we can just comment this out in the component and add a comment on top as a `TODO: need to discuss when to trigger animation`. That means we would need to comment out the import for `<CircularProgress />` as well. At least for the time being.

That way we can keep what we have in this PR and not load the animation until we figure out what to do with them.

src/components/Form/FetchDocumentForm.jsx
leekahung: Also, refrain from closing conversations other people started. In this case, other developers might have the same issue after all. This way, they could check up on the progress.

src/components/Notification/StatusNotification.jsx
```diff
@@ -19,14 +21,22 @@ import StatusMessage from './StatusMessage';
  */
 
 const StatusNotification = ({ notification, statusType, defaultMessage, locationUrl = '' }) => (
-  <dl className="display">
-    <dt>{statusType}:</dt>
+  <Typography
```
leekahung, line 24: Line 24 `<Typogrphy>` component has a `<div>` inside it (Line 37). This will trigger an error in the console log. Switch this to perhaps a `<Box>`.

src/components/Notification/StatusNotification.jsx
```diff
@@ -19,14 +21,22 @@ import StatusMessage from './StatusMessage';
...
+    sx={{
+      marginTop: 3,
+      display: 'flex',
+      flexDirection: 'column',
+      justifyContent: 'center',
+      alignItems: 'center'
+    }}
+  >
+    <>{statusType}:</>
```
leekahung: Considering it was suppose to be a `<dt>` element, you could wrap this with `<Typography>`

src/components/Notification/StatusNotification.jsx
leekahung: Line 24 `<Typogrphy>` component has a `<div>` inside it (Line 37). This will trigger an error in the console log. Switch this to perhaps a `<Box>`.

src/components/Users/UserSection.jsx
```diff
@@ -23,12 +29,26 @@ const UserSection = ({ loadingUsers, loadingActive }) => {
...
+          <Box
+            sx={{
+              marginTop: 3,
+              minWidth: 120,
+              display: 'flex',
+              flexDirection: 'column'
+            }}
+          >
+            <Paper elevation={2} sx={{ display: 'inline-block', mx: '2px', padding: '20px' }}>
+              <Typography sx={{ marginBottom: '20px', textAlign: 'center' }}>
```
leekahung, line 40: The `<Typography>` component in Line 42 will trigger an error since there's a `<div>` inside it at Line 46. Could replace this with a `<Box>`.

src/style.css
```diff
@@ -34,16 +34,12 @@ header {
   color: purple;
 }
 
-.display {
+/* .display {
```
leekahung: I think we can remove these unused stylings if we're replacing them with MUI components used in this PR.

leekahung (changes requested):
See feedback for updates.

src/components/Form/DeleteDocumentForm.jsx
```diff
@@ -50,30 +55,29 @@ const DeleteDocumentForm = () => {
...
-            Delete Document
-          </button>
-        </div>
-      </form>
+      <Box display="flex" justifyContent="center" sx={{ minWidth: 120 }}>
+        <form onSubmit={handleDeleteDocument}>
+          <InputLabel id="delete-doctype">
+            <i>Select Document Type to Delete:</i>
+          </InputLabel>
+          <FormControl variant="standard" fullWidth>
```
leekahung, line 70: This is good. Since we're starting to add `<FormControl>` around DocumentSelection, we should be able to edit DocumentSelection soon with MUI `<Select>`/`<MenuItem>` instead of HTML `<select>`/`<option>`. Should be able to do it once we get the other forms with DocumentSelection moved over to MUI. We'll leave that to a different PR.

src/components/Form/DeleteDocumentForm.jsx
```diff
@@ -50,30 +55,29 @@ const DeleteDocumentForm = () => {
...
-          <DocumentSelection htmlId="delete-doctype" />{' '}
-          <button disabled={state.processing} type="submit">
-            Delete Document
-          </button>
-        </div>
-      </form>
+      <Box display="flex" justifyContent="center" sx={{ minWidth: 120 }}>
+        <form onSubmit={handleDeleteDocument}>
+          <InputLabel id="delete-doctype">
+            <i>Select Document Type to Delete:</i>
```
leekahung: Think we should use `<em>` instead for emphasizing text like how we're using `<strong>` instead of `<b>`.

src/components/Notification/StatusNotification.jsx
```diff
@@ -19,14 +22,33 @@ import StatusMessage from './StatusMessage';
...
-    <dt>{statusType}:</dt>
+  <Box>
+    <Typography
+      sx={{
+        marginTop: 3,
+        display: 'flex',
+        flexDirection: 'column',
+        justifyContent: 'center',
+        alignItems: 'center'
+      }}
```
leekahung: These should be in `<Box>` instead of `<Typography>`.

src/components/Notification/StatusNotification.jsx
```diff
@@ -19,14 +22,33 @@ import StatusMessage from './StatusMessage';
...
       <StatusMessage notification={notification} locationUrl={locationUrl} />
     ) : (
-      <dd>{defaultMessage}</dd>
+      <Typography
+        sx={{
+          display: 'flex',
+          flexDirection: 'column',
+          justifyContent: 'center',
+          alignItems: 'center'
+        }}
```
leekahung: We can remove Lines 41 to 46 since the styling of the `Box` contents would be handled by `<Box>` in Line 25 once you move the sx from Line 27 to 33 into it.

Screen Shot 2023-05-27 at 9 48 11

Screen Shot 2023-05-27 at 9 50 28

src/theme.js
```diff
@@ -2,7 +2,15 @@ import { createTheme } from '@mui/material/styles';
 
 const theme = createTheme({
   typography: {
-    fontFamily: 'Roboto, sans-serif'
+    // button: {
```
leekahung: Can remove these since they're unused

src/components/Notification/StatusMessage.jsx
```diff
@@ -18,19 +24,35 @@ import React from 'react';
...
 
   return (
-    <dd className="labelStatus" role="alert">
+    <Typography
+      sx={{
+        display: 'flex',
+        flexDirection: 'column',
+        justifyContent: 'center',
+        alignItems: 'center'
+      }}
```
leekahung: Lines 45 to 50 can be removed. Styling for centering the text is already handled by StatusNotification via `<Box>`.

src/components/Notification/StatusMessage.jsx
```diff
@@ -18,19 +24,35 @@ import React from 'react';
...
-        {notification}{' '}
-        <a href={locationUrl} target="_blank" rel="noreferrer">
-          {locationUrl}
+      <Typography
+        variant="dd"
+        sx={{
+          display: 'flex',
+          justifyContent: 'center',
+          alignItems: 'center'
+        }}
```
leekahung: Lines 29 to 33 can be removed. Styling for centering the text is already handled by StatusNotification via `<Box>`.

leekahung (changes requested):
See requested feedback of the latest changes.

Additionally run a git pull for this update before doing your changes again. It looks like there's a rogue package-lock.json file in there, which shouldn't happen with this PR unless package-lock.json was delete and regenerated with npm install. So I've copied a good package-lock.json file from your branch a few commits back and pushed it back into the branch.

src/components/Users/UserSection.jsx
```diff
@@ -21,14 +27,28 @@ const UserSection = ({ loadingUsers, loadingActive }) => {
...
       <ManageUsers />
       {loadingUsers ? (
-        <section className="panel">
-          <strong>Users List</strong>
-          <br />
-          <br />
-          <div style={{ marginBottom: '20px', textAlign: 'center' }}>Loading users list...</div>
-        </section>
-      ) : (
         <UsersList loadingActive={loadingActive} />
```
leekahung: Minor thing, I think I can update this since it's close to being finished. But the ordering of this ternary should be reversed.

If loadingUsers is true, we should have the component with the loading animation and not UserList.

leekahung: Fixed the ordering, looks good now.

leekahung (approved):
Alright, looks good. I believe all the major points from everyone who reviewed it has been touched. Think this is ready for merging.

PR #181, opened 2023-05-24
Clean up warning messages in Github Actions

timbot1789:
Github was generating some warnings when running, one about outdated actions, the other about the wrong version of node being used in the unit tests. This fixes them.

Also, lint and test no longer depend on build succeeding. The change should let those tests run faster.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,12 +55,16 @@ jobs:
   lint:
     name: Lint
     runs-on: ubuntu-latest
-    needs: build
     permissions:
       contents: read
     steps: 
       - name: Code Checkout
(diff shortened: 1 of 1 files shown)


xscottxbrownx:
Makes sense to me!

leekahung:
I think we can keep Linting and Formatting as part of the CI process.

timbot1789:
> 
I'm not sure what you mean. This PR doesn't remove Linting, just cleans up the action code for it a bit.

leekahung:
>  I'm not sure what you mean. This PR doesn't remove Linting, just cleans up the action code for it a bit.

Okay,  that should be fine then. Thought having these detached will separate the checks from the CI process.

PR #183, opened 2023-05-25
Optimizing functions in session-core and updating JSDoc comments

leekahung:
This PR provides minor optimizations to code existing in session-core and updates JSDoc comments related to functions in session-core with no breaking changes. Helper functions from session-helper has been adjusted to allow for these optimizations.

## New unit tests

Considering this PR affects `setDocAclForUser` from session-helper the most, a unit test for the function has been built for it.

This unit tests checks for whether certain variables are generated correctly in preparation for running the helper function setupAcl, which takes in resourceAcl, webId, and accessObject.

The variable resourceAcl is dependent on a new variable called `generateType`, which either creates a new ACL file via `getSolidDataset` and `createAcl` or update an existing ACL file via `getSolidDatasetWithAcl` and `getResourceAcl`.

### DIFF
--- a/src/utils/network/session-core.js
+++ b/src/utils/network/session-core.js
@@ -8,9 +8,6 @@ import {
   saveSolidDatasetInContainer,
   deleteContainer,
   deleteFile,
-  saveAclFor,
-  getSolidDatasetWithAcl,
-  getResourceAcl,
   overwriteFile,
   getThingAll,
   saveSolidDatasetAt,
@@ -22,11 +19,10 @@ import {
 import { RDF_PREDICATES, UPLOAD_TYPES } from '../../constants';
 import {
   getContainerUrl,
-  setupAcl,
   placeFileInContainer,
   getContainerUrlAndFiles,
   hasTTLFiles,
-  createDocAclForUser,
+  setDocAclForUser,
   createResourceTtlFile,
   updateTTLFile,
   SOLID_IDENTITY_PROVIDER,
@@ -37,6 +33,10 @@ import { getUserSigningKey, signDocumentTtlFile } from '../cryptography/credenti
 
 /**
  * @typedef {import('@inrupt/solid-ui-react').SessionContext} Session
+ */
+
+/**
+ * @typedef {import('@inrupt/solid-client').Access} Access
  */
 
 /**
@@ -60,25 +60,19 @@ import { getUserSigningKey, signDocumentTtlFile } from '../cryptography/credenti
  * @function setDocAclPermission
  * @param {Session} session - Solid's Session Object {@link Session}
  * @param {string} fileType - Type of document
- * @param {string} fetchType - Type of fetch (to own Pod, or "self-fetch" or to
- * other Pods, or "cross-fetch")
+ * @param {Access} permissions - The Access object for setting ACL in Solid
  * @param {string} otherPodUsername - Username to other user's Pod or empty string
  * @returns {Promise} Promise - Sets permission for otherPodUsername for given
  * document type, if exists, or null
  */
 
-export const setDocAclPermission = async (session, fileType, accessType, otherPodUsername) => {
+export const setDocAclPermission = async (session, fileType, permissions, otherPodUsername) => {
   const documentUrl = getContainerUrl(session, fileType, 'self-fetch');
-  const podResouceWithAcl = await getSolidDatasetWithAcl(documentUrl, { fetch: session.fetch });
-
-  const resourceAcl = getResourceAcl(podResouceWithAcl);
   const webId = `https://${otherPodUsername}.${
     SOLID_IDENTITY_PROVIDER.split('/')[2]
   }/profile/card#me`;
-  const accessObject = accessType === 'Give' ? { read: true } : { read: false };
 
-  const updatedAcl = setupAcl(resourceAcl, webId, accessObject);
-  await saveAclFor(podResouceWithAcl, updatedAcl, { fetch: session.fetch });
+  await setDocAclForUser(session, documentUrl, 'update', webId, permissions);
 };
 
 /**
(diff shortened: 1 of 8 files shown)


src/utils/network/session-core.js
```diff
@@ -69,22 +69,13 @@ import { getUserSigningKey, signDocumentTtlFile } from '../cryptography/credenti
...
-  let accessObject;
-  switch (accessType) {
-    case 'Give':
-      accessObject = { read: true };
-      break;
-    default:
-      accessObject = { read: false };
-      break;
-  }
+  const accessObject = accessType === 'Give' ? { read: true } : { read: false };
```
leekahung: Since there's only two cases (Give or Revoke), there's no need to use a switch statement

src/utils/network/session-core.js
```diff
@@ -117,15 +108,10 @@ export const setDocContainerAclPermission = async (session, accessType, otherPod
...
-  let accessObject;
-  switch (accessType) {
-    case 'Give':
-      accessObject = { read: true, write: true, append: true };
-      break;
-    default:
-      accessObject = { read: false, write: false, append: false };
-      break;
-  }
+  const accessObject =
```
leekahung: Similar to before, since there's only two cases (Give or Revoke), there's no need to use a switch statement

src/utils/network/session-core.js
```diff
@@ -182,8 +170,9 @@ export const uploadDocument = async (
   const ttlFileExists = hasTTLFiles(datasetFromUrl);
 
   // Guard clause will throw function if container already exist with ttl file
+  // ttl file indicates update instead of upload document
   if (ttlFileExists) {
-    throw new Error('Container already exists');
+    throw new Error('TTL file already exists');
```
leekahung, line 159: It's fine if a container exist. However, a TTL file is the main indicator of whether or not a file has previously been uploaded.

leekahung (commented):
See comments.

src/utils/network/session-core.js
timbot1789: We can make this even simpler:
```
const accessObject = {
  read: accessType === 'Give'
}
```

However, it looks like this function's core responsibility is setting permissions to a file, not building the permissions object itself. Why not pass in the `accessObject` from the caller? We can change the function signature to
`setDocAclPermissions = async (session, fileType, accessObject, otherPodUsername)`

Then update line 40 of `SetAclPermissionForm.jsx` to something like
```
const permissions = {
  read: (event.target.setAclPerms.value === 'Give')
}
```

And pass it into this function at line 80 of `SetAclPermissionForm.jsx`
`await setDocAclPermission(session, docType, permissions, podUsername)`

This way, `setDocAclPermission` can set more than just read permissions if we need it to, and it doesn't need to worry about interpreting the strings `Give` or `Revoke`, which are essentially display text.

timbot1789: Either way, the JSDoc string for this function looks like it needs updating.  It's missing the access parameter

src/utils/network/session-core.js
timbot1789: These two functions, `setDocContainerAclPermission` and `setDocAclPermission` seem to do very similar things. Any reason they need to be separate? It seems to me the only difference is that `setDocContainerAclPermission` sets multiple Acl documents in its `forEach` loop. Could we call `setDocAclPermission` in that `forEach` loop?

timbot1789 (commented):
Since you're touching these functions, can we add a few unit tests for the AclPermissions functions?

src/utils/network/session-core.js
leekahung: That's a really nice optimization! Will update to this.

leekahung:
> Since you're touching these functions, can we add a few unit tests for the AclPermissions functions?

Yeah, that'll be great. Think you could build them in this branch so that the tests are closer to the updated form of these functions.

timbot1789:
> > Since you're touching these functions, can we add a few unit tests for the AclPermissions functions?
> 
> Yeah, that'll be great. Think you could build them in this branch so that the tests are closer to the updated form of these functions.

I meant could you add them to the PR? Unit tests are most effective if the creator of PRs adds them to any code they write as part of the PR process.

src/utils/network/session-core.js
leekahung: Function and functionality for setDocAclPermissions and SetAclPermissionForm.jsx has been updated with the Access object called permissions.

src/utils/network/session-core.js
leekahung: That is a sensible approach. Think I'll use setDocAclPermission with setDocContainerAclPermission.

leekahung: In order to make this more streamlined, I've updated all functions related to `createDocAclForUser` in session-helper, which is now `setDocAclForUsers`. This includes the `setDocAclPermission`, `setDocContainerAclPermission`, `uploadDocument`, `createDocumentContainer`, `generateUsersList`, and `generateActivityTTL`  in session-core and `savePassUserCredentials` in credentials-helper

The new `setDocAclForUsers` accepts two additional parameters called `generateType` and `webId`, the first being a string ("create" or "update") and the latter being just the webId the ACL is being set to. (See JSDoc below)

```
/**
 * Function that generates ACL file for container containing a specific document
 * type and turtle file and give the user access and control to the Solid
 * container
 *
 * @memberof utils
 * @function setDocAclForUser
 * @param {Session} session - Solid's Session Object (see {@link Session})
 * @param {URL} documentUrl - Url link to document container
 * @param {string} generateType - A string for "create" or "update"
 * @param {URL} webId - The webId of the user permissions are set to
 * @param {Access} accessObject - Access Object to use when creating the file
 * @returns {Promise} Promise - Generates ACL file for container and give user
 * access and control to it and its contents
 */
```

By default, it'll uses `accessObject = { read: true, append: true, write: true, control: true }` which is defined in `setDocAclForUser`. This can be changed if you pass in your own accessObject.

leekahung:
> I meant could you add them to the PR? Unit tests are most effective if the creator of PRs adds them to any code they write as part of the PR process.

Yeah, I could try adding them. I'll take a crack at it.

leekahung:
Not really sure how can we mock the authenticated sessions for fetch the .acl files. I guess we can check if Access objects are passed into functions correctly?

leekahung:
Created two new unit tests in session-helper.test.js to check if resourceAcl is generated correctly for setupAcl. Depending on `generateType`, resourceAcl could either be a new ACL file or an updated ACL file

timbot1789 (approved):

PR #184, opened 2023-05-25
Optimizing functions in session-helper and updating JSDoc comments

leekahung:
Similar to session-core, minor optimizations to session-helper and updated JSDoc comments.

### DIFF
--- a/src/utils/network/session-helper.js
+++ b/src/utils/network/session-helper.js
@@ -49,6 +49,10 @@ import { RDF_PREDICATES } from '../../constants';
  */
 
 /**
+ * @typedef {import('crypto-js').CryptoJS.lib.WordArray} WordArray
+ */
+
+/**
  * The URL to a specific Solid Provider
  *
  * @name SOLID_IDENTITY_PROVIDER
@@ -91,8 +95,7 @@ export const placeFileInContainer = async (session, fileObject, containerUrl) =>
  * @memberof utils
  * @function hasTTLFiles
  * @param {SolidDataset} solidDataset - Solid's dataset object on Pod
- * @returns {boolean} Boolean - A boolean on whether a ttl file exist from
- * dataset
+ * @returns {boolean} Boolean - A boolean on whether a ttl file exist from dataset
  */
 
 export const hasTTLFiles = (solidDataset) => {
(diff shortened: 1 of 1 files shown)


src/utils/network/session-helper.js
```diff
@@ -101,12 +104,7 @@ export const hasTTLFiles = (solidDataset) => {
     return false;
   }
 
-  const ttlFiles = items.find((item) => item.url.slice(-3) === 'ttl');
-  if (ttlFiles) {
-    return true;
-  }
-
-  return false;
+  return items.some((item) => item.url.endsWith('ttl'));
```
leekahung, line 109: Just learned about endsWith. Think it'll be a good replacement for .slice(-3) === 'ttl'

Also, just need to check if there's any ttl file, so .some() should suffice.

src/utils/network/session-helper.js
```diff
@@ -129,11 +127,11 @@ export const getContainerUrlAndFiles = (solidDataset) => {
   let directory = '';
   const files = [];
 
-  items.forEach((file) => {
-    if (!file.url.slice(-3).includes('/')) {
-      files.push(file);
+  items.forEach((item) => {
```
leekahung, line 132: Replace file with item, since item could be a fileUrl or a directoryUrl

src/utils/network/session-helper.js
```diff
@@ -156,12 +154,10 @@ export const getContainerUrlAndFiles = (solidDataset) => {
  */
 
 export const getContainerUrl = (session, fileType, fetchType, otherPodUsername) => {
-  let POD_URL;
-  if (fetchType === 'self-fetch') {
-    POD_URL = String(session.info.webId.split('profile')[0]);
-  } else {
-    POD_URL = `https://${otherPodUsername}.${SOLID_IDENTITY_PROVIDER.split('/')[2]}/`;
-  }
+  const POD_URL =
```
leekahung, line 159: Another binary case, this time with fetchType. Replacing if/else with ternary

leekahung (commented):
See comments.

src/utils/network/session-helper.js
timbot1789: If `fetchType` is binary, why not pass it in as a boolean? Then `getContainerUrl` doesn't need to worry about the string parsing in addition to everything else it does.

leekahung: Think we can leave this as is for now.

PR #186, opened 2023-05-26
Created function to build outbox when logging in for the first time

### LINKED ISSUE #182: Create an outbox for messages
While we have a dedicated inbox for messages, we're currently saving a copy of the message sent inside the sender's inbox. So both the sender and the recipient receive the message in their inboxes.

This is because we have no dedicated outbox to store sent messages. If a user were to send themselves a message, for example, you'll get duplicate messages sent your inbox since you're both sender and recipient. This will likely lead to confusion if we were to render messages from the inbox.

Ideally, we'll create a new outbox container on the Solid Pod to store messages being sent instead of storing these messages inside the inbox. The ACL settings for outbox should be limited to just the message sender and no one else.

leekahung:
A small PR. This PR sets up the initial outbox to be used for storing sent messages. Container creation is triggered only when logging in for the first time or if outbox has been deleted.

### DIFF
--- a/src/utils/network/session-core.js
+++ b/src/utils/network/session-core.js
@@ -696,7 +696,7 @@ export const updateUserActivity = async (session) => {
 /*
   User Inbox Section
 
-  Functions here deal primarily with user inbox on PASS
+  Functions here deal primarily with user inbox/outbox on PASS
 */
 
 /**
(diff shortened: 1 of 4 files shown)


src/utils/network/session-core.js
```diff
@@ -764,3 +764,22 @@ export const sendMessageTTL = async (session, messageObject) => {
...
+ * @returns {Promise} Promise - Generates an outbox for Pod upon log in if
+ * user's Pod does not have the an outbox to begin with
+ */
+
+export const createOutbox = async (session) => {
+  const outboxContainerUrl = getContainerUrl(session, 'Outbox', 'self-fetch');
+  await createContainerAt(outboxContainerUrl, { fetch: session.fetch });
+
+  // Generate ACL file for container
+  await createDocAclForUser(session, outboxContainerUrl);
```
leekahung: Will be replaced with the new `setDocAclForUser` function in PR #183. Will merge this as is if merged before #183 and after review.

PR #187, opened 2023-05-28
Issue 174/cross pod search styling updates

### LINKED ISSUE #174: Cross Pod Search Styling Updates
Adding MUI styling to the Cross Pod Search component, continuing from https://github.com/codeforpdx/PASS/pull/131. Will begin once https://github.com/codeforpdx/PASS/pull/173 is complete.

andycwilliams:
Applying styling to the Cross Pod Search component based on #173.

Note how the document selection remains unstyled, just as with previous merges. As it is an imported function that affects multiple components, it will be updated once those have had their initial styling. This is to limit the chance of errors and breaking functionality. Styling will ultimately look similar to the enter username function.


src/components/Form/CrossPodQueryForm.jsx
```diff
@@ -76,42 +82,42 @@ const CrossPodQueryForm = () => {
...
+              placeholder={selectedUser}
+              label="Enter username"
+              InputProps={{
+                type: 'search'
+              }}
+              required
+            />
+          </FormControl>
+          <InputLabel id="cross-search-doctype">
+            <em>Select Document Type to Search: </em>
```
leekahung: Just got reminded of it when looking at the site again. Consider the section is pretty clearly defined about what it does (and since we're considering shortening phrases like "Document located at" to "Document found"), I think we can shorten this to just "Select Document Type".

The same could be said for Delete Document.
Screen Shot 2023-05-28 at 1 04 24

src/components/Form/CrossPodQueryForm.jsx
```diff
@@ -76,42 +82,42 @@ const CrossPodQueryForm = () => {
...
+          <FormControl variant="standard" fullWidth>
+            <TextField
+              id="cross-search-doc"
+              name="crossPodQuery"
+              {...username}
+              placeholder={selectedUser}
+              label="Enter username"
+              InputProps={{
+                type: 'search'
+              }}
```
leekahung: This looks like something that would be useful during validation, but I don't think we need this just yet. 

Not until we figure out what to do with our username/webId system at the very least. Can be removed for now.

leekahung (changes requested):
Nothing too big, see feedback.

src/components/Form/DocumentSelection.jsx
```diff
@@ -1,6 +1,6 @@
 // React Imports
 import React from 'react';
-// Utiltiy Imports
+// Utility Imports
```
leekahung, line 3: Good catch!

src/components/Form/DeleteDocumentForm.jsx
```diff
@@ -65,7 +65,7 @@ const DeleteDocumentForm = () => {
       <Box display="flex" justifyContent="center" sx={{ minWidth: 120 }}>
         <form onSubmit={handleDeleteDocument}>
           <InputLabel id="delete-doctype">
-            <em>Select Document Type to Delete:</em>
+            <em>Select Document Type:</em>
```
leekahung: I think we can simplify these further with the colons in "Select Document Type:"

Since they'll just going to be labels for the input fields in MUI, think we can have "Select Document Type" as the final form

src/components/Form/CrossPodQueryForm.jsx
```diff
@@ -76,42 +82,39 @@ const CrossPodQueryForm = () => {
...
-          <label htmlFor="cross-search-doctype">Select document type to search: </label>
-          <DocumentSelection htmlId="cross-search-doctype" />{' '}
-          <button disabled={state.processing} type="submit">
-            Search Pod
-          </button>
-        </div>
-      </form>
+      <Box display="flex" justifyContent="center" sx={{ minWidth: 120 }}>
+        <form onSubmit={handleCrossPodQuery} autoComplete="off">
+          <FormControl variant="standard" fullWidth>
```
leekahung: Looking at the styling of the fields, I think we can just use the default variant for all of the `<FormControl>` components

leekahung (commented):
See feedback

leekahung (approved):
Pretty straightforward changes to styling. I'm good with this.

PR #191, opened 2023-05-29
Issue 190/render message outbox

### LINKED ISSUE #182: Create an outbox for messages

### LINKED ISSUE #190: Render message outbox
Similar to Inbox, we should render what's sent over the the outbox. The only difference is that there'll be no `<NewMessage>` component.

Create a basic outbox component similar to inbox.

leekahung:
# PR Details

This PR is part 3 of the new message system for PASS. Like before, messages being sent to other users are still being sent to their inbox. However, we've repurposed functions to store outbound message for the sender within a dedicated outbox instead of the sender's inbox similar to emails.

Screen Shot 2023-05-31 at 16 51 41 (2)

A new pagination feature has been included to display 5 messages per "page" and a new "outbox" route has been included to accommodate for the new outbox.

## Future considerations

- Refactoring Inbox and Outbox Contexts
- Potential update to routing for /PASS/inbox and /PASS/outbox with /PASS/messages/inbox and /PASS/messages/outbox or just to /PASS/messages with inbox and outbox components that could toggled within messages
- An archive system for messages sent to the outbox
- Improvements to Message system UI


package.json
```diff
@@ -39,6 +39,7 @@
     "crypto-js": "^4.1.1",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "react-paginate": "^8.2.0",
```
timbot1789, line 42: Can you talk a bit about why you chose `react-paginate` for pagination? Pagination's a really useful tool, and there are a number of options out there. I'm curious what made this one stand out.

timbot1789 (commented):
Would love to see some unit tests and snapshots for this

src/utils/network/session-core.js
```diff
@@ -383,35 +383,37 @@ export const createDocumentContainer = async (session, podUrl) => {
...
- * @returns {Promise<inboxListObject[]>} inboxList - An array of inbox messages
+ * @param {string} boxType - The message box being called "Inbox" or "Outbox"
+ * @param {messageListObject[]} listMessages - List of messages
+ * @returns {Promise<messageListObject[]>} inboxList - An array of inbox messages
  * from the user's inbox on Solid in JSON format
  */
-export const getInboxMessageTTL = async (session, inboxList) => {
-  const inboxContainerUrl = getContainerUrl(session, 'Inbox', 'self-fetch');
+
+export const getMessageTTL = async (session, boxType, listMessages) => {
```
timbot1789: There's a lot going on in this function. Is it possible to break up into smaller pieces? We could split the message fetching and message parsing into 2 functions at least. There are probably a few other simplifications that can be made as well.

src/utils/network/session-core.js
```diff
@@ -383,35 +383,37 @@ export const createDocumentContainer = async (session, podUrl) => {
...
-
-  Functions here deal primarily with user inbox/outbox on PASS
+  User Message Section
+  
+  Functions here deal primarily with user messages on PASS
 */
 
 /**
  * Function that gets list of inbox TTL file messages and returns the messages as
  * JSON
```
timbot1789: Nitpick, but this function returns an object, not JSON. JSON is a string.

src/utils/network/session-core.js
```diff
@@ -507,8 +509,8 @@ export const sendMessageTTL = async (session, messageObject) => {
 
   try {
     await Promise.all([
-      await saveMessageTTLInInbox(session, containerUrl, newSolidDataset, messageSlug),
-      await saveMessageTTLInInbox(session, inboxUrl, newSolidDataset, messageSlug)
+      await saveMessageTTL(session, containerUrl, newSolidDataset, messageSlug),
```
timbot1789: Are the `awaits` here necessary?

src/App.jsx
```diff
@@ -85,11 +100,14 @@ const App = () => {
       <SelectUserContext.Provider value={selectedUserObject}>
         <UserListContextProvider session={session}>
           <InboxMessageContext.Provider value={inboxMessageObject}>
-            <AppRoutes
-              isLoggedIn={session.info.isLoggedIn}
-              loadingActive={false}
-              loadMessages={loadMessages}
-            />
+            <OutboxMessageContext.Provider value={outboxMessageObject}>
```
timbot1789, line 111: Is this inbox data universal to the app? It seems to me only components under inbox and outbox would be interested in it. If that's the case, this context would probably be better in that tab.

timbot1789, line 111: Is this inbox data universal to the app? It seems to me only components under inbox and outbox would be interested in it. If that's the case, this context would probably be better in that tab.

timbot1789, line 111: Is this inbox data universal to the app? It seems to me only components under inbox and outbox would be interested in it. If that's the case, this context would probably be better in that tab.

src/utils/network/session-core.js
leekahung: Have moved the parsing of the message TTL file into session-helper as the new parseMessageTTL function.

src/utils/network/session-core.js
leekahung: Gotcha. Switching to just object.

src/utils/network/session-core.js
leekahung: Yeah, it's needed for the async function that saveMessageTTL uses underneath.

src/App.jsx
leekahung: Yeah, it's mainly used for the Inbox, but considering the time it takes to load the existing message, thought it'll be better to start the process in App.jsx.

We'll probably have to think about how to make this more performant (perhaps play around with web workers for the async calls?), but I'm not sure at this moment.

leekahung: Duplicate comment.

leekahung: Two duplicate comments.

package.json
leekahung: @benzelcurry

src/App.jsx
timbot1789: Yeah I'm noticing the performance of these test servers is pretty poor. Not sure how much web workers will really help with that.

If we want to load the data up front, I'd at least move the fetch calls into the OutboxContextProvider like I did with the userList. That way the network calls can all be sent out at once, without blocking each other. Or convert the `await` statements into `Promise().resolve().reject`. That way the load call can just fire them all off and be done. Doesn't have to wait around.

As is though, that `fetchData` call is a pretty large performance bottleneck.

leekahung: Yeah, that's probably a good idea. Think we should do just that in a follow up and wrap both Inbox and Outbox contexts.

Hoping it'll take the load off fetchData.

src/utils/network/session-core.js
```diff
@@ -460,14 +450,14 @@ export const getInboxMessageTTL = async (session, inboxList) => {
...
     session,
     'Inbox',
     INTERACTION_TYPES.CROSS,
     recipientUsername
   );
-  const inboxUrl = getContainerUrl(session, 'Inbox', INTERACTION_TYPES.SELF);
+  const outboxUrl = getContainerUrl(session, 'Outbox', INTERACTION_TYPES.SELF);
 
   const senderUsername = session.info.webId.split('profile')[0].split('/')[2].split('.')[0];
   const recipientWebId = `https://${recipientUsername}.${
```
timbot1789, line 491: The send function breaks on Community Solid Server because of this webId generation. Just pass in the Pod URL.

timbot1789 (changes requested):
Refresh button doesn't seem to be doing anything.

I'd also expect that when I send a message, the message immediately appears in my outbox without me needing to refresh. We should be able to display this even just on the frontend if we put a "pending" status on the message.

And loading of the messages is really slow. There seems to be unnecessary back and forth in the network requests. We should be able to optimize that and make it load a fair deal faster.

[Screencast from 06-04-2023 03:30:10 PM.webm](https://github.com/codeforpdx/PASS/assets/37914436/6faabfae-d1f5-4a2a-8282-656de789dce8)

src/utils/network/session-core.js
```diff
@@ -460,14 +450,14 @@ export const getInboxMessageTTL = async (session, inboxList) => {
...
   const containerUrl = getContainerUrl(
     session,
     'Inbox',
     INTERACTION_TYPES.CROSS,
     recipientUsername
   );
-  const inboxUrl = getContainerUrl(session, 'Inbox', INTERACTION_TYPES.SELF);
+  const outboxUrl = getContainerUrl(session, 'Outbox', INTERACTION_TYPES.SELF);
 
   const senderUsername = session.info.webId.split('profile')[0].split('/')[2].split('.')[0];
```
timbot1789: This breaks on community solid server.

src/utils/network/session-core.js
timbot1789: No, I just tested this, and the `awaits` on lines 493 and 494 are unnecessary, and seem to slow the code down. These lines should be:
```
await Promise.all([
  saveMessageTTL(session, containerUrl, newSolidDataset, messageSlug),
  saveMessageTTL(session, outboxUrl, newSolidDataset, messageSlug)
])

```
You already handle the situation correctly on line 492. "Send out this list of promises, and wait until all of them return." These extra awaits render the `Promise.all` irrelevant. It forces the code to wait for the first `saveMessageTTL` to resolve before it sends the second request, which isn't the behavior we want.

src/utils/network/session-helper.js
```diff
@@ -389,21 +391,99 @@ export const getUserProfileName = async (session, webId) => {
...
+ * @param {Date} date - JavaScript Date object
+ * @param {string} senderName - Name of sender
+ * @param {string} recipientName - Name of recipient
+ * @param {URL} recipientWebId - webId of recipient
+ * @param {Thing[]} messageTTLThing - List of message Things from message boxes
+ * @returns {object} messageObject - An object containinng the message content,
+ * title, uploadDate, sender, and recipient
+ */
+
+export const buildMessageTTL = (
```
timbot1789, line 481: This can be in a follow up PR, but these could be some other useful fields:

- `delivery-status`: a string representing if the message has been delivered. e.g. `pending`, `delivered`, `failed to send`, `refused`, etc.
- `message-id`: a string containing a universally unique identifier. Will allow us to tell if a message in one person's inbox is the same as a message in another person's outbox. Can also be used to check if a message was successfully delivered. If a message with that `message-id` isn't in the inbox, it wasn't delivered
- `in-reply-to`: a `message-id` of a different message. Allows us to implement message threading and conversations

leekahung:
>  Screencast.from.06-04-2023.03.30.10.PM.webm

That's strange. Was suppose to load in outbox right after sending. I'll see if I can fix the bug.

leekahung:
Found the bug I think. getMessageTTL has been updated to use podUrl in App.jsx, but the ones in outbox and inbox haven't updated yet and still use just 3 parameters, gonna fix it.

leekahung:
Got the fix. But yes, the initial loading of data is pretty bad. Would need to think about how to optimize that in the future.

https://github.com/codeforpdx/PASS/assets/14917816/43cea648-68f6-46fa-ad59-81daf38dd1d6

src/utils/network/session-core.js
leekahung: Got it. Gonna pass in PodUrl directly into the sendMessageTTL function.

src/utils/network/session-core.js
leekahung: Got this switched to using podUrl

src/utils/network/session-helper.js
leekahung: Will make note of it

package.json
benzelcurry: > Can you talk a bit about why you chose `react-paginate` for pagination? Pagination's a really useful tool, and there are a number of options out there. I'm curious what made this one stand out.

From the research I did before going with `react-paginate`, it appeared to be the most widely used pagination library for React applications. Additionally, it's relatively easy to use and customize right out of the box, so I figured it would be a good option at least for the time being.

That being said, I'm not opposed to switching to something else if you have any suggestions. I saw that MUI has some pagination stuff, which could potentially be worth looking into. The other option that I contemplated was just hardcoding the pagination, although I figured that would be opening up a can of worms for any other developers that might join the group in the future who end up working on the messaging components.

timbot1789: No worries. It all looks good to me!

src/utils/network/session-core.js
leekahung: Got it, I see what you mean. The await have already been handled. With the await outside the Promise.all

Fixing this.

leekahung:
@timbot1789 @xscottxbrownx Let me know if there's still any other issues with regards to this PR. Think it's close to being wrapped up?

xscottxbrownx:
> @timbot1789 @xscottxbrownx Let me know if there's still any other issues with regards to this PR. Think it's close to being wrapped up?

Oh, I wasn't sure you were done as there were still things not marked as RESOLVED.  I can look at this first thing in morning again.

---

off the top of my head though for a future PR/Issue... I say we make the `Navbar` link just `Messages`, and then in that page have icons/toggle/small sidebar whatever of `Inbox` and `Sent` that switches the views.

src/App.jsx
```diff
@@ -49,16 +54,21 @@ const App = () => {
...
   const initialInboxList = [];
   const [inboxList, setInboxList] = useState(initialInboxList);
   const inboxMessageObject = useMemo(() => ({ inboxList, setInboxList }), [inboxList]);
 
+  /** @type {messageListObject[]} */
+  const initialOutboxList = [];
+  const [outboxList, setOutboxList] = useState(initialOutboxList);
+  const outboxMessageObject = useMemo(() => ({ outboxList, setOutboxList }), [outboxList]);
+
   useEffect(() => {
```
xscottxbrownx, line 72: This useEffect from line 72 - 104...

how does the dependency array only have this run on the first time the user logs in? Would this not run every time they login or logout?

src/routes/Outbox.jsx
```diff
@@ -0,0 +1,78 @@
...
+      id="outbox"
+      className="panel"
+      style={{ display: 'flex', flexDirection: 'column', gap: '20px' }}
+    >
+      <div style={{ display: 'flex', gap: '10px' }}>
+        <StyledButton onClick={handleOutboxRefresh} type="button">
+          Refresh
+        </StyledButton>
+      </div>
+      {loadMessages ? <div>Loading Messages...</div> : <PaginatedMessages messages={outboxList} />}
```
xscottxbrownx, line 62: This doesn't seem to occur on refresh, but feel it should for UX sake. Same in `Inbox`.

src/AppRoutes.jsx
```diff
@@ -20,7 +21,8 @@ const AppRoutes = ({ isLoggedIn, loadingActive, loadMessages }) => {
       <Route element={<ProtectedRoute isLoggedIn={isLoggedIn} />}>
         <Route path="/PASS/home" element={<UserSection loadingActive={loadingActive} />} />
         <Route path="/PASS/forms" element={<Forms />} />
-        <Route path="/PASS/inbox" element={<Inbox loadMessages={loadMessages} />} />
+        <Route path="/PASS/inbox" element={<Inbox loadMessages={loadInboxMessages} />} />
```
xscottxbrownx: Would prefer just a `Messages` route and user can get to Inbox and Outbox in there, but that can be future PR.

src/components/Messages/Pagination.jsx
```diff
@@ -0,0 +1,83 @@
...
+        />
+      </PaginationContainer>
+    </>
+  );
+};
+
+// Helper component for displaying paginated results
+const Messages = ({ currentMessages }) =>
+  currentMessages &&
+  currentMessages.map((message) => <MessagePreview key={uuidv4()} message={message} />);
```
xscottxbrownx, line 54: It wasn't changed in this PR, but not tested thoroughly in a previous PR...

The `MessagePreview` has issues:


https://github.com/codeforpdx/PASS/assets/71395931/e54159a2-452f-4862-9c4f-4a61f4906c43

src/components/NavBar/NavbarLinks.jsx
```diff
@@ -24,7 +24,8 @@ const NavbarLinks = () => {
   const routesArray = [
     { label: 'Home', path: '/PASS/home' },
     { label: 'Forms', path: '/PASS/forms' },
-    { label: 'Inbox', path: '/PASS/inbox' }
+    { label: 'Inbox', path: '/PASS/inbox' },
```
xscottxbrownx, line 27: Same comment as `AppRoutes`, would prefer just `Messages`

src/routes/Inbox.jsx
```diff
@@ -62,19 +63,11 @@ const Inbox = ({ loadMessages }) => {
...
-      {loadMessages ? (
-        <div>Loading Messages...</div>
-      ) : (
-        <div style={{ display: 'flex', flexDirection: 'column', gap: '5px' }}>
-          {inboxList.map((message) => (
-            <MessagePreview key={uuidv4()} message={message} />
-          ))}
-        </div>
-      )}
+      {loadMessages ? <div>Loading Messages...</div> : <PaginatedMessages messages={inboxList} />}
```
xscottxbrownx, line 74: This doesn't seem to occur on refresh, same in `Outbox`.

src/components/Messages/Pagination.jsx
```diff
@@ -0,0 +1,83 @@
+// React Imports
+import React, { useState } from 'react';
+import ReactPaginate from 'react-paginate';
+// Styling Imports
+import styled from 'styled-components';
```
xscottxbrownx, line 5: Not an issue, as we'll get rid of both, but import from MUI:

"import { styled } from '@mui/material/styles';"

src/routes/Outbox.jsx
```diff
@@ -0,0 +1,78 @@
...
+  const podUrl = session.info.webId.split('profile')[0];
+
+  // Handler function for refreshing PASS outbox
+  const handleOutboxRefresh = async () => {
+    const messagesInSolid = await getMessageTTL(session, 'Outbox', outboxList, podUrl);
+    messagesInSolid.sort((a, b) => b.uploadDate - a.uploadDate);
+    setOutboxList(messagesInSolid);
+  };
+
+  // Re-sorts messages upon outboxList updating
```
xscottxbrownx, line 42: We should get this into `Inbox` as well. It doesn't update automatically.

src/components/Messages/Pagination.jsx
```diff
@@ -0,0 +1,83 @@
...
+    </>
+  );
+};
+
+// Helper component for displaying paginated results
+const Messages = ({ currentMessages }) =>
+  currentMessages &&
+  currentMessages.map((message) => <MessagePreview key={uuidv4()} message={message} />);
+
+const PaginationContainer = styled.div`
```
xscottxbrownx, line 56: Could be future PR, but using some provided classes from `react-paginate` to style the pagination controls.

Maybe make the `prev` and `next` the red color, and the page numbers background be a version of the main green (like `primary.light`), then the active page be the main green with the yellow number?

Screenshot 2023-06-06 at 8 51 21 AM

src/components/Messages/MessagePreview.jsx
xscottxbrownx, line 1: Seems like a lot of wasted space on desktop (to the right.)

Not sure of a suggestion yet, but maybe show a snippet of the message?
Screenshot 2023-06-06 at 8 55 26 AM

src/App.jsx
leekahung: Yes, this specific call will only be used during login. It's reliant on podUrl to run right now, but it should be fixed in #214 when we can call podUrl through a context.

There's another useEffect routine in the new components so I'm thinking of just refactoring it all there once that's fixed.

I'll make note of it.

src/routes/Outbox.jsx
leekahung: Think we can use the setLoadMessages to be triggered to true on refresh and false when the new data is fetched.

leekahung: Included loading on refresh. Example for outbox is shown below, but have the same fix for inbox.

https://github.com/codeforpdx/PASS/assets/14917816/9884f5d5-9e56-4879-8eec-906fda1e90c5

src/AppRoutes.jsx
leekahung: Think so. Was planning to wrap them up as different routes like /PASS/messages/inbox and /PASS/messages/outbox

src/components/Messages/Pagination.jsx
leekahung: Think we can force the text to be wrapped within the container.

leekahung: Actually, is that one long continuous string? If the message is coming out in sentences, it seems to work okay.

https://github.com/codeforpdx/PASS/assets/14917816/519bcbb3-0048-47d7-9df2-eafe7ed0b1f0

But think we should handle this edge case.

leekahung: Fixed with style={{wordWrap: "break-word"}} for edge case with long string.

src/components/NavBar/NavbarLinks.jsx
leekahung: Think we can refactor this in another PR.

src/routes/Inbox.jsx
leekahung: Fixed.

src/components/Messages/Pagination.jsx
leekahung: Noted.

src/routes/Outbox.jsx
leekahung: For inbox, it's kind not really possible to automatically be updated outside of initial login or refresh, unless we use some kind of setInterval to refresh the list at a set time. Could open this up as an new issue.

It's because messages in inbox is not dependent on the recipient. It's depending on the sender.

src/components/Messages/MessagePreview.jsx
leekahung: Not sure at the moment. Could definitely consider moving the date over to the right and have a snippet in the middle in the future.

src/components/Messages/Pagination.jsx
leekahung: It's pretty customizable, think we can certainly mess with it on an update.

timbot1789 (commented):
Would love some unit tests for all these

leekahung:
> Would love some unit tests for all these

Would you like to have the unit tests in this PR or you think it'll be okay if we leave it in a follow-up PR?

timbot1789:
> > Would love some unit tests for all these
> 
> Would you like to have the unit tests in this PR or you think it'll be okay if we leave it in a follow-up PR?

Those tests look good! Thanks. In the future we can add tests for the UI as well, since it looks like the UI here is mostly a placeholder. Although it looks like one of them is failing.

Unit tests should be included in the PR that makes the code changes. Part of the PR process is demonstrating that the code works as described, and that it doesn't cause any regressions in working code. Unit tests help us do that. If the new unit tests describe the intended behavior of the PR, they all pass, and no old unit tests fail, we can be fairly certain the new code works correctly. If they're committed separately, we don't get that assurance.

src/routes/Outbox.jsx
timbot1789: Yeah updating inbox dynamically would require hooking up to Solid's notification system or something similar. That's a fairly involved networking process, and IMO can be left to another PR.

src/model-helpers/User.js
leekahung, line 1: There seem to have been a problem with active.ttl being updated. Added these fixes to updateUserActivity.

leekahung:
> Those tests look good! Thanks. In the future we can add tests for the UI as well, since it looks like the UI here is mostly a placeholder. Although it looks like one of them is failing.

Think I've found the failing one from credentials-helper. I've updated it based on what you've changed in your recent branch and the test passed.

> Unit tests should be included in the PR that makes the code changes. Part of the PR process is demonstrating that the code works as described, and that it doesn't cause any regressions in working code. Unit tests help us do that. If the new unit tests describe the intended behavior of the PR, they all pass, and no old unit tests fail, we can be fairly certain the new code works correctly. If they're committed separately, we don't get that assurance.

Okay! I'll make a note of it to try and implement unit tests along with my PRs.

src/App.jsx
timbot1789: @xscottxbrownx this is harder to do than it seems. Solid is based around the idea that the user, not the application, controls all data on the pod. PASS is just a guest in the user's space, and the user can delete data created by PASS at any time. That means PASS will have a hard time assuming that a file it creates on the user's pod will still be there 2 minutes later.

So even if we do run this code on first login only, the user may delete the files before second login. Then we have a problem.

The code for `createPublicContainer` and its siblings I think handles the situation correctly. It has a `try{}catch{}` block. Try to fetch the resource, and if the resource isn't there, create it. Although we probably don't need to run this setup on login. We can probably delay it to the first time the application tries to fetch data from that location.

timbot1789 (approved):
This all looks good as a first draft of a messaging system

Some ideas for future enhancements:

- Clean up the UI more. Scott has some good suggestions for that
- Allow inboxes to update when a new message is received in a pod. We should be able to use Solid's notification protocol for that
- Add ability to delete/archive old messages
- Add conversations / reply-to
- Add message sending status. Has it been received or not?
- Sort messages by active client
- Add database pagination. Currently, we're getting all the messages a user has ever received in one single server response, and then simply displaying them in a paginated fashion. This may become a performance bottleneck if a user has a large amount of messages (like hundreds). SPARQL queries should be able to help us paginate these. Not sure what kind of support they have serverside though.

PR #192, opened 2023-05-29
tried fixing the login button

anubhav1206:


leekahung:
Hey @anubhav1206 thanks for your interest in committing changes to the project!

However, we currently have a system set up such that we only update markdown files and the such on Master. Please try again from a feature branch and merge it into Development instead

PR #198, opened 2023-05-30
Issue 197/refactoring-routes

### LINKED ISSUE #197: Refactoring PASS routing with React Router
While the number of "pages" and routes are currently smaller, it'll eventually needs to be revamped and organized in order to handle future expansions. What we have now in App.jsx is a good start to get some basic routing work done, but it's already getting a bit cluttered.

Refactor Routes into its own file and pass it into App.jsx. While we're at it, attempt to optimize existing routes with the concept of "Protected Routes" with React Router.

leekahung:
This PR includes the refactoring of React Routes for PASS and the optimization of current routing logic.

### DIFF
--- a/src/components/Users/index.js
+++ b/src/components/Users/index.js
@@ -1,4 +1,0 @@
-import UserSection from './UserSection';
-import UsersList from './UsersList';
-
-export { UserSection, UsersList };
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,6 +1,5 @@
 // React Imports
 import React, { useEffect, useMemo, useState } from 'react';
-import { Routes, Route, Navigate } from 'react-router-dom';
 // Inrupt Imports
 import { useSession } from '@inrupt/solid-ui-react';
 // Utility Imports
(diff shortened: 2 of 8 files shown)


xscottxbrownx (commented):
**Overall looks great.**

I couldn't find much about it so want to ask here,  is the `replace` attribute used to replace the navigation history?

---

The only other thing that I'd like to consider is whether or not you felt it appropriate if we move the rest of the routes into the `routes` folder with this PR?

`Forms`, `Inbox`, and `UserSection`

leekahung:
> I couldn't find much about it so want to ask here, is the `replace` attribute used to replace the navigation history?

Though I don't think a typical user would use the back button a lot from their browser for PASS, it's simply there to reset the local history stack if a user logs out or refresh.

> The only other thing that I'd like to consider is whether or not you felt it appropriate if we move the rest of the routes into the `routes` folder with this PR?
> 
> `Forms`, `Inbox`, and `UserSection`

Think we can do that in this PR.

src/components/Users/index.js
leekahung, line 1: Removed this since UserList is by itself at this time after moving UserSection to routes. No need to use a named import until there are more components in Users.

xscottxbrownx:
All looks good to me now.

xscottxbrownx (approved):
love the refactored routes, and moving them to the `routes` folder.

PR #201, opened 2023-05-31
Issue 200/update generate container functions

### LINKED ISSUE #200: Updating generateUserList, createDocumentContainer, and createOutbox functions
**Describe the bug**
The three functions generateUserLIst, createDocumentContainer, and createOutbox functions in session-core were meant to only run once when the user first register their pod on Solid or if the user had deleted the "Documents", "Users", or "Outbox" containers. However, there's a significant lag after login for user's who already have those containers set up in their Pods. It indicates that these functions are running even after they're created.

**Expected behavior**
To have the functions trigger only if the user logs into their Solid Pod for the first time with PASS or if the user manually deletes the "Documents", "Users", or "Outbox" containers on their Pod.

**Potential Solution**
Use the getSolidDataset function from Inrupt to see if the container exist and place it within a try/catch.

If the containers does exist from the getSolidDataset function, it'll simply exist the function without running anything else. If it doesn't, it'll throw an error which is caught within the catch and generates the missing containers.

leekahung:
# PR Details

This PR provides updates to three functions in session-core and create two new functions for createInbox and createPublicContainer in case these PASS specific containers does not exist. All these functions are suppose to only run if container does not exist within the specified Solid Pod upon login.

The update enforces this with a try/catch where the functionality to generate container and relevant files/directories only run if container doesn't exist.

This was previously attempted using the hasTTLFiles and an early createContainerAt to generate the container, but the function should really be checking on whether the container exist first before executing this.

This mean, we don't really need the hasTTLFile function at this point, all conditions could be handled by Inrupt's getSolidDataset function. Inrupt's getSolidDataset function throws a 404 error if container doesn't exist, but gives 200 if it does. This gives us a clean way to run our try/catch, where try will simply exit the function and catch will generate our container if it doesn't exist.


src/App.jsx
```diff
@@ -99,7 +104,7 @@ const App = () => {
 
     if (session.info.isLoggedIn) {
       localStorage.setItem('loggedIn', true);
-      fetchData();
+      setupPod();
```
timbot1789, line 87: Does this function need to be called on every login? I feel like we can call this once, the first time a user enters PASS, setup any files we need, then never call it again.

leekahung: It's for cases when users manually delete their containers on their Pods (whether the reason may be).

Regardless, the functions here are made to trigger only if the container doesn't exist in their Pod (typically when logging in for the first time), so I think it's fine as is.

src/utils/network/session-core.js
```diff
@@ -534,5 +559,37 @@ export const createOutbox = async (session) => {
...
+ *
+ * @memberof utils
+ * @function createInbox
+ * @param {Session} session - Solid's Session Object {@link Session}
+ * @returns {Promise} Promise - Generates an outbox for Pod upon log in if
+ * user's Pod does not have the an outbox to begin with
+ */
+
+export const createInbox = async (session) => {
+  const inboxContainerUrl = getContainerUrl(session, 'Inbox', 'self-fetch');
```
timbot1789: Now that we have the signed-in user's pod url, could we pass that in and generate inbox Url off of that?

src/App.jsx
timbot1789: That sounds fine then

timbot1789 (commented):
Would love to see some unit tests

src/utils/network/session-core.js
leekahung: Yeah, I think we could start replacing these functions with podUrl

leekahung:
> Would love to see some unit tests

Not sure if what I've done is enough, but I've included a unit test for the kind of logic I'm using for the create container functions (createDocumentContainer, createPublicContainer, createInbox, and createOutbox).

test/utils/session-core.test.js
```diff
@@ -0,0 +1,249 @@
...
+  });
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("prints out correct documentUrl for Driver's License and webId for other user", () => {
+    const fileType = "Driver's License";
+    const otherPodUsername = 'pod2';
+    const documentUrl = getContainerUrl(session, fileType, INTERACTION_TYPES.SELF);
+    const webId = `https://${otherPodUsername}.${
```
timbot1789: You generally shouldn't need complex string parsing like this in unit tests. If there's a string you want to declare, just declare it.

`const webId = 'https://pod2.example.com/profile/card#me'`

Or, since our system doesn't actually care about webId format:
```
const webId = `${mockSolidIdentityProvider}pod2`
```

test/utils/session-core.test.js
```diff
@@ -0,0 +1,249 @@
...
+      mockSolidIdentityProvider.split('/')[2]
+    }/profile/card#me`;
+
+    expect(documentUrl).toBe('https://pod.example.com/Drivers%20License/');
+    expect(webId).toBe('https://pod2.example.com/profile/card#me');
```
timbot1789: Not sure what you're testing here. The webId is generated from test data. It doesn't get any data from`getContainerUrl`

test/utils/session-core.test.js
```diff
@@ -0,0 +1,249 @@
...
+  const actual = await vi.importActual('@inrupt/solid-client');
+  return {
+    ...actual,
+    getSolidDataset: vi.fn((url) => Promise.resolve(mockSolidDatasetFrom(url))),
+    createContainerAt: vi.fn((url) => Promise.resolve(mockSolidDatasetFrom(url))),
+    saveSolidDatasetInContainer: vi.fn((url) => Promise.resolve(mockSolidDatasetFrom(url)))
+  };
+});
+
+const getSolidDatasetRejected = vi.fn(() => Promise.reject(Error('dataset not found')));
```
timbot1789: It's probably not the best idea to globally declare mocks like this. You'll get confusing behavior when it gets cleared by `vi.clearAllMocks()`. You can just declare the mocks when you use them.

test/utils/session-core.test.js
leekahung: Wanted to see if the parsing is correct for that test case (it's in the original function). If webId comes out wrong, it'll error out.

test/utils/session-core.test.js
```diff
@@ -0,0 +1,249 @@
...
+      `${containerUrl}Passport/`,
+      `${containerUrl}Drivers%20License/`
+    ];
+
+    const webId = `https://${otherPodUsername}.${
+      mockSolidIdentityProvider.split('/')[2]
+    }/profile/card#me`;
+
+    urlsToSet.forEach(async (url) => {
+      await setDocAclForUser(session, url, 'update', webId, permissions);
```
timbot1789: This doesn't call the `setDocAclForUser` from `session-core`. It calls the variable you defined on line 56.

test/utils/session-core.test.js
leekahung: Alright, I'll pull out the ones I'm using here outside of these global declarations.

test/utils/session-core.test.js
```diff
@@ -0,0 +1,249 @@
...
+    };
+  });
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("prints out correct documentUrl for Driver's License and webId for other user", () => {
+    const fileType = "Driver's License";
+    const otherPodUsername = 'pod2';
+    const documentUrl = getContainerUrl(session, fileType, INTERACTION_TYPES.SELF);
```
timbot1789: If you can avoid it, don't call out to other functions in unit tests unless you're trying to build a complex piece of data. It makes the unit tests harder to understand.

This can just be:

`const documentUrl = 'https://pod.example.com/Drivers%20License/'`

test/utils/session-core.test.js
leekahung: At the moment the function setDocAclPermission currently uses the other person's username to build a webId. The test here is to check if the generated variables correct because being passed into setDocAclForUser:

```JavaScript
export const setDocAclPermission = async (session, fileType, permissions, otherPodUsername) => {
  const documentUrl = getContainerUrl(session, fileType, INTERACTION_TYPES.SELF);
  const webId = `https://${otherPodUsername}.${
    SOLID_IDENTITY_PROVIDER.split('/')[2]
  }/profile/card#me`;

  await setDocAclForUser(session, documentUrl, 'update', webId, permissions);
};
```

Could add an expect for setDocAclForUser that it's expected to called with expected values

test/utils/session-core.test.js
```diff
@@ -0,0 +1,249 @@
...
+      info: {
+        webId: `${mockPodUrl}profile/card#me`
+      }
+    };
+  });
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('calls setDocAclForUser 4 times', async () => {
```
timbot1789: What are you trying to test here? What's the functionality of `setDocContainerAclPermissions` that you're interested in?

test/utils/session-core.test.js
leekahung: Can do. I'll swap out the getContainerUrl function for the mock value. I'll add the unit test for getContainerUrl instead in session-helper.test.js

test/utils/session-core.test.js
leekahung: Wanted to check if setDocAclForUsers are called for all the urls found in urlsToSet

test/utils/session-core.test.js
leekahung: The setDocAclForUser here is a mock function

leekahung:
Alright, think I've found where the issues were from our discussion. Updated unit tests based on that.

test/utils/session-core.test.js
timbot1789: If the code is in the original function, then you should generate the ID by calling that function. That way, if we ever change how we generate webIds in code, the tests will automatically update with the changes.

test/utils/session-core.test.js
```diff
@@ -0,0 +1,282 @@
...
+      info: {
+        webId: `${mockPodUrl}profile/card#me`
+      }
+    };
+  });
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('prints out correct documentUrl and webId for other user and calls setDocAclForUser', async () => {
```
timbot1789: Unfortunately this test still isn't actually testing anything. All the `expect` statements only check data generated in the test. The `expect` on line 42 is checking the `documentUrl` value set on line 34. The `expect` on line 43 is checking the `webId` value from line 35, and the `expect` on line 41 is checking that the function call you made on line 39 has the same parameters as what you passed in on line 39.

I'd expect the test to look something like this:

```
const result = await setDocAclForUser(params);

expect(saveAclFor).toBeCalledWith(testParams); // Whatever you expect to be passed to Solid
expect(result).toMatchObject(object); // Whatever you expect the return value to look like, if there is any

File: test/utils/session-core.test.js [hunk 10]
```diff
@@ -0,0 +1,282 @@
...
+} from '@inrupt/solid-client';
+import { expect, vi, it, describe, afterEach, beforeEach } from 'vitest';
+import { getDocuments } from '../../src/utils/network/session-core';
+
+const mockPodUrl = 'https://pod.example.com/';
+const mockSolidIdentityProvider = 'https://example.com/';
+let session = {};
+
+describe('setDocAclPermission', () => {
+  const setDocAclForUser = vi.fn(() => Promise.resolve());
```
timbot1789: This line should be deleted. It'll interfere with the tests.

test/utils/session-core.test.js
timbot1789: Correct, but it probably shouldn't be. You generally only want to mock external services like databases, networks, etc. In our case, we generally should only be mocking the Inrupt services that interact with the solid servers. Our own code generally shouldn't be mocked.

test/utils/session-core.test.js
```diff
@@ -0,0 +1,282 @@
...
+      info: {
+        webId: `${mockPodUrl}profile/card#me`
+      }
+    };
+  });
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('expect try block to be ran if getSolidDataset resolves', async () => {
```
timbot1789: This one also doesn't test anything outside the test. It just runs a function you defined on line 91.

test/utils/session-core.test.js
```diff
@@ -0,0 +1,282 @@
...
+
+      expect(mockGetSolidDataset).toBeCalled();
+    } catch {
+      catchExecuted = true;
+    }
+
+    expect(catchExecuted).toBe(false);
+  });
+
+  it('expect catch block to be ran if rejected', async () => {
```
timbot1789: Same as above

test/utils/session-core.test.js
```diff
@@ -0,0 +1,282 @@
...
+  beforeEach(() => {
+    session = {
+      fetch: vi.fn()
+    };
+  });
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('run try only if container exists', async () => {
```
timbot1789: This also only tests functions defined in the test

test/utils/session-core.test.js
```diff
@@ -0,0 +1,282 @@
...
+    } catch {
+      catchExecuted = true;
+      await createContainerAt(documentUrl, { fetch: session.fetch });
+    }
+
+    expect(createContainerAt).not.toBeCalled();
+    expect(catchExecuted).toBe(false);
+  });
+
+  it('executes catch if getSolidDataset fails', async () => {
```
timbot1789: This also doesn't test anything outside the test

test/utils/session-core.test.js
```diff
@@ -0,0 +1,282 @@
...
+      info: {
+        webId: `${mockPodUrl}profile/card#me`
+      }
+    };
+  });
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('expect createContainerAt to be called 3 times', async () => {
```
timbot1789: This also doesn't test anything outside the test

test/utils/session-helper.test.js
```diff
@@ -87,3 +90,57 @@ describe('setDocAclForUser', () => {
     expect(resourceAcl).toEqual(mockResourceAcl);
   });
 });
+
+describe('getContainerUrl', () => {
```
timbot1789, line 68: These tests are good.

test/utils/session-core.test.js
leekahung: Removing test for this.

test/utils/session-core.test.js
leekahung: Removing test for this.

test/utils/session-core.test.js
leekahung: Removing test for this.

test/utils/session-core.test.js
leekahung: Removing test for this.

test/utils/session-core.test.js
leekahung: Removing test for this.

test/utils/session-helper.test.js
leekahung: Thanks, I'll see what I can do about the tests in session-core.

test/utils/session-core.test.js
leekahung: I'll try to update this to call the actual setDocAclForUser, but it doesn't return anything, so I just expect the saveAclFor to be called?

test/utils/session-helper.test.js
```diff
@@ -1,7 +1,31 @@
...
+} from '../../src/utils/network/session-helper';
 import { INTERACTION_TYPES } from '../../src/constants';
 
+const mockPodUrl = 'https://pod.example.com/';
+const mockSolidDataset = addMockResourceAclTo(mockSolidDatasetFrom(mockPodUrl));
+let session = {};
+
+vi.mock('@inrupt/solid-client');
+
+vi.mock('@inrupt/solid-client', async () => {
```
timbot1789: You can remove lines 21 - 27 if you merge in the changes I made in that small unit test PR. The case you're trying to handle here will be handled by line 19.

test/utils/session-helper.test.js
```diff
@@ -22,3 +46,88 @@ describe('createResourceTtlFile', () => {
...
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("returns false if there's no TTL file in solidDataset", () => {
+    const result = hasTTLFiles(mockSolidDataset);
+    expect(result).toBe(false);
+  });
+
+  it("returns true if there's something in solidDataset", () => {
```
timbot1789, line 60: You can write this test a bit more cleanly (without using mocks) with the following code:

```
  it("returns true if there's something in solidDataset", () => {
    const thing = buildThing(createThing('test')).build();

    const result = hasTTLFiles(setThing(mockSolidDataset, thing));
    expect(result).toBe(true);
  });
```

There is something in the solidDataset (a test thing), therefore it should return true. 

Oddly though, this test fails. I think it actually reveals a bug in the function itself. That's good. That's what unit tests are supposed to do.

The function seems to be wanting to test if there is any data in a dataset file. It does this by:

a) fetching all the `things` in a dataset
b) checking if any of the URLs of those things end in a `.ttl`

However, that's not usually how these URLs are created in the `solid-client` library. The `.ttl` file refers the _dataset_, not the _thing_. The thing is referred to by `.ttl#thing_id`, so the `thing` url should never end in `.ttl`

test/utils/session-core.test.js
```diff
@@ -0,0 +1,103 @@
...
+  deleteDocumentFile,
+  getDocuments
+} from '../../src/utils/network/session-core';
+
+const mockPodUrl = 'https://pod.example.com/';
+let session = {};
+
+vi.mock('@inrupt/solid-client');
+
+vi.mock('../../src/utils/network/session-helper', async () => {
```
timbot1789, line 15: It's generally best to keep mocks to just things like network access, database access, and other things we don't control. Mocking our own functions like this lowers test coverage, and makes it harder to tell if a change to a function signature breaks other parts of the system.

test/utils/session-helper.test.js
leekahung: I'm a bit confused about this. The function seems to be working as intended after checking the functions again with console.logs.

https://github.com/codeforpdx/PASS/assets/14917816/ae335ce4-81bc-4f93-b1fa-e2b6cef382a3

test/utils/session-helper.test.js
leekahung: I'll need a getThingAll which mocks getting a list of Things.

timbot1789 (approved):

leekahung:
Got the unit tests discussed with Tim.  Have either move mocks close to the actual unit test rather than having them set globally unless necessary (i.e. being used in all tests) and clearing up/removing unneeded mocks from unit tests.

PR #204, opened 2023-06-01
Refactoring custom hooks

### LINKED ISSUE #203: Refactoring hooks directory
Not much of an issue at the moment because we have so little custom hooks, but could try refactoring existing index.js.

Could refactor the existing custom hooks into their own file and export them into index.js for imports.

leekahung:
# PR Details

Refactoring out each custom hook into their own file and using index.js to export them as named imports.


timbot1789 (commented):
It seems like all this PR is doing is placing these functions in separate files. What advantage does this have?

leekahung:
> It seems like all this PR is doing is placing these functions in separate files. What advantage does this have?

It's not really an advantage per se, think it'll be easier to see and manage the different hooks if they're in their own files.

timbot1789 (approved):
LGTM

PR #209, opened 2023-06-03
Issue 196/clients page

### LINKED ISSUE #160: Rename UserList/UserSection to ClientList/ClientSection
For PASS, UserList/UserSection was intended to help case-workers manage a list of clients. These components would eventually be hidden once we figure out how to distinguish different users using PASS. While clients are technically "users", it'll be more intuitive to rename these as ClientList/ClientSection.

Simply rename sections related to UserList/UserSection to ClientList/ClientSection.

### LINKED ISSUE #196: Create CLIENTS page (1st iteration) - from "Home" (UsersSection)
# Describe Issue/Goal:

Goal is to get a more proper MVP of the page that displays once a caseworker is logged in (currently the "Home" link in Navbar.)

Most likely using some form of list/table to show all clients of said caseworker, then can start linking any of the `forms`/functions of the forms that may make sense/be required for this page.

---
Screenshot 2023-05-29 at 6 09 16 PM
Screenshot 2023-05-29 at 6 09 33 PM

xscottxbrownx:
# This PR is addressing ISSUE #196 

I don't want this PR to grow out of control, so I am uploading it here as a draft to get feedback.  Basically how much this single PR/Issue should address, any thoughts on alternative approaches to some issues currently in this code.

---

### Things that are left:

-make clientlist load quicker
- make the key unique for each row/ each client (it's using PodURL right now, assume we will switch to web ID when we can to solve)
- make the "priority" field on .ttl file (boolean true/false) so pin icon can work properly
- better style the Add Client pop-up modal (I just moved the whole adduser form into it. But can be cleaned up for better user experience.
- the statusnotification at bottom of table can be changed to a better look/user experience (maybe it's own modal message?)
- better approach for select checkbox (don't want to select all with a single click here, and is .ttl file field necessary for this too?
- make the alert popup on delete icon click be a mui modal component as well to match

---

Need some more eyes, direction, and thoughts on it all at the moment.


xscottxbrownx:
@leekahung @andycwilliams @timbot1789 @benzelcurry whenever you have time to take a look and make any comments of thoughts…

leekahung:
At the moment, I think this looks good for a first iteration (styling aside). It's certainly headed in the right direction, moving add users/clients to a modal and userList into a workable "dashboard" like page.

I'll touch on points which I think you're looking for feedback to on top of just the general direction of the PR. The points I didn't touch on, I think they're on-point for what we could/should do. If you believe it'll be too much for this PR, feel free to leave TODOs comments. We can certainly address them in a follow-up.

---

> make the key unique for each row/ each client (it's using PodURL right now, assume we will switch to web ID when we can to solve)

Think we can use webId as the unique key in the future since webIds themselves were meant to be fully unique when compare to username, which would no longer be unique in the case of multiple Solid Providers.

> the statusnotification at bottom of table can be changed to a better look/user experience (maybe it's own modal message?)

StatusNotification would need to be overhauled eventually both UI-wise and Development-wise, but we can leave that for later in case this gets too large. Nevertheless, some general updates to styling are more than welcomed.

> better approach for select checkbox (don't want to select all with a single click here, and is .ttl file field necessary for this too?

For the select checkbox, which was for picking specific users to fill out/auto-populate forms within just PASS, I think it'll be better to store the boolean in a local state with useState.

----

On other things, I've noticed data used for rendering the user list, I don't think we'll need to worry at the moment. It can be solved after resolving conflicts between branches and removing the old userlist.ttl from Solid.

The updated userlist.ttl using webId has a slightly different format than that the userlist.ttl with usernames and has additional fields, which is why some fields are populated correctly, but others does not.

timbot1789 (commented):
Judging from the issue, I like the design of this a lot. I think we could break this up into 3-ish PRs:

- One PR simply renaming everything from `User` to `Client`
- One updating the list display
- One moving the current user registration form into the modal

Also I merged a change that changes the user's backend, so you'll need to merge those changes into this PR.

timbot1789 (commented):
Some of your questions:

> Make clientlist load quicker

The new updates in Development now should help with that. However, the UI should be independent of resource loading, so it shouldn't be a concern for this PR

> Make the `priority` field on .ttl file (boolean true/false)

That's easy to add, but it may be best to represent it in the ttl file as a number, so if in the future we need to add varying priorities, we can.

> better style to add client

We need to fully rework client signup anyway. I'm working on that in a separate PR. I wouldn't worry about it too much here

src/components/Clients/ClientListTable.jsx
```diff
@@ -0,0 +1,167 @@
...
+  };
+
+  const handlePinClick = () => {
+    // TODO: change from a state to a key/value field of the .ttl file (priority: true/false)
+    // TODO: add function to move the row to the top of list if priority: true
+    setPinned(!pinned);
+  };
+
+  // ======= MAKE ANY COLUMN HEADER CHANGES HERE =======
+  const columnTitlesArray = ['Select', 'Client', 'Pod URL', 'Last Activity', 'Pin', 'Delete'];
```
timbot1789: `Pod URL` should be renamed `WebId`

xscottxbrownx:
`Development` branch's `ManageUsers.jsx` is my `AddClient.jsx`...

should I just be moving over the changes from `ManageUsers`?  Seems to work on `Development`.  BUT seems I can add anything, like there is no check if it is a real user?  Didn't do that before.

---

@leekahung @timbot1789

src/components/Clients/ClientListTable.jsx
xscottxbrownx: done

xscottxbrownx:
> The updated userlist.ttl using webId has a slightly different format than that the userlist.ttl with usernames and has additional fields, which is why some fields are populated correctly, but others does not.

Where can I see these fields/keys? I use them in the table to fill the cells, and need to figure out what they have been renamed to or what the possibilities are.

leekahung:
> `Development` branch's `ManageUsers.jsx` is my `AddClient.jsx`...
> 
> should I just be moving over the changes from `ManageUsers`?  Seems to work on `Development`.  BUT seems I can add anything, like there is no check if it is a real user?  Didn't do that before.
> 
> ---
> 
> @leekahung @timbot1789 
> 

There's the userListObject in UserList.jsx in Development. The related fields are in src/model-helper/Users.js Line 148.

`{ person, username, givenName, familyName, webId, podUrl, dateModified }`

xscottxbrownx:
> There's the userListObject in UserList.jsx in Development. The related fields are in src/model-helper/Users.js Line 148.
> 
> `{ person, username, givenName, familyName, webId, podUrl, dateModified }`

THANK YOU!

So, there is still `podUrl`?  Based on Tim's comment further up I assumed we weren't using, or it was gone.

---

And where did these go? Or where are they rendered unnecessary and so deleted?

Screenshot 2023-06-04 at 8 15 42 AM
Screenshot 2023-06-04 at 8 16 09 AM

leekahung:
>THANK YOU!
>
>So, there is still `podUrl`?  Based on Tim's comment further up I assumed we weren't using, or it was gone.

Yeah, podUrl is still there. It's being used under the hood to run other functions

leekahung:
> And where did these go? Or where are they rendered unnecessary and so deleted?
> 
> Screenshot 2023-06-04 at 8 15 42 AM Screenshot 2023-06-04 at 8 16 09 AM

Yeah, a lot of the functions related to the User/UserList has been gutted from session-core.

Functions like deleteUsersFromPod has been replaced with functions found in src/model-helper/. Should be able to remove the older functions safely from the code.

timbot1789:
> `Development` branch's `ManageUsers.jsx` is my `AddClient.jsx`...
> 
> should I just be moving over the changes from `ManageUsers`? Seems to work on `Development`. BUT seems I can add anything, like there is no check if it is a real user? Didn't do that before.
> 
> @leekahung @timbot1789

I've got no issue with you moving the manage user changes to add client. Just be sure to keep the webid field and associated functionality. Should still hook into the userlist context.

I'm not sure we've ever checked to make sure the user actually exists, but either way I don't think we should worry too much about checking that kind of stuff while we're in the mvp stage. We can think about data validations once we have a clearer sense of how this system will be used, and what data is really needed.

xscottxbrownx:
> > `Development` branch's `ManageUsers.jsx` is my `AddClient.jsx`...
> > should I just be moving over the changes from `ManageUsers`? Seems to work on `Development`. BUT seems I can add anything, like there is no check if it is a real user? Didn't do that before.
> > @leekahung @timbot1789
> 
> I've got no issue with you moving the manage user changes to add client. Just be sure to keep the webid field and associated functionality. Should still hook into the userlist context.
> 
> I'm not sure we've ever checked to make sure the user actually exists, but either way I don't think we should worry too much about checking that kind of stuff while we're in the mvp stage. We can think about data validations once we have a clearer sense of how this system will be used, and what data is really needed.

I was just commenting that it wasn't the case before, I could only add legitimate users.  But found the issue most likely. Ka Hung reminded me that I forgot about your notice and needed to delete original .ttl file

xscottxbrownx:
## Updated to the new Development branch changes

think this is a decent first iteration, *unless anybody has any quick thoughts on an approach for both the checkboxes and the pin icons...*

Currently, the checked/not checked is controlled by a state which is why all checkboxes get checked when one does. Same goes with the 2 different pin icons (outlined/red fill.)

---

I believe other issues besides the above should be handled in future PR's, BUT I'm open to any thoughts/feedback.

leekahung:
> Currently, the checked/not checked is controlled by a state which is why all checkboxes get checked when one does. Same goes with the 2 different pin icons (outlined/red fill.)

Have you tried out using an array of booleans as the initial state? You could associate the index from the array to a person and the value of the boolean in the array as the trigger for check/unchecked.

For example, we get [false, false] for our array of two people index 0 is for person 1 and index 1 is for person 2. If checkbox for person 1 is selected (with value set to index 0 in the checkbox), the array will update to [true, false] and only the first checkbox will be checked.

xscottxbrownx:
> Have you tried out using an array of booleans as the initial state? You could associate the index from the array to a person and the value of the boolean in the array as the trigger for check/unchecked.
> 
> For example, we get [false, false] for our array of two people index 0 is for person 1 and index 1 is for person 2. If checkbox for person 1 is selected (with value set to index 0 in the checkbox), the array will update to [true, false] and only the first checkbox will be checked.

The issue I ran into/thought of for this idea is that I don't understand how the userList/clientList is created or at least displayed. Adding people and deleting them sometimes seems to place them in random spots in the table, not always adding them to top or bottom.  Almost seems to re-arrange sometimes.

## Example

I added 5 test clients (and at adding number 5 they got out of order):
Screenshot 2023-06-05 at 2 32 31 PM

Then after deleting "test 5" this is what table looks like:
Screenshot 2023-06-05 at 2 32 52 PM

---

If you have any thoughts on what to check for here, I'm all ears. Seems strange.

timbot1789:
> > Have you tried out using an array of booleans as the initial state? You could associate the index from the array to a person and the value of the boolean in the array as the trigger for check/unchecked.
> > For example, we get [false, false] for our array of two people index 0 is for person 1 and index 1 is for person 2. If checkbox for person 1 is selected (with value set to index 0 in the checkbox), the array will update to [true, false] and only the first checkbox will be checked.
> 
> The issue I ran into/thought of for this idea is that I don't understand how the userList/clientList is created or at least displayed. Adding people and deleting them sometimes seems to place them in random spots in the table, not always adding them to top or bottom. Almost seems to re-arrange sometimes.
> 
> ## Example
> I added 5 test clients (and at adding number 5 they got out of order): Screenshot 2023-06-05 at 2 32 31 PM
> 
> Then after deleting "test 5" this is what table looks like: Screenshot 2023-06-05 at 2 32 52 PM
> 
> If you have any thoughts on what to check for here, I'm all ears. Seems strange.

There's no enforcement of any order right now for the clients list in the backend. They can be received from the backend in any order, and in a different order every request. You can enforce an order in that table if you sort on a column, or if you sort the `userList` array.

xscottxbrownx:
Feel pretty good about where it's at now for a first iteration...

---

**These are the issues I can think of and would create to enhance this more:**

- Sort clientList array before displaying table (may need to do when delete and add as well.)
- Update the add client modal so buttons work as intended/styled better
- Make a function to run onClick of pin icon to move the client to the top of the table (this will most likely need to be done after clientList array sorting is done.)
- Update the UI of how the StatusNotification is shown (maybe small modal with auto-disappear when complete?)
- Make a "Loading clients..." placeholder to show when the table data is being loaded (currently it shows the "Add Clients to Your List" as if there isn't a list at all.
- Get rid of styled components styling and switch to pure MUI in `ClientListTable.jsx` and `ClientListTableRow.jsx`
- Make the client name a clickable link (the webID field link), and replace the webID field with a Notes field somehow.
- Update UI to optimize for mobile view as well.

src/components/Clients/AddClient.jsx
leekahung, line 1: I think you've already mentioned it as a potential issue to work on, but yes, we'll probably would want to merge AddClient.jsx together with AddClientModal.jsx. Had been confused earlier that the Add Client button at the bottom was just a placeholder for closing the modal like Cancel and the actual Add Client button is still inside the AddClient form.

Screen Shot 2023-06-06 at 12 20 44 (2)

src/components/Clients/ClientList.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * ClientList Component - Component that generates ClientList section for PASS
+ * which interfaces with Solid Pod to fetch user list
+ *
+ * @memberof Clients
+ * @name ClientList
+ */
+
+const ClientList = ({ loadingActive }) => {
```
leekahung: The loadingActive state in Line 20 and 26 is no longer in use. Can be removed without affecting it.

src/components/Clients/ClientListTable.jsx
```diff
@@ -0,0 +1,87 @@
...
+import ClientListTableRow from './ClientListTableRow';
+
+/**
+ * ClientListTable Component - Component that generates table of clients from data within ClientList
+ *
+ * @memberof Clients
+ * @name ClientListTable
+ */
+
+const ClientListTable = ({ loadingActive, statusType, defaultMessage }) => {
```
leekahung: The loadingActive state in Line 27 and 52 is no longer in use. Can be removed without affecting it.

src/components/Clients/ClientListTableRow.jsx
```diff
@@ -0,0 +1,146 @@
...
+// determine what gets rendered in the table body
+const ClientListTableRow = ({ loadingActive, labelId, client, state, dispatch }) => {
+  const theme = useTheme();
+  const { setSelectedUser } = useContext(SelectUserContext);
+  const { removeUser } = useContext(UserListContext);
+  const [pinned, setPinned] = useState(false);
+  const [selected, setSelected] = useState(false);
+
+  // determine what gets rendered in the date modified cell of table
+  const determineDateModifiedCell = () => {
```
leekahung: Think we can have this function removed since it would seem loadingActive is no longer in use in favor of having all data loaded in userList (including last active date) before passing it into PASS.

src/routes/Clients.jsx
```diff
@@ -0,0 +1,75 @@
...
+import { UserListContext } from '../contexts';
+
+/**
+ * Clients Component - Component that generates Clients section for PASS
+ *
+ * @memberof GlobalComponents
+ * @name Clients
+ */
+
+const Clients = ({ loadingActive }) => {
```
leekahung: Can remove loadingActive in Line 25 and 68 without affecting code.

leekahung (changes requested):
Alright, overall, it looks pretty good for a first iteration! Just things related to an old state that's no longer in use and some styling suggestions for future PRs.

src/components/Clients/ClientList.jsx
xscottxbrownx: Done

src/components/Clients/ClientListTable.jsx
xscottxbrownx: Done

src/components/Clients/ClientListTableRow.jsx
xscottxbrownx: Updated

src/routes/Clients.jsx
xscottxbrownx: Done

leekahung (approved):
Saw the updates for loadingActive. For my part, I think this looks okay. I'm good with approving this. Waiting reviews from others if needed.

src/components/Clients/AddClient.jsx
xscottxbrownx: Yes, future PR - I'll start adding issues now.

PR #210, opened 2023-06-03
Issue 195/set file perms styling updates

### LINKED ISSUE #195: Set File Permission Styling Updates
Adding MUI styling to "Permission to Files" (`SetAclPermissionForm.jsx`).

To begin once https://github.com/codeforpdx/PASS/pull/194 is complete.

andycwilliams:
Applying MUI to "Permission to Files" based on styling previous components.

Note that both `fullWidth`s on the `FormControl`s affect the layout.

Also note that `Select Document Type` remains unstyled, just as with previous merges. As it is an imported function that affects multiple components, it will be updated once those have had their initial styling. This is to limit the chance of errors and breaking functionality. Styling will ultimately look similar to the enter username function.

TODO: Consider whether to shorten response messages. Example: `Set permissions failed. Reason: Current user Pod cannot change container permissions to itself` is possibly overly verbose. Could potentially refactor into an alert/pop-up in a future pull request.

Closes #195

### DIFF
--- a/src/components/Form/SetAclPermissionForm.jsx
+++ b/src/components/Form/SetAclPermissionForm.jsx
@@ -2,6 +2,17 @@
 import React, { useContext } from 'react';
 // Inrupt Library Imports
 import { useSession } from '@inrupt/solid-ui-react';
+// Material UI Imports
+import Box from '@mui/material/Box';
+import Button from '@mui/material/Button';
+import FormControl from '@mui/material/FormControl';
+import FormControlLabel from '@mui/material/FormControlLabel';
+import FormLabel from '@mui/material/FormLabel';
+import InputLabel from '@mui/material/InputLabel';
+import Radio from '@mui/material/Radio';
+import RadioGroup from '@mui/material/RadioGroup';
+import TextField from '@mui/material/TextField';
+import Typography from '@mui/material/Typography';
 // Utility Imports
 import { SOLID_IDENTITY_PROVIDER, runNotification, setDocAclPermission } from '../../utils';
 // Custom Hook Imports
@@ -98,10 +109,6 @@ const SetAclPermissionForm = () => {
     }
   };
 
-  const formRowStyle = {
-    margin: '20px 0'
-  };
-
   /* eslint-disable jsx-a11y/label-has-associated-control */
   return (
     <FormSection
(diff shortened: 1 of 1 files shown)


xscottxbrownx:
## I believe lines 133-138 may be off a little...

**1)** I think the `<InputLabel>` should also be in the `<FormControl>`, but I haven't figured out how to get it to display correctly while doing so...

**2)** We want to link that label with the `<DocumentSelection>` input.
     - I think it should be `htmlFor` instead of `id` on the `<InputLabel>`, we don't want to give the label the `id` - we want to link the label to an `id` of the input.

src/components/Form/SetAclPermissionForm.jsx
```diff
@@ -110,34 +117,55 @@ const SetAclPermissionForm = () => {
...
+            <TextField
+              id="set-acl-to"
+              name="setAclTo"
+              {...username}
+              placeholder={selectedUser}
+              label="Search username"
+              required
+            />
+          </FormControl>
+          <InputLabel id="set-acl-doctype">
```
leekahung: Like what Scott mentioned, I believe should be an `htmlFor` similar to how `<label>` uses `htmlFor` and uses `id` for the input fields.

src/components/Form/SetAclPermissionForm.jsx
```diff
@@ -110,34 +117,55 @@ const SetAclPermissionForm = () => {
...
-          <label htmlFor="set-acl-perm-give">Give</label>
-          <input type="radio" id="set-acl-perm-revoke" name="setAclPerms" value="Revoke" />
-          <label htmlFor="set-acl-perm-revoke">Revoke</label>
-        </div>
-        <button disabled={state.processing} type="submit">
-          Set Permission
-        </button>
-      </form>
+          <FormControl fullWidth>
+            <FormLabel id="set-acl-perm-label">Select permission setting:</FormLabel>
```
leekahung: Similar to above, I believe this should be using `htmlFor`. The ones for `<FormControlLabel>` below should be okay since they're `id`s for the input radio checkboxes.

andycwilliams:
When I move `<InputLabel>` into the `<FormControl>`, this is the result:



As stated in the original post, this is not what `Document Selection` will ultimately look like. As it is an imported function that affects multiple components, it will be updated once those have had their initial styling. This is to limit the chance of errors and breaking functionality. As well as not breaking the layout. After it's worked on, it'll look different and more semantically appropriate.

leekahung:
> When I move `<InputLabel>` into the `<FormControl>`, this is the result:
> 
> 

Should be able to move `<InputLabel>` inside `<FormControl>` once we have DocumentSelection updated with `<Select />` and `<MenuItem />`. Think that's the main reason why it's showing that styling problem. Think we're getting close to actually performing the switch.

Had played around with it in a test branch for DocumentSelection a while back (notice the old UI and old TextField, we're not using that anymore). It should work properly afterwards.

https://github.com/codeforpdx/PASS/assets/14917816/67d9f68c-b9cf-4313-b254-835244e06243

Screen Shot 2023-06-05 at 19 17 06

Screen Shot 2023-06-05 at 19 17 26

leekahung (approved):
This looks okay. Should be able to switch out HTML elements in DocumentSelection soon. That'll allow us to fully transition over these forms to MUI at the very least and get a proper implementation with `<FormControl>` with `<InputLabel>` without avoiding it.

PR #211, opened 2023-06-04
Simplify Module Mocks in test cases

timbot1789:
I've noticed that a lot of our test files are going to be mocking the same functions from inrupt's libraries. Vitest lets us add default module mocks in a top level `__mocks__` folder. I added a file for `solid-client` with some sensible defaults.

Now most unit tests can just put `vi.mock('@inrupt/solid-client');` at the top of their file, and have their tests behave a bit more cleanly.

The defaults are a bit simple for now. We may want to replace them with more complex implementations later. It's also probably best to limit these mocks to networking functions, since the network is hard to replicate in testing.

### DIFF
--- a/__mocks__/@inrupt/solid-client.js
+++ b/__mocks__/@inrupt/solid-client.js
@@ -1,0 +1,15 @@
+import { vi } from 'vitest'
+import { mockSolidDatasetFrom, addMockResourceAclTo } from '@inrupt/solid-client';
+
+export * from '@inrupt/solid-client';
+
+const exampleUrl = 'https://example.com'
+
+export const getPodUrlAll = vi.fn(() => Promise.resolve([]));
+export const saveAclFor = vi.fn((url) => Promise.resolve(
+  {
+    ...addMockResourceAclTo(mockSolidDatasetFrom(url)),
+    internal_accessTo: exampleUrl
+  }));
+export const saveSolidDatasetAt = vi.fn((url) => Promise.resolve(addMockResourceAclTo(mockSolidDatasetFrom(url))));
+export const getSolidDataset = vi.fn((url) => Promise.resolve(addMockResourceAclTo(mockSolidDatasetFrom(url))));
--- a/test/components/NavBar/OidcLoginComponent.test.jsx
+++ b/test/components/NavBar/OidcLoginComponent.test.jsx
@@ -4,6 +4,16 @@ import React from 'react';
 import { expect, it, vi, afterEach } from 'vitest';
 import OidcLoginComponent from '../../../src/components/NavBar/OidcLoginComponent';
 
+vi.mock('@inrupt/solid-ui-react', () => ({
+  LoginButton: ({ children, oidcIssuer, redirectUrl }) => (
+    <div>
+      {oidcIssuer}
+      {redirectUrl}
+      {children}
+    </div>
+  )
+}));
+
 afterEach(() => {
   vi.clearAllMocks();
   cleanup();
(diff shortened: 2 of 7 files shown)


__mocks__/@inrupt/solid-client.js
```diff
@@ -0,0 +1,15 @@
+import { vi } from 'vitest'
+import { mockSolidDatasetFrom } from '@inrupt/solid-client';
+
+export * from '@inrupt/solid-client';
```
timbot1789, line 4: Vitest overwrites the module export with whatever we have in this file. So we want to re-export any functions we're not mocking.

__mocks__/@inrupt/solid-client.js
```diff
@@ -0,0 +1,15 @@
...
+
+const exampleUrl = 'https://example.com'
+
+export const getPodUrlAll = vi.fn(() => Promise.resolve(exampleUrl));
```
timbot1789: Add some implementations for the most common networking utilities.

test/components/NavBar/OidcLoginComponent.test.jsx
```diff
@@ -4,6 +4,16 @@ import React from 'react';
 import { expect, it, vi, afterEach } from 'vitest';
 import OidcLoginComponent from '../../../src/components/NavBar/OidcLoginComponent';
 
+vi.mock('@inrupt/solid-ui-react', () => ({
```
timbot1789, line 7: I learned that vi actually hoists these `vi.mock` calls to the top of the file, so putting them in the `describe` block doesn't do anything.

__mocks__/@inrupt/solid-client.js
```diff
@@ -0,0 +1,18 @@
...
+
+export const getPodUrlAll = vi.fn(() => Promise.resolve([]));
+export const saveAclFor = vi.fn((url) => Promise.resolve(
+  {
+    ...mockSolidDatasetFrom(url),
+    internal_accessTo: exampleUrl
+  }));
+export const saveSolidDatasetAt = vi.fn((url) => Promise.resolve(mockSolidDatasetFrom(url)));
+export const getSolidDataset = vi.fn((url) => Promise.resolve(mockSolidDatasetFrom(url)));
+export const setAgentResourceAccess = vi.fn();
```
timbot1789, line 16: These `setResourceAccess` functions aren't networked, but they require a specific format that's only generated by correctly accessing the network, and that isn't provided by `mockSolidDatasetFrom`. If that format is not provided, they error.

Mocking these here because we generally aren't interested in the output of these functions themselves; we just pass the output around. If you are interested in the output, mock the return value.

test/utils/session-helper.test.js
```diff
@@ -28,62 +22,3 @@ describe('createResourceTtlFile', () => {
     expect(Object.keys(result.predicates)).toHaveLength(7);
   });
 });
-
-describe('setDocAclForUser', () => {
```
timbot1789, line 32: These tests don't test anything in the actual codebase. Deleting them.

test/model-helpers/UserList.test.js
```diff
@@ -5,26 +5,17 @@ import {
...
-    getSolidDataset: vi.fn((url) => Promise.resolve(mockSolidDatasetFrom(url))),
-    setAgentResourceAccess: vi.fn(),
-    setAgentDefaultAccess: vi.fn()
-  };
-});
+vi.mock('@inrupt/solid-client');
 
 const makeMockUserListDataset = (userList, url) => {
-  let dataset = mockSolidDatasetFrom(url);
+  let dataset = addMockResourceAclTo(mockSolidDatasetFrom(url));
```
timbot1789, line 18: We can build valid solid datasets with this pattern.  The `setAcl` functions error if we pass them a mock dataset that didn't have `addMockResourceAclTo` called on it.

leekahung (approved):
Think this looks okay. I'll approve this.

PR #214, opened 2023-06-05
Add signed in user context

timbot1789:
3 main things in this PR:

1. Moves the `podUrl` value to a `SignedInUserContext`. That way any component can access it without needing to fetch it.
2. Removed a performance bottleneck in `fetchData` on app startup that was making things fetch slowly. Now all the fetchData requests are sent out at once, instead of one after the other
3. Refactored App.jsx so it doesn't need to pass down login state. Login is handled top level, and no other components need to worry about it.

Also added unit tests to the files.


test/App.test.jsx
```diff
@@ -0,0 +1,72 @@
...
+import { ThemeProvider } from '@mui/material/styles';
+import { useSession, SessionProvider } from '@inrupt/solid-ui-react';
+import { BrowserRouter as Router } from 'react-router-dom';
+import flushPromises from './testHelpers';
+
+import theme from '../src/theme';
+
+import App from '../src/App';
+
+const AppWithContexts = () => (
```
timbot1789: This seems to be the recommended way to test React Contexts with React Testing Library.
https://testing-library.com/docs/example-react-context/

test/App.test.jsx
```diff
@@ -0,0 +1,72 @@
...
+        <App />
+      </ThemeProvider>
+    </SessionProvider>
+  </Router>
+);
+
+vi.mock('@inrupt/solid-client');
+
+vi.mock('@inrupt/solid-ui-react', async () => {
+  const lib = await vi.importActual('@inrupt/solid-ui-react');
```
timbot1789, line 10: Need to mock `useSession` (its data is initialized by network requests), but don't want to replace the session provider or other components.

test/App.test.jsx
```diff
@@ -0,0 +1,72 @@
...
+      session: {
+        info: {
+          isLoggedIn: true,
+          webId: 'https://example.com/'
+        },
+        fetch: vi.fn()
+      }
+    });
+    const { getByLabelText } = render(<AppWithContexts />);
+    await flushPromises();
```
timbot1789: Get the `onMount` events to finish. I added this function to a test helper file.

src/routes/LandingPage.jsx
```diff
@@ -0,0 +1,40 @@
+// React Imports
+import React from 'react';
```
timbot1789: Renamed the original 'Home' component to 'Landing Page'. We have two possible 'Homes': the first place a user goes when not logged in, and the first place a user goes when they are logged in. It would be nice to have names that clearly distinguish the two.

The route to the main screen following user login was already named `home`, so this page gets renamed.

test/routes/LandingPage.test.jsx
```diff
@@ -0,0 +1,24 @@
...
+);
+
+describe('Landing Page', () => {
+  afterEach(() => {
+    cleanup();
+  });
+
+  it('renders', () => {
+    const { container } = render(<LandingPageWithContexts />);
+    expect(container).toMatchSnapshot();
```
timbot1789: Most of the size of this PR comes from this one snapshot, and that snapshot is really hard to read.  I'm still not sure of the best way to handle snapshotting with react test library. Generally I think we should only snapshot components that generate HTML directly, since that's what the user will see. We shouldn't really need to snapshot components that simply render other react components. However, since we're using MUI, very little of our codebase generates HTML, so we may want to find another method.

src/routes/LandingPage.jsx
xscottxbrownx: > The route to the main screen following user login was already named `home`, so this page gets renamed.

This route is being renamed to `Clients` in PR #209  hahah.  I still think what you did is fine.  

I also plan on renaming the `forms` route to `Documents`, and then the `inbox` and `outbox` to `Messages`

timbot1789: Haha well sounds like one of us will have a few merge conflicts to resolve when the other merges their PR first

src/layouts/Layouts.jsx
xscottxbrownx, line 1: Can you rename this file to `Layout.jsx`?

src/AppRoutes.jsx
```diff
@@ -1,28 +1,22 @@
...
-
-const AppRoutes = ({ isLoggedIn, loadingActive, loadMessages }) => {
+const AppRoutes = ({ loadingActive, loadMessages }) => {
   const restorePath = localStorage.getItem('restorePath');
   const path = restorePath ?? '/PASS/home';
 
   return (
     <Routes>
-      <Route exact path="/PASS/" element={isLoggedIn ? <Navigate to={path} replace /> : <Home />} />
-      <Route element={<ProtectedRoute isLoggedIn={isLoggedIn} />}>
```
xscottxbrownx, line 27: Able to get rid of this because the `Home` or `LandingPage` in `App.jsx` basically creates the protected route now correct?

timbot1789: Pretty much, yeah

src/AppRoutes.jsx
```diff
@@ -1,28 +1,22 @@
...
-        <Route path="/PASS/home" element={<UserSection loadingActive={loadingActive} />} />
-        <Route path="/PASS/forms" element={<Forms />} />
-        <Route path="/PASS/inbox" element={<Inbox loadMessages={loadMessages} />} />
-        <Route path="*" element={<Navigate to={restorePath} replace />} />
-      </Route>
+      <Route exact path="/PASS/" element={<Navigate to={path} replace />} />
+      <Route path="/PASS/home" element={<UserSection loadingActive={loadingActive} />} />
```
leekahung: Believe we've deprecated loadingActive in PR #169 could be removed now.

src/routes/Home.jsx
```diff
@@ -1,38 +1,72 @@
...
+  useEffect(() => {
+    fetchData();
+  }, []);
+
+  return (
+    <Layout ariaLabel="Home Page">
+      <SelectUserContext.Provider value={selectedUserObject}>
+        <UserListContextProvider>
+          <InboxMessageContext.Provider value={inboxMessageObject}>
+            <AppRoutes loadingActive={false} loadMessages={loadMessages} />
```
leekahung: Think we can eliminate loadingActive state since it's no longer being used.

src/routes/Home.jsx
```diff
@@ -1,38 +1,72 @@
...
+   * @function fetchData
+   */
+  const fetchData = async () => {
+    await Promise.all([
+      updateUserActivity(session, podUrl),
+      createPublicContainer(session, podUrl),
+      createDocumentContainer(session, podUrl),
+      createOutbox(session, podUrl),
+      createInbox(session, podUrl),
+      getInboxMessageTTL(session, inboxList).then((messages) => {
```
leekahung: getInboxMessageTTL will be updated with getMessageTTL from PR #191, might need to deal with merge conflict once it's merged.

timbot1789: updated

src/routes/Home.jsx
timbot1789: removed

src/AppRoutes.jsx
timbot1789: removed

src/AppRoutes.jsx
timbot1789: Actually, this turned out to be an issue with the PR. Refactored it to remove these route changes

test/routes/LandingPage.test.jsx
timbot1789: Refactored this so the snapshot isn't as large. This Snapshot works better now.

package.json
```diff
@@ -12,8 +12,8 @@
     "preview": "vite preview",
     "docs": "jsdoc --readme ./docs/README.md -c jsdoc.config.json",
     "docs:clear": "node clearDocs.js",
-    "lint": "eslint 'src/**/*.js' 'src/**/*.jsx' 'test/**/*.js' 'test/**/*.jsx'",
-    "lint:fix": "eslint --fix 'src/**/*.js' 'src/**/*.jsx' 'test/**/*.js' 'test/**/*.jsx'",
+    "lint": "eslint 'src/**/*.js' 'src/**/*.jsx' 'test/**/*.js' 'test/**/*.jsx' '__mocks__/**/*.js'",
```
timbot1789, line 15: A merge issue caused these mocks to start erroring. We can have eslint alert us to the issue ahead of time, and avoid some headache.

src/model-helpers/User.js
```diff
@@ -55,7 +55,7 @@ export const updateUserActivity = async (session, podUrl) => {
 
   try {
     activityDataset = await getSolidDataset(activityDocUrl, { fetch: session.fetch });
-    const activeTTLThing = getThing(activityDataset);
+    const activeTTLThing = getThing(activityDataset, `${activityDocUrl}#active`);
```
timbot1789, line 58: `getThing` always needs to be called with the URL of the desired thing. Otherwise it errors.

src/App.jsx
```diff
@@ -1,28 +1,18 @@
 // React Imports
-import React, { useEffect, useMemo, useState } from 'react';
+import React from 'react';
 // Inrupt Imports
```
leekahung: Could remove comments for unused imports Line 3 to 5

timbot1789: updated

PR #224, opened 2023-06-06
Update package-lock dependencies to fix high-security vulnerability

timbot1789:
NPM is reporting that a high security vulnerability was found in vite. Let's do a dependency update to fix that.

**Testing**
1. Pull Branch
2. Run `npm ci`
3. Check there are no vulnerabilities reported by npm during install
4. Smoke test the App

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "pass",
   "version": "0.0.4",
-  "lockfileVersion": 2,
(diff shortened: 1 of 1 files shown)


leekahung (commented):
Was the file generated using npm audit fix? I think we can do this during the version update from Development directly.

leekahung:
Seems like package-lock.json was generated before react-paginate. Update from latest from Development.

PR #235, opened 2023-06-07
Temporarily remove user activity from users in UserList

timbot1789:
I'm thinking we should remove user activity from the users displayed in users list for the time being. It's causing performance issues, and I'm not sure of the best solution to it.  This is a more complex problem than it seems. Picking the wrong direction here could have bad implications down the line. I'd like to talk about the architectural issue with the group more instead of charging ahead with something poorly considered. The feature also doesn't seem to be needed for the MVP, so I don't think there's a major issue removing it.

The issue specifically is how to create a client object on the front end when some of the information for the object is stored in the client pod, and some is stored on the case manager's pod.

However, this is a case of a more general problem: How do we build a front end data model when the data for that model is distributed across multiple pods. That's not trivial to answer

Some possibilities:
- Don't create data models with distributed data. Instead let models reflect only data stored in one (non networked) dataset. The data models store links to networked data
  -  Pros: makes data models easier to write, exposes the graph nature of the resources directly
  - Cons: Puts greater burden on the UI to organize data from different locations
- Create a model system that allows loading of data from several sources, and that stores its own loading state
  - Pros: Removes the burden of collecting data off the UI
  - Cons: Difficult to maintain, it's generally a bad idea to roll your own systems over finding a well maintained library
-  Find a javascript model library that allows for loading data from several sources
   - Pros: we don't have to write our own stuff, the community has probably written a better version anyway
   - Cons: I don't know if one exists that integrates with React. We may be able to use an ORM library like Knex and then pass it LDP or SPARQL queries, however I don't know if they work in browser environments. They are also designed to work with relational database, and we're using a graph database.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
         "@mui/x-date-pickers": "^6.5.0",
         "@zxing/browser": "^0.1.3",
         "crypto-js": "^4.1.1",
+        "dayjs": "^1.11.8",
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
         "react-paginate": "^8.2.0",
@@ -6482,6 +6483,11 @@
       "engines": {
         "node": ">=6"
       }
+    },
+    "node_modules/dayjs": {
+      "version": "1.11.8",
+      "resolved": "https://registry.npmjs.org/dayjs/-/dayjs-1.11.8.tgz",
+      "integrity": "sha512-LcgxzFoWMEPO7ggRv1Y2N31hUf2R0Vj7fuy/m+Bg1K8rr+KAs1AEy4y9jd5DXe8pbHgX+srkHNS7TH6Q6ZhYeQ=="
     },
     "node_modules/debug": {
       "version": "4.3.4",
@@ -20111,6 +20117,11 @@
       "requires": {
         "time-zone": "^1.0.0"
       }
+    },
+    "dayjs": {
+      "version": "1.11.8",
+      "resolved": "https://registry.npmjs.org/dayjs/-/dayjs-1.11.8.tgz",
+      "integrity": "sha512-LcgxzFoWMEPO7ggRv1Y2N31hUf2R0Vj7fuy/m+Bg1K8rr+KAs1AEy4y9jd5DXe8pbHgX+srkHNS7TH6Q6ZhYeQ=="
     },
     "debug": {
       "version": "4.3.4",
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "@mui/x-date-pickers": "^6.5.0",
     "@zxing/browser": "^0.1.3",
     "crypto-js": "^4.1.1",
+    "dayjs": "^1.11.8",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-paginate": "^8.2.0",
--- a/src/components/Clients/AddClient.jsx
+++ b/src/components/Clients/AddClient.jsx
@@ -137,17 +137,17 @@ const AddClient = () => {
       >
         <div>
           <label htmlFor="add-user-given-name">First/given name: </label>
-          <input id="add-user-given-name" name="addUserGivenName" {...userGivenName} />{' '}
+          <input id="add-user-given-name" name="addUserGivenName" {...userGivenName} />
         </div>
         <br />
         <div>
           <label htmlFor="add-user-last-name">Last/family name: </label>
-          <input id="add-user-last-name" name="addUserFamilyName" {...userFamilyName} />{' '}
+          <input id="add-user-last-name" name="addUserFamilyName" {...userFamilyName} />
         </div>
         <br />
         <div>
           <label htmlFor="add-username">
-            Add username to client list (i.e., username without {currentOidcIssuer.split('/')[2]}):{' '}
+            Add username to client list (i.e., username without {currentOidcIssuer.split('/')[2]}):
           </label>
           <br />
           <br />
@@ -157,7 +157,7 @@ const AddClient = () => {
             size="25"
             value={username}
             onChange={(e) => wrappedSetUsername(e.target.value)}
-          />{' '}
+          />
         </div>
         <br />
         <label htmlFor="add-webId">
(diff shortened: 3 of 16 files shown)


leekahung:
The idea for Last Active Date was intended to check up on client activity, like whether or not they've interacted with the application (that way a case worker could decide to prioritize/manage cases based on that date), and whether or not a user is a PASS user since active.ttl is PASS-specific and is only generated/updated when logging into pass.

Think we can phase this out or correct this once a solution is found.

For prioritization, I could see it being replaced with the pin toggle.

For checking to see if someone's a PASS user, I'm thinking maybe run an await getSolidDataset call for someone PASS specific and if it fails, we know they're not a PASS user. (An indicator could be shown in the ClientList with something like "verified" tag based on that result)

For checking up on user activity, think we need something flushed out.

xscottxbrownx:
What's the status of this?

What performance issues were you seeing?

timbot1789:
> 

@leekahung  I do like the idea of an activity tracker. You're right it could be useful to an end user. However, I think there's a lot of info in addition to the user activity that we'd want to expose to the case manager. Maybe we can set up a user profile model, which fetches a document stored on the user's pod, and that's the information we display in the table. The user activity can be one field in that profile.

I liked Gabby's form for client creation a lot that she had in figma:
https://www.figma.com/file/BfsGjGbFUVvtqNq18sykoY/Pass-Project?type=design&node-id=1216-2986

She's suggesting to collect a lot of data on the client there (household size, head of household, etc.). All of that information is pretty personal and should probably be stored on the client's pod. When the case manager signs in, we could then fetch that data from the pod for each client and display that.

However, I don't think the current architecture I have setup for user list facilitates fetching individual client profiles like that well.

timbot1789:
> What's the status of this?
> 
> What performance issues were you seeing?

Still want to talk with the group about how best to set up the architecture for the app.

Performance issue was loading the clients table when a case manager had several clients. If a case manager has 15 clients, we'd send out 15 fetch requests and wait for all of them to return before displaying the table. That system doesn't scale. We can make a specific solution for just the user's table, but the nature of the solid protocol means we'll almost certainly encounter issues like it again. We don't have a single table or database we can fetch all the application data from. Data can be almost anywhere on the web.

xscottxbrownx:
> > 
> 
> @leekahung I do like the idea of an activity tracker. You're right it could be useful to an end user. However, I think there's a lot of info in addition to the user activity that we'd want to expose to the case manager. Maybe we can set up a user profile model, which fetches a document stored on the user's pod, and that's the information we display in the table. The user activity can be one field in that profile.
> 
> I liked Gabby's form for client creation a lot that she had in figma: https://www.figma.com/file/BfsGjGbFUVvtqNq18sykoY/Pass-Project?type=design&node-id=1216-2986
> 
> She's suggesting to collect a lot of data on the client there (household size, head of household, etc.). All of that information is pretty personal and should probably be stored on the client's pod. When the case manager signs in, we could then fetch that data from the pod for each client and display that.
> 
> However, I don't think the current architecture I have setup for user list facilitates fetching individual client profiles like that well.

i like this too...

Originally I had planned to replace the webID link column with a "Notes" column for some sort of quick note about client, and make the webID link just be on the client name (it would show the client name but be a link to that.)  Of course, the "Notes" column would be rather small and I think a larger spot for it would be needed. I kind of envisioned this as if you clicked on a client a larger view would appear with a lot more info/options.

I'll have to relook at the figma file, it's been a little bit for me.

timbot1789:
> Originally I had planned to replace the webID link column with a "Notes" column

Yeah I think that would be a good idea. I agree `webID` isn't something we really need to expose to the case manager. They care about the person, not the person's ID number. It's useful in these early stages to show how precisely the system is working (and make sure the system is working properly), but in Prod we'd probably want to display more interesting information.

The idea of adding "notes" is an interesting backend problem though, since it means we'd have two data sources for the client profile: the info the client has given us (stored on the client's pod), and the notes the case manager has written about them (stored on the case manager's pod). Collecting both these sources into one document to give to the front end is a fun problem.

xscottxbrownx:
> > Originally I had planned to replace the webID link column with a "Notes" column
> 
> Yeah I think that would be a good idea. I agree `webID` isn't something we really need to expose to the case manager. They care about the person, not the person's ID number. It's useful in these early stages to show how precisely the system is working (and make sure the system is working properly), but in Prod we'd probably want to display more interesting information.
> 
> The idea of adding "notes" is an interesting backend problem though, since it means we'd have two data sources for the client profile: the info the client has given us (stored on the client's pod), and the notes the case manager has written about them (stored on the case manager's pod). Collecting both these sources into one document to give to the front end is a fun problem.

Another option instead of free-form notes would be "categories."   Like some dropdown options of the category of client, or even status of client.

This is just a thought at the moment, but categories could be like: Housing, Food, Physical Healthcare, Mental Healthcare...
Status could be like: Waiting on client, waiting on organization, submitted docs, receiving care...

leekahung:
> Still want to talk with the group about how best to set up the architecture for the app.
> 
> Performance issue was loading the clients table when a case manager had several clients. If a case manager has 15 clients, we'd send out 15 fetch requests and wait for all of them to return before displaying the table. That system doesn't scale. We can make a specific solution for just the user's table, but the nature of the solid protocol means we'll almost certainly encounter issues like it again. We don't have a single table or database we can fetch all the application data from. Data can be almost anywhere on the web.

Yeah, this problem of having to fetch all relevant ttl files is plaguing inbox/outbox as well. I do remember discussions of SparQL and MongoDB. We could potentially go from there for solutions.

xscottxbrownx:
> For prioritization, I could see it being replaced with the pin toggle.

I had another thought about "priority".  My thought with the pinning (which we don't have to implement it this way) was to just make those that are pinned stick to the top of the list.   I do see that as a kind of priority, but really it could be more of a "favorite" or "most used/interacted with" honestly.

---

I've been rethinking priority and I kind of like from the Figma file, the idea of the star icon.   BUT I think it would be awesome to have 3 priority levels:

Priority 3 - just normal, no priority really. Outlined star icon in black.
Priority 2 - more priority. Half filled in star, gold color fill and outline.
Priority 1 - top priority. Fully filled in star, gold color.

I would also prefer the number be inside the star (3, 2, and 1) probably.   

---

This kind of came from Tim's idea of having making the "priority" filed on the .ttl file a number instead of just a boolean. 

---

If we did something like this, maybe there isn't a pin icon (priority star takes it's place.)  AND the clientlist gets sorted before rendering by priority, then maybe last name within that priority?

PR #241, opened 2023-06-10
[Feature/Hotfix] Use PASS container and use oidcIssuer for webId generation for usernames

### LINKED ISSUE #240: Organizing PASS-related containers
Currently, we naming our containers with the name it's intended to be used for. However, we won't know if another Solid Provider will be using the same container name for something else. In order to avoid clashes, we could rename containers to use PASS as a prefix similar to what PASS_Credentials are doing.

Another approach is to just place them all inside a dedicated container called PASS.

Two possible solutions to this as described above:

1. Add prefixes to existing containers and update relevant functions related to that
2. Place them all within a dedicated container called PASS

leekahung:
Feature updates in this PR include updates to the following containers to use a PASS container:

- Users => PASS/Users
- Documents => PASS/Documents
- inbox => PASS/Inbox
- outbox => PASS/Outbox
- public => PASS/Public
- PASS_Credentials => PASS/Credentials

That way, it'll prevent potential clashes with other Solid Providers that might serve containers with the same name but might have different ACL settings. Unit Tests related to this change have all been updated.

The location of where uploadDocument for self-uploads and cross-pod-uploads are now the same.

- Bank Statement => PASS/Documents/Bank_Statement
- Passport => PASS/Documents/Bank_Statement
- Drivers License => PASS/Documents/Drivers_License

Since they're all located at the same place, functionality originally made for setting ACL permissions to document containers outside of Documents could now be used within Documents.

Additionally, this PR include updates to functions that utilize usernames, which now use a function called getPodUrl to generate webIds. The getPodUrl function pieces the podUrl together using oidcIssuer and the username as their starting point. It's a general-purpose function that takes in any usernames.

The reason why some functions are breaking at the moment is because they're still using SOLID_IDENTITY_PROVIDER, but they should really be using the oidcIssuer. Using the former only works if you intend to just use the default provider. Otherwise, it'll call the wrong podUrl for users for those functions utilizing username. So this PR is meant to be a hotfix for that as well.

In the future, we might want to switch this over to be just webIds too, but for now, this should suffice.

Am considering of merging this into Master as `v0.0.5.1` after merging into Development because of the breaking changes for using alternative providers.


leekahung:
Value expected in snapshot depends on development or production build. So even if the tests pass locally, it'll fail on GitHub.

I'm not sure how to fix that.

leekahung:
Need to fix some of the unit tests

leekahung:
Fixed all unit tests except for the snapshot with OidcLogin

timbot1789 (commented):
For that last failing test, you just need to update the snapshot. Run `npm run test`, and then when tests finish running, press `u`. Then commit the updated snapshot and run the test again.

src/utils/network/session-core.js
```diff
@@ -91,14 +90,13 @@ export const setDocContainerAclPermission = async (session, permissions, otherPo
   const containerUrl = getContainerUrl(session, 'Documents', INTERACTION_TYPES.SELF);
   const urlsToSet = [
     containerUrl,
-    `${containerUrl}Bank%20Statement/`,
-    `${containerUrl}Passport/`,
-    `${containerUrl}Drivers%20License/`
+    `${containerUrl}PASS_Bank_Statement/`,
```
timbot1789: Wouldn't moving these all into a `PASS` container have the same effect?
e.g. `kahung.solidcommunity.net/pod/PASS/Bank_Statement` , `kahung.solidcommunity.net/pod/PASS/Passport`

leekahung:
> For that last failing test, you just need to update the snapshot. Run `npm run test`, and then when tests finish running, press `u`. Then commit the updated snapshot and run the test again.

Yeah, it passed on my local machine. I don't think I could do that on GitHub though for the CI.

Screen Shot 2023-06-10 at 9 48 06

Screen Shot 2023-06-10 at 9 48 55

src/utils/network/session-core.js
leekahung: It would actually. I wasn't sure if that's something we wanted at the moment, so I've just renamed them for now with just prefixes.

leekahung: Should be an easy swap anyways, might as well try that for this update. 👍 

Think I'll do the same with PASS_Credentials.

leekahung: Updated containers to be stored within a PASS container

leekahung:
Alright, it's probably not the proper way to do this, but created a workaround for the test to use a mock SOLID_IDENTITY_PROVIDER.

It'll just assume that the mock SOLID_IDENTITY_PROVIDER the correct one from import.meta.env in the rendering of OidcLoginComponent.

src/constants/solid_provider.js
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * The URL to a specific Solid Provider after login
+ *
+ * @name currentOidcIssuer
+ * @type {URL}
+ * @memberof utils
+ */
+
+export const currentOidcIssuer = localStorage.getItem('oidcIssuer') ?? SOLID_IDENTITY_PROVIDER;
```
timbot1789, line 26: This is not a constant value, so should probably be somewhere else, but it'll do for a hotfix.

timbot1789 (approved):

PR #243, opened 2023-06-10
Update selectedUserContext to hold user object rather than username

timbot1789:
The user object is more informative than the username, so it is of more use to components than just the username.

### DIFF
--- a/src/components/Clients/ClientListTableRow.jsx
+++ b/src/components/Clients/ClientListTableRow.jsx
@@ -42,6 +42,10 @@ const ClientListTableRow = ({ labelId, client, state, dispatch }) => {
   // Event handler for selecting client from client list
   const handleSelectClient = async (clientToSelect) => {
     runNotification(`Client "${clientToSelect.person}" selected.`, 3, state, dispatch);
+    if (clientToSelect.webId === selectedUser.webId) {
+      setSelectedUser({});
+      return;
+    }
     setSelectedUser(clientToSelect);
   };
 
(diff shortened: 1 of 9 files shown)


src/components/Form/CheckAclPermsDocContainerForm.jsx
```diff
@@ -111,7 +110,7 @@ const CheckAclPermsDocContainerForm = () => {
               id="set-acl-to"
               name="setAclTo"
               {...user}
```
leekahung: Since moving over to MUI, the selectedUser seems to only seems to set the placeholder, it doesn't set the value for the `<TextField>` like we had with `<input>`.

https://github.com/codeforpdx/PASS/assets/14917816/24684eda-b921-4ab4-a5db-e6c6247eed90

Think we could start replacing the `useField` hooks and use:

```
const [username, setUsername] = useState("")
```

And where we could set `value={selectedUser.username ?? username}` with `onChange={(e) => setUsername(e.target.value)}` inside `<TextField>`

Should be similar to what we're doing in `<OidcLoginComponent>`.

timbot1789: Updated. Instead of `selectedUser.username ?? username`, I just default the username state to `selectedUser.username`

src/components/Clients/ClientListTableRow.jsx
leekahung, line 1: While I could check different boxes to select a user, I can't seem to uncheck the checkbox anymore after selecting a user from the list. So currently it's behaving more like a radio box than a checkbox. This is likely due to the useState for selected being removed from the onChange in Checkbox.

https://github.com/codeforpdx/PASS/assets/14917816/c7eae3d7-0ee2-4d1f-ac2c-fb8adc23a817

leekahung (commented):
Other than that checkbox issue, I think everything else looks great! 👍

src/components/Clients/ClientListTableRow.jsx
timbot1789: Good catch! Fixed

leekahung (approved):
Alright, I think functionality is working as intended and the code looks good. I approve.

PR #245, opened 2023-06-11
Refactoring StatusNotification/StatusMessage to truncate long file names

### LINKED ISSUE #244: Create function to truncate long file names
At the moment, file names being uploaded are displayed in full in PASS' UI. Because of it, it'll stretch out the containers until it reaches the length of the string.

While extremely long file names could be uncommon, we could nevertheless create a function to truncate them when being displayed. The file names will still be stored in their entirety, it'll just be truncated when shown on PASS.

**Alternative solution**
An alternative to this would be to use styling instead for truncation.

leekahung:
Switching from the original approach with using a dedicated function to truncate file names, this PR now refactors StatusNotification/StatusMessage in order to handle potentially long file names. FormSubmissionHelper.js has been updated as well for this change, but mainly for altering the message being sent to StatusMessage for file upload/updates.

https://github.com/codeforpdx/PASS/assets/14917816/09001dd1-2b04-4be8-a0a3-5eb6cef74bb0

### DIFF
--- a/src/components/Form/FormSection.jsx
+++ b/src/components/Form/FormSection.jsx
@@ -48,12 +48,7 @@ const FormSection = ({ title, state, statusType, defaultMessage, children }) => 
           {title}
         </Typography>
         {children}
-        <StatusNotification
-          notification={state.message}
-          statusType={statusType}
-          defaultMessage={defaultMessage}
-          locationUrl={state.documentUrl}
-        />
+        <StatusNotification state={state} statusType={statusType} defaultMessage={defaultMessage} />
       </Paper>
     </Box>
   </Container>
--- a/src/components/Form/UploadDocumentForm.jsx
+++ b/src/components/Form/UploadDocumentForm.jsx
@@ -59,6 +59,9 @@ const UploadDocumentForm = () => {
   );
 
   const handleDocumentUpload = async (event) => {
+    event.preventDefault();
+    dispatch({ type: 'SET_PROCESSING' });
+
     if (!docType) {
       runNotification('Search failed. Reason: No document type selected.', 5, state, dispatch);
       setTimeout(() => {
--- a/src/components/Notification/StatusMessage.jsx
+++ b/src/components/Notification/StatusMessage.jsx
@@ -1,6 +1,7 @@
 // React Imports
 import React from 'react';
 // Material UI Imports
+import Box from '@mui/material/Box';
 // import CircularProgress from '@mui/material/CircularProgress';
 import Typography from '@mui/material/Typography';
 
@@ -21,7 +22,7 @@ import Typography from '@mui/material/Typography';
  * notification and locationUrl, which is optional (see {@link statusMessageProps})
  */
 
-const StatusMessage = ({ notification, locationUrl }) => {
+const StatusMessage = ({ notification, locationUrl, filename }) => {
   if (locationUrl) {
     return (
       <Typography sx={{ fontWeight: 'bold' }}>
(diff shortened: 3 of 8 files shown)


timbot1789 (commented):
I see the problem, but I feel like this would be better solved by an overflow property on an HTML element. That way the user would still be able to copy-paste the full file name, and we'd be able to reactively truncate the name based off the available screen space. We also wouldn't risk an inattentive programmer accidentally using the truncated value instead of the real value.

Kinda like this using CSS:
```
p {
  white-space: nowrap;
  overflow: hidden;
  text-overflow: ellipsis;
  max-width: 200px;
}
```

leekahung:
> I see the problem, but I feel like this would be better solved by an overflow property on an HTML element. That way the user would still be able to copy-paste the full file name, and we'd be able to reactively truncate the name based off the available screen space. We also wouldn't risk an inattentive programmer accidentally using the truncated value instead of the real value.
> 
> Kinda like this using CSS:
> 
> ```
> p {
>   white-space: nowrap;
>   overflow: hidden;
>   text-overflow: ellipsis;
>   max-width: 200px;
> }
> ```

Probably an indication we need StatusNotification/StatusMessage refactored, but file name is currently being called together under a single string along with the message. So if we truncate that string, it'll cut off the message instead of just the file name.

We could parse file name out and place it at the end, but we'll probably have to rephrase and restructure the StatusMessages.

To avoid the kind of problem you mentioned, how about explicitly calling the function only in places where truncation is needed as oppose to being called in one central location. An alternative is to just rename the variable for fileName to processedFileName to indicate it have been "processed with a function".

timbot1789:
> Probably an indication we need StatusNotification/StatusMessage refactored.

That seems like a good inferrence to me. So why not fix it there? If the issue is "StatusNotification doesn't display very long file names well" then it seems to me the issue is with StatusNotification, not the file name. 

> To avoid the kind of problem you mentioned, how about explicitly calling the function only in places where truncation is needed as oppose to being called in one central location.

That sounds reasonable. How would you enforce that? In OOP terms that sounds a lot private member of a class or component. Maybe add it as an internal function to StatusNotification? Or allow StatusNotification to take in a data transform function?

In general, it seems to me that this fix is conflating a _display_ issue with a _data_ issue. The data (filename) is correct. No need to edit it. The issue is the display (html and css) does not show the data to the user well. So we should try editing the display first, before trying to edit the data.

leekahung:
> > Probably an indication we need StatusNotification/StatusMessage refactored.
> 
> That seems like a good inferrence to me. So why not fix it there? If the issue is "StatusNotification doesn't display very long file names well" then it seems to me the issue is with StatusNotification, not the file name.
> 
> > To avoid the kind of problem you mentioned, how about explicitly calling the function only in places where truncation is needed as oppose to being called in one central location.
> 
> That sounds reasonable. How would you enforce that? In OOP terms that sounds a lot private member of a class or component. Maybe add it as an internal function to StatusNotification? Or allow StatusNotification to take in a data transform function?
> 
> In general, it seems to me that this fix is conflating a _display_ issue with a _data_ issue. The data (filename) is correct. No need to edit it. The issue is the display (html and css) does not show the data to the user well. So we should try editing the display first, before trying to edit the data.

Think I'll get a crack at that. Let's see what I can do.

leekahung:
Have altered approach to truncating file name using styling instead of dedicated function. As a result StatusMessage/StatusNotification has been updated.

timbot1789 (approved):
Nice!

xscottxbrownx (approved):
looks good to merge

PR #246, opened 2023-06-11
Include ability to unselect users from Clients and Forms without submitting forms

### LINKED ISSUE #159: Create button to clear selected user
At the moment, the SelectUser context could be triggered by clicking "select" within the Users List component. However, how it works now is that it gets cleared after every form submission related to cross-pod interactions.

There could be cases where the user want to do multiple cross-pod interactions. Having the selected user state cleared after form submission (while fine in the case of single cross-pod interactions) defeats the point of a select user feature.

We should create a button somewhere that lets users to explicitly clear selected user rather than having it auto-reset on form submissions.

leekahung:
Before this PR, the only way to clear selected users was through Forms in PASS.

However, the implementation in Forms was not ideal. If a case worker/manager were to select a client from the Clients List in order to fill out their Forms, it'll be better if all the forms remains populated until the case worker are done with them. How it's implemented now is that form submission clears selected user.

This is a bad workflow since the case worker would have to go back to forms to select the user again in case they don't remember the user's username. In fact, we might want to move on from that as well in the future with webIds and podUrls remain strictly under the hood.

So in this PR, it builds on top of what's already done in Clients and included a conditional to "unselect" users from Clients via the checkbox and adds a new button in Forms to clear selectedUser with instead of using form submissions to clear selectedUser.

https://github.com/codeforpdx/PASS/assets/14917816/2eba6485-30be-4bb0-89fd-5033893d6903

On a relevant note, selectedUser is being updated in PR #243 to use the more data-rich userListObject, so we could potentially use that to update features made here. But for now, this PR deals primarily with the selection and deselection of selectedUser on PASS.

### DIFF
--- a/src/components/Clients/ClientListTableRow.jsx
+++ b/src/components/Clients/ClientListTableRow.jsx
@@ -27,10 +27,9 @@ import { SelectUserContext, UserListContext } from '../../contexts';
 // determine what gets rendered in the table body
 const ClientListTableRow = ({ labelId, client, state, dispatch }) => {
   const theme = useTheme();
-  const { setSelectedUser } = useContext(SelectUserContext);
+  const { selectedUser, setSelectedUser } = useContext(SelectUserContext);
   const { removeUser } = useContext(UserListContext);
   const [pinned, setPinned] = useState(false);
-  const [selected, setSelected] = useState(false);
 
   // determine what gets rendered in the date modified cell of table
   const modifiedDate = client.dateModified
@@ -41,14 +40,13 @@ const ClientListTableRow = ({ labelId, client, state, dispatch }) => {
   const pinnedIcon = pinned ? <PushPinIcon color="secondary" /> : <PushPinOutlinedIcon />;
 
   // Event handler for selecting client from client list
-  const handleSelectClient = async (clientToSelect, selectedClientUrl) => {
-    if (!selected) {
-      runNotification(`Client "${clientToSelect}" selected.`, 3, state, dispatch);
-      setSelectedUser(selectedClientUrl.split('/')[2].split('.')[0]);
-    } else {
-      runNotification(`Client "${clientToSelect}" unselected.`, 3, state, dispatch);
-      setSelectedUser('');
+  const handleSelectClient = async (clientToSelect) => {
+    runNotification(`Client "${clientToSelect.person}" selected.`, 3, state, dispatch);
+    if (clientToSelect.webId === selectedUser.webId) {
+      setSelectedUser({});
+      return;
     }
+    setSelectedUser(clientToSelect);
   };
 
   // Event handler for deleting client from client list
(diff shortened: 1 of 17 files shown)


src/routes/Forms.jsx
```diff
@@ -25,8 +29,29 @@ const Forms = () => {
 
   localStorage.setItem('restorePath', location.pathname);
 
+  const { selectedUser, setSelectedUser } = useContext(SelectUserContext);
+
   return (
-    <>
+    <div
```
xscottxbrownx: Change `div` to `Box` and change `style` object to a `sx` object.

Just less MUI refactoring later.

xscottxbrownx (changes requested):
1 change requested for MUI.

---

Honestly, I don't really like the way the MUI inputs handle this. I would prefer if the name was already input, but it seems it's more of a placeholder.

src/routes/Forms.jsx
leekahung: Got `div` changed into `Box`

leekahung:
> Honestly, I don't really like the way the MUI inputs handle this. I would prefer if the name was already input, but it seems it's more of a placeholder.

Think we'll have a more concrete solution to this in PR #243.

src/components/Clients/ClientListTableRow.jsx
```diff
@@ -43,6 +43,7 @@ const ClientListTableRow = ({ labelId, client, state, dispatch }) => {
   const handleSelectClient = async (clientToSelect) => {
     runNotification(`Client "${clientToSelect.person}" selected.`, 3, state, dispatch);
```
xscottxbrownx: does it always run line 44? Even if they are already selected?  

If so, it shouldn't. This should be an if/else no?

leekahung:
@xscottxbrownx Ready for another review?

leekahung:
> 1 change requested for MUI.
> 
> Honestly, I don't really like the way the MUI inputs handle this. I would prefer if the name was already input, but it seems it's more of a placeholder.

This got fixed after PR #243 merged into Development so it's now incorporated into this PR after updating the branch with Development. 👍

xscottxbrownx (commented):
one question/comment on `ClientListTableRow`

src/components/Clients/ClientListTableRow.jsx
leekahung: Not really sure what kind of JS voodoo happening or the function calls are just really fast, it seems to run fine.

https://github.com/codeforpdx/PASS/assets/14917816/7992ad1c-c287-4377-9342-c09e786780e5

But you're right. Line 44 always seem to run regardless of condition. This would probably break if something is slowing down the function calls. I'm moving this to after the if statement.

xscottxbrownx (approved):
with the new change in ClientListTableRow - I approve.

PR #249, opened 2023-06-12
Issue 135/allow upload, fetch, update, and delete of multiple documents

timbot1789:
This is a rebuild of our documents and uploads feature. Documents are now handled by a context much like the user list. The pod storage has changed has well. Document information are now stored in a single Turtle File, `doclist.ttl`, and there are no longer separate containers for each file type. Everything is stored directly under the Documents container. This change has come with a few new features: 

Previously, you were only able to upload one document of a type. Now, you can upload as many documents as you like (although you can only upload one document of a given name). We also automatically display the documents in a user's pod in a table, instead of requiring you to fetch them. That table also contains the delete action. It is no longer a separate form.

[Screencast from 06-19-2023 12:11:52 PM.webm](https://github.com/codeforpdx/PASS/assets/37914436/207baf96-45fe-4da7-8279-56a2ff03a04a)

Client Pod selection has also been updated. Now, users always have exactly one pod available to interact with in the Forms page. They are interacting with either a client pod, or their own pod. Currently, they change this pod by selecting a different user in the client page. If they do not have any user selected, their own pod is selected.

[Screencast from 06-19-2023 12:13:27 PM.webm](https://github.com/codeforpdx/PASS/assets/37914436/a1eb1c92-0d77-477c-8982-69373d1e02ad)

Because a user is no longer able to interact with multiple pods at a time, the Cross-pod upload forms have been removed. Now, they can use the standard upload form to upload to whatever pod they have selected.
 
[Screencast from 06-19-2023 12:13:27 PM.webm](https://github.com/codeforpdx/PASS/assets/37914436/ce35b2bc-6e70-4e77-899a-814b2fc7dcf1)

NOTE: This disables permission setting on the documents container. You will need to set permissions manually for now.

This also breaks the signature system. This will be reimplemented here: https://github.com/codeforpdx/PASS/issues/259


test/helpers/test-globals.js
```diff
@@ -0,0 +1,5 @@
+/* eslint-disable */
+export const setup = () => {
+  process.env.TZ = 'UTC';
```
timbot1789, line 3: Set the timezone of tests to `UTC` instead of local time.

xscottxbrownx:
@timbot1789 is this ready to be looked at now? Or still working on it?

Should we work on the fix for the issues this one causes, at the same time? The search and updating.

timbot1789:
> @timbot1789 is this ready to be looked at now? Or still working on it?
> 
> Should we work on the fix for the issues this one causes, at the same time? The search and updating.

Yeah it's ready for review. I want to get feedback on the API change and structure. Then I can put up another PR to move document fetching over to this.

Since this is a breaking API change, merging this as-is will break features in Development.

src/components/Form/DeleteDocumentForm.jsx
```diff
@@ -75,6 +72,14 @@ const DeleteDocumentForm = () => {
               docType={docType}
             />
             <br />
+            <TextField
```
leekahung: If the user have access to delete files in that container, they could see what files are already in that container. As such, it'll be better for users to select the file to delete rather than typing out the file name.

src/model-helpers/Document.js
```diff
@@ -0,0 +1,193 @@
...
+  } catch {
+    const docThing = await docDescToThing(fileDescription, docUrl, file);
+    const doc = await saveDescription(docThing, createSolidDataset(), session, docUrl);
+    await saveFileInContainer(docUrl, file, {
+      fetch: session.fetch,
+      slugSuggestion: file.name
+    });
+    return parseDocument(getThingAll(doc)[0]);
+  }
+  throw Error('A file of this name already exists');
```
leekahung: I believe createDocument is the replacement for uploadDocument. If that's the case, think we should have the ability to update documents as well like what we had with updateDocument. There would be cases where people wish to store an updated file with the same name.

src/model-helpers/Document.js
```diff
@@ -0,0 +1,193 @@
...
+  const name = getStringNoLocale(documentThing, RDF_PREDICATES.name);
+  const type = getStringNoLocale(documentThing, RDF_PREDICATES.identifier);
+  const endDate = getDate(documentThing, RDF_PREDICATES.endDate);
+  const checksum = getStringNoLocale(documentThing, RDF_PREDICATES.sha256);
+  const description = getStringNoLocale(documentThing, RDF_PREDICATES.description);
+  const fileUrl = getUrl(documentThing, RDF_PREDICATES.URL);
+  return { uploadDate, name, type, endDate, checksum, description, fileUrl };
+};
+
+const deleteRecursively = async (dataset, options) => {
```
leekahung: Should move this function right above deleteDocument. Like what's done with createFileChecksum, addAdditionalInfo and docDescToThing

src/components/Form/CrossPodWriteForm.jsx
```diff
@@ -20,69 +19,50 @@ import FormSection from './FormSection';
...
 
-    if (!docType) {
-      runNotification('Search failed. Reason: No document type selected.', 5, state, dispatch);
-      setTimeout(() => {
-        dispatch({ type: 'CLEAR_PROCESSING' });
-      }, 3000);
-      return;
+    try {
+      await createDocument(file, fileDesc, session, `${podUrl}PASS/`);
+      runNotification(`File "${file.name}" updated on Solid.`, 5, state, dispatch);
```
leekahung: Think this would be an upload message

src/components/Form/CrossPodWriteForm.jsx
```diff
@@ -20,69 +19,50 @@ import FormSection from './FormSection';
...
+    } catch (error) {
+      runNotification(`File failed to upload. Reason: ${error.message}`, 5, state, dispatch);
```
leekahung: This this should trigger a second function if it's updating a file and it's not an upload error.

timbot1789: This disabled file update. I don't think we actually have a use or UI for that in the system. It definitely shouldn't be in the same flow as upload.

src/components/Form/CrossPodWriteForm.jsx
```diff
@@ -97,19 +77,11 @@ const CrossPodWriteForm = () => {
       statusType="Upload status"
       defaultMessage="To be uploaded..."
     >
-      <form onSubmit={handleCrossPodUpload} autoComplete="off">
+      <form onSubmit={handleFormSubmission} autoComplete="off">
         <div style={formRowStyle}>
-          <label htmlFor="cross-upload-doc">Upload document to username: </label>
+          <p>Upload document to {person}: </p>
```
leekahung: Should have a fallback for the value of person if person from the SelectUserContext.

src/components/Form/CrossPodWriteForm.jsx
timbot1789: If you want to upload a new document, just upload a new document. If you want to replace a document, delete the old document, then upload a new one. If you want to update an existing doc's description, well we don't have a UI for that anyway right now.

src/components/Form/CrossPodWriteForm.jsx
leekahung: But then again, think it might be better if this remains an input field for username like before. We could have a line which shows the person's name above the input field if a person has been selected

src/components/Form/CrossPodWriteForm.jsx
```diff
@@ -20,69 +19,50 @@ import FormSection from './FormSection';
...
+    runNotification(`Uploading "${file.name}" to Solid...`, 3, state, dispatch);
 
-    if (!docType) {
-      runNotification('Search failed. Reason: No document type selected.', 5, state, dispatch);
-      setTimeout(() => {
-        dispatch({ type: 'CLEAR_PROCESSING' });
-      }, 3000);
-      return;
+    try {
+      await createDocument(file, fileDesc, session, `${podUrl}PASS/`);
```
leekahung: Function seems to bug out when trying to upload files.

Screen Shot 2023-06-14 at 14 35 23 (2)

It looks like it's trying to build a new sub container within Bank Statement, but couldn't (see Pod after trying to upload). So it couldn't really place the file in it.

Screen Shot 2023-06-14 at 14 32 52 (2)

src/components/Form/CrossPodWriteForm.jsx
leekahung: We had, that was part of updateDocument. If you were to change the description, it'll update the description in the document.ttl. The same goes for expire date.

src/utils/network/session-core.js
```diff
@@ -285,89 +144,6 @@ export const checkContainerPermission = async (session, otherPodUsername) => {
...
- * Function that generates the Documents container for users
- *
- * @memberof utils
- * @param {Session} session - Solid's Session Object
- * @param {URL} podUrl - The user's Pod URL
- * @returns {Promise} Promise - Creates Documents container for storage of
- * documents being uploaded by authorized users
- */
-
-export const createDocumentContainer = async (session, podUrl) => {
```
leekahung, line 326: Is there an equivalent function for createDocumentContainer? In case of a container being deleted, it'll no longer be regenerated without this.

src/components/Form/CrossPodWriteForm.jsx
timbot1789: Is that how the update flow is supposed to work? That's not how I found it working in testing. Here's what I tried doing in the current version (attempted action -> result): 

1. Upload a new file by selecting a type (passport) and a file (passport.pdf) to upload -> File and document.ttl are added to a pod
2. Try to update the file description by selecting the type passport and entering a new description (intending to target passport.pdf) -> cannot upload. upload button is disabled because there is no file added
3. Try to update a file description by selecting the type passport, entering a new description, and uploading passport.pdf again -> Prompted that the file already exists do I want to update? - > I click yes -> Description in document.ttl is updated, but it is unclear what uploading passport.pdf did. The system either deleted the original passport.pdf and replaced it with a new upload, or it ignored the new passport.pdf I added, and kept the original file the same.
4. Try to upload a new passport document `myNewPassport.pdf` with a description-> Get prompted that myNewPassport doesn't exist (I know?) do I want to upload? -> Click yes -> `myNewPassport.pdf` gets added to the pod. `document.ttl` gets updated with a new description, however the url field still points to `passport.pdf`, not `myNewPassport.pdf`. No new document.ttl is created for `myNewPassport.pdf`. The result is a document.ttl pointing to `passport.pdf` with a description intended for `myNewPassport.pdf`.

I'm not really sure what the intended flow is here. 

And besides, updating these documents (which are all actually collections of multiple documents and files) isn't a trivial problem. We can't update the files themselves (only delete and replace them), we can only update the descriptions we generate and store in document.ttl. However, a direct update of document.ttl not currently possible in the system (or if it is, it's too unintuitive to be useful). Also any updates to the documents invalidate signatures (at least how they're currently implemented). That should be taken into account when updating anything.

src/components/Form/CrossPodWriteForm.jsx
timbot1789: That's an authentication error. That will happen when you don't have authorization to write to a container. Does your user have access to the pod it's trying to write to?

We may need to work on case manager permissions in another PR, but an upload failing when a user doesn't have permission to access the pod is expected behavior.

leekahung: This is a case where the person is trying to upload the document on their own pod, which is why I found it a bit puzzling.

src/components/Form/CrossPodWriteForm.jsx
leekahung: How it's implemented now, especially after the initial upload, is, unfortunately, poorly-written on my part. However, I'll do my best to detail what was intended before being moved to `/PASS/Documents/`. (see below)

---

If a user wishes to upload `document.pdf` to "Bank Statement", the user can simply:

1. Select `Bank Statement` as document type
2. Fill out the form
3. Choose the file
4. Click Upload document

Under the hood, the file will be:

5. Generate a new container at `<podUrl>Bank_Statement/`
6. Place and save the file inside the container at `<podUrl>Bank_Statement/document.pdf`
7. Assigned the file will be assigned the URL `<podUrl>Bank_Statement/document.pdf` 
8. Generate a `document.ttl` file will be generated in that container.
9. Assign the container and `document.pdf` with default ACL settings for the user.

Up to here, that's basically the workflow for uploadDocument. Every afterward is a part of the updateDocument function.

If a user wishes to upload another document to the same directory, say `newDoc.pdf`, to "Bank Statement", the user can follow the same 1-4 as before.

This time, because the container is not empty, a function is called to alert the person that the container is not empty, and the user could decide to either upload the new document or cancel the operation.

If the user cancels the operations. No actions will be taken. If the user wishes to continue, then under the hood:

10. The file will be uploaded to `<podUrl>Bank_Statement/`
11. The file will be assigned the URL `<podUrl>Bank_Statement/newDoc.pdf` 
12. `document.ttl` will be appended by a new Thing instead of being generated using what's filled out in the form in 1-4 (this step is not implemented correctly unfortunate since it uses .set for updating document description instead the .add functions)
13. Assign ACL for `document.pdf` with default ACL settings for user.

At this point, there'll be three files in "Bank Statement" with `document.ttl` pointing to both `<podUrl>Bank_Statement/newDoc.pdf` and `<podUrl>Bank_Statement/document.pdf`

If a user wishes to update a file with the same name, say `document.pdf`, the user would once again perform 1-4, but this time a function will be called alerting the user that "document.pdf" already exist in "Bank Statement" do they wish to continue. This part is the same as before with a new file. If the user cancels the operation, nothing happens. If they wish to continue, then under the hood:

14. The file will be overwritten in `<podUrl>Bank_Statement/document.pdf`
15. `document.ttl` will be updated instead of generated using what's filled out in the form in 1-4 (this part for updating with .set is correct)

So the final result would have "Bank Statement" with `document.ttl` pointing to both `<podUrl>Bank_Statement/newDoc.pdf` and `<podUrl>Bank_Statement/document.pdf`, where `document.pdf` being updated and reflected in the `document.ttl`.

For the signatures, I could see it fitting in nicely between or after 5-9 and 10-13 for file uploads. If `signature.ttl` would need its own container, I could see us using sub containers with the file name and store that in there. But other than that, I don't see any clashes for these two scenarios.

For 14-15 with file updates, I definitely see a need for re-validation as mentioned. By default, the original `signature.ttl` would be invalidate because it's not associated with the updated file, so we can have that removed before the update starts. From there, everything else should be the same as if a new file is being uploaded, verification included.

timbot1789: Thinking about it more, we can actually get decent, stable functionality by just deleting and reuploading. Updated the PR to do that. You can now 'update' (delete and replace) files by uploading a file with the same name.

src/components/Form/DeleteDocumentForm.jsx
timbot1789: To do this, we will need to implement fetch, which will be in a separate PR.

src/components/Form/CrossPodWriteForm.jsx
timbot1789: I removed the username input component because it's unused. We don't upload to usernames, we upload to Pod URLs. Whatever we input, it needs to map to a Pod URL. Usernames don't map to pod URLs any better than human names do, so why not just use their human name? We could have this be a dropdown that lets case managers select from their clients list, but that's a larger UI overhaul.

src/components/Form/CrossPodWriteForm.jsx
timbot1789: updated

src/model-helpers/Document.js
timbot1789: updated

src/model-helpers/Document.js
timbot1789: added in `replaceDocument`

src/components/Form/CrossPodWriteForm.jsx
timbot1789: Oh, it looks like I forgot to pass the session fetch to one of the network calls. Fixed it. This should work now.

src/utils/network/session-core.js
timbot1789: We create the containers with `saveSolidDatasetAt`. That will create all necessary containers along the path, and will avoid creating unused containers. We can inherit permissions from the defaults set in either the root `.acl` or the PASS container `.acl`.

src/components/Form/DeleteDocumentForm.jsx
```diff
@@ -34,25 +38,18 @@ const DeleteDocumentForm = () => {
...
     dispatch({ type: 'SET_PROCESSING' });
-
-    if (!docType) {
-      runNotification('Search failed. Reason: No document type selected.', 5, state, dispatch);
-      setTimeout(() => {
-        dispatch({ type: 'CLEAR_PROCESSING' });
-      }, 3000);
-      return;
-    }
+    const activePod = selectedUser.podUrl || podUrl;
```
leekahung: More of a UI thing not a functionality thing, think it'll be good for the user to see podUrl is active in the form as well.

Could include a line like `name: person` in the JSX with person being pulled from the SelectUserContext and person pull from the profile or the like if it's the user themselves. Could be considered in a future PR or Issue.

src/contexts/SignedInUserContext.jsx
```diff
@@ -51,8 +51,7 @@ export const SignedInUserContextProvider = ({ children }) => {
         });
         await Promise.all([
           updateUserActivity(session, podUrl),
-          createPublicContainer(session, podUrl),
-          createDocumentContainer(session, podUrl)
```
leekahung, line 55: Removing createDocumentContainer should be fine.

The new functionality for createDocument would have build the path to the file so it'll generate Document along the way when uploading any file for the first time and deleteDocument will only delete the specific container with the document in question inside it.

src/model-helpers/Document.js
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * @function createDriversLicenseTtlFile
+ * @memberof utils
+ * @function createDriversLicenseTtlFile
+ * @param {fileObjectType} fileObject - Object containing information about file
+ * @returns {Promise<ThingLocal>} TTL file Thing - Processes a barcode using zxing
+ * and returns a new TTL file Thing
+ */
+const createDriversLicenseTtlFile = async (fileObject) => {
+  const dlData = await getDriversLicenseData(fileObject.file);
```
leekahung: The function originally takes in the file Blob which were contained in fileObject.

It's now just `file` in the codebase, so we can just use `fileObject` for the parameter for getDriversLicenseData as is. We just need to use `file` when passing in for the `createDriversLicenseTtlFile` function.

src/model-helpers/Document.js
```diff
@@ -0,0 +1,206 @@
...
+ * from form submission (see {@link fileObjectType})
+ * @returns {Promise<WordArray>} Promise - Generates checksum for uploaded file
+ * using the SHA256 algorithm
+ */
+const createFileChecksum = async (file) => {
+  const text = await file.text(); // only hash the first megabyte
+  return sha256(text);
+};
+
+const addAdditionalInfo = (thing, fileObject) => {
```
leekahung: Include the `file` parameter so that it could be passed into `createDriversLicenseTtlFile` in Line 103 as `createDriversLicenseTtlFile(file)`

src/model-helpers/Document.js
```diff
@@ -0,0 +1,206 @@
...
+  let thing = buildThing(createThing({ name: docDesc.name }))
+    .addDate(RDF_PREDICATES.uploadDate, new Date())
+    .addStringNoLocale(RDF_PREDICATES.name, docDesc.name)
+    .addStringNoLocale(RDF_PREDICATES.identifier, docDesc.type)
+    .addStringNoLocale(RDF_PREDICATES.sha256, checksum)
+    .addStringNoLocale(RDF_PREDICATES.description, docDesc.description)
+    .addUrl(RDF_PREDICATES.url, `${documentUrl}${file.name}`);
+
+  if (docDesc.date) thing.addDate(RDF_PREDICATES.endDate, new Date(docDesc.date));
+  thing = addAdditionalInfo(thing, docDesc);
```
leekahung: Can change to `thing = addAdditionalInfo(thing, docDesc, file)`

src/model-helpers/Document.js
```diff
@@ -0,0 +1,206 @@
...
+  await saveFileInContainer(docUrl, file, {
+    fetch: session.fetch,
+    slug: file.name
+  });
+  return parseDocument(getThingAll(doc)[0]);
+};
+
+export const createDocument = async (file, fileDescription, session, passUrl) => {
+  const { type, name } = fileDescription;
+  const docUrl = `${passUrl}Documents/${type}/${name}/`;
```
leekahung: Think Solid might have a problem with generating a container with the apostrophe in "Driver's License". We could just remove it for naming the container.

https://github.com/codeforpdx/PASS/assets/14917816/eded1d50-fbd0-434c-9489-851f5adb1492

src/model-helpers/Document.js
```diff
@@ -0,0 +1,206 @@
...
+ * using the SHA256 algorithm
+ */
+const createFileChecksum = async (file) => {
+  const text = await file.text(); // only hash the first megabyte
+  return sha256(text);
+};
+
+const addAdditionalInfo = (thing, fileObject) => {
+  if (fileObject.type === "Driver's License") {
+    return thing.addUrl(RDF_PREDICATES.about, createDriversLicenseTtlFile(fileObject));
```
leekahung: Following the errors with Driver's License, this line might also be the issue for failing to upload a new container. The data is coming out properly, but .addUrl was expecting a URL rather than a Thing indicated by the [[object Promise]] in the clip.

https://github.com/codeforpdx/PASS/assets/14917816/8419b3e3-1bdd-4665-b506-0a9f0348ed98

Think we could solve this by chaining the two Things with setThing into one TTL file rather than putting inside the Thing for docDesc.name.

So we can build the Thing for docDesc.name first in `docDescToThing` and build the Thing with "license information" with `createDriversLicenseTtlFile(file)` inside `saveDescription` if `fileObject.type === "Driver's License"`, then set both Things with setThing to newDataset

src/utils/network/session-core.js
leekahung: Just saw the containers being generated along the path when testing out he upload, so this looks good to me.

leekahung:
Not a big issue at the moment, we could add in JSDoc once we know everything is in working order and in the right form.

src/model-helpers/Document.js
timbot1789: I'm not sure I understand. That's the same as `fileObject` being passed in already. I was able to remove some parameters because they were redundant -- they were already being added. Now instead of having separate parsing functions for each document type, we have one main parsing function that adds the fields common to all documents, and then if a type needs additional info (which is the case for driver's licenses), that's added in the `addAdditionalInfo` function.

Although the nesting for the fields was different than original. Updated the function to keep the original structure.

src/model-helpers/Document.js
timbot1789: See above

src/model-helpers/Document.js
timbot1789: Fixed

src/model-helpers/Document.js
timbot1789: Fixed

src/model-helpers/Document.js
timbot1789: Oh nevermind, I see the bug now. Fixed.

src/model-helpers/Document.js
timbot1789: fixed

src/contexts/SignedInUserContext.jsx
timbot1789: correct

src/components/Form/UploadDocumentForm.jsx
```diff
@@ -24,53 +23,73 @@ const UploadDocumentForm = () => {
...
+      date: expiryDate,
+      description
+    };
+    runNotification(`Uploading "${file.name}" to Solid...`, 3, state, dispatch);
+    const activePod = selectedUser.podUrl || podUrl;
+
+    try {
+      const doc = await createDocument(file, fileDesc, session, `${activePod}PASS/`);
+      if (verifyFile)
+        await signDocument(doc, session, `${activePod}PASS/${docType}/${file.name.split('.')[0]}/`);
```
leekahung: Just notice this now, it looks like an issue that was missed in PR #241.

It seems the url path to signature.ttl is still pointing to the old upload path of `/PASS/<docType>/` instead of the newer `/PASS/Documents/<docType>/` so signature.ttl was sent to the incorrect place.

Screen Shot 2023-06-16 at 0 36 01

src/components/Form/UploadDocumentForm.jsx
```diff
@@ -84,14 +103,14 @@ const UploadDocumentForm = () => {
...
+      <form onSubmit={handleDocUpload} autoComplete="off">
         <label htmlFor="verify-checkbox">
           Verify File on upload:
           <input
             id="verify-checkbox"
             type="checkbox"
-            value={state.verifyFile}
-            onClick={() => dispatch({ type: 'TOGGLE_VERIFY_FILE' })}
+            value={verifyFile}
+            onChange={(e) => setVerifyFile(e.target.value)}
```
leekahung: The setVerifyFile seems to just set the initial state of false for verifyFile since it is referencing the value in line 112.

https://github.com/codeforpdx/PASS/assets/14917816/f1aff375-de6d-483d-a811-ea2b558810a6

Would need to set the onChange to `() => setVerifyFile(!verifyFile)` instead. Think we could use the condition for `checked={verifyFile}` to toggle the state with onChange.

https://github.com/codeforpdx/PASS/assets/14917816/b0f65807-20e8-49f7-a0bb-86ad0c8fc17f

src/components/Form/UploadDocumentForm.jsx
```diff
@@ -24,53 +23,73 @@ const UploadDocumentForm = () => {
   const { session } = useSession();
   const { state, dispatch } = useStatusNotification();
   const [docType, setDocType] = useState('');
+  const [verifyFile, setVerifyFile] = useState(false);
+  const [expiryDate, setExpiryDate] = useState('');
+  const [description, setDescription] = useState('');
+  const [file, setFile] = useState('');
```
leekahung: I think `null` would be a better initial state than an empty string since file will set to a File Object when a file is chosen.

src/components/Form/CrossPodWriteForm.jsx
```diff
@@ -20,69 +19,61 @@ import FormSection from './FormSection';
...
-
 const CrossPodWriteForm = () => {
   const { session } = useSession();
   const { state, dispatch } = useStatusNotification();
-  const { selectedUser } = useContext(SelectUserContext);
-  const [username, setUsername] = useState('');
   const [docType, setDocType] = useState('');
+  const [expiryDate, setExpiryDate] = useState('');
+  const [description, setDescription] = useState('');
+  const [file, setFile] = useState('');
```
leekahung: Similar to `UploadDocumentForm`, think `null` will be the better choice for initial state since setFile takes in a File Object from choose file.

src/components/Form/CrossPodWriteForm.jsx
```diff
@@ -20,69 +19,61 @@ import FormSection from './FormSection';
...
+  const { podUrl, person } = useContext(SelectUserContext);
```
leekahung: Both `podUrl` and `person` comes out as undefined when a user is selected. Should pull data from `selectedUser` instead with `selectedUser.podUrl` and `selectedUser.person`

src/components/Form/CrossPodWriteForm.jsx
```diff
@@ -97,19 +88,11 @@ const CrossPodWriteForm = () => {
       statusType="Upload status"
       defaultMessage="To be uploaded..."
     >
-      <form onSubmit={handleCrossPodUpload} autoComplete="off">
+      <form onSubmit={handleFormSubmission} autoComplete="off">
         <div style={formRowStyle}>
-          <label htmlFor="cross-upload-doc">Upload document to username: </label>
+          <p>Upload document to {person}: </p>
```
leekahung: See comment above about `person` for SelectUserContext.

src/components/Form/UploadDocumentForm.jsx
timbot1789: fixed

src/components/Form/UploadDocumentForm.jsx
timbot1789: fixed

src/components/Form/UploadDocumentForm.jsx
timbot1789: fixed

src/components/Form/CrossPodWriteForm.jsx
timbot1789: fixed

src/components/Form/CrossPodWriteForm.jsx
timbot1789: fixed

src/utils/network/session-core.js
leekahung: Had just tested the cross pod upload, we might want to keep this function after all, but leave it such that it'll only generate the base Documents container.

https://github.com/codeforpdx/PASS/assets/14917816/c44a0d20-e45d-4e09-9215-ff2efbf6e51c

The `setDocContainerAclPermission` would need an anchor point for the user to given permissions to other Pods. We can simply generate the Documents container with the initial ACL set for the user and the `container.ttl`.

So we can essentially delete everything after the first try/catch block for `createDocumentContainer` removing everything related to createContainerList.

For `setDocContainerAclPermission`, it'll be simplified to just:

``` JavaScript
export const setDocContainerAclPermission = async (session, permissions, otherPodUsername) => {
  const otherPodUrl = getPodUrl(otherPodUsername);
  const webId = `${otherPodUrl}profile/card#me`;
  await setDocAclForUser(session, containerUrl, 'update', webId, permissions);
};
```

I've tried out the changes on my computer and it works well with the new functions created in this PR.

https://github.com/codeforpdx/PASS/assets/14917816/ee697b31-514c-4fe7-a946-3b3d8602882b

https://github.com/codeforpdx/PASS/assets/14917816/0aacffc4-cd9b-4201-95df-af21299b7859

After these fixes, I think everything works properly as intended in this PR:
- user can upload documents to their own pod
- user can upload multiple document files to their own pod
- user can update documents to their own pod
- user can delete documents in their own pod
- cross pod uploads can be achieved by other users after given permission by pod owner

The fetch functionality for delete document would be good if it's in this PR, but that could be put in a follow-up. I'm good either way.

Think this is close to complete.

leekahung:
After this PR, "Search Document", "Cross Pod Search", and "Permission to Files" forms would need to be updated.

For "Search Document" and "Cross Pod Search", it's mainly due to the new routing. The fixes would probably follow the same solution we have for deleting files from the Pod except instead of deleting we're just returning a container url of that file.

For "Permission to Files", This is related to child container under Documents not having ACL files attached to them and `setDocAclPermission` explicitly searches for that ACL to update permissions. A solution for this could be to attach an ACL file for the container with the file.

src/utils/network/session-core.js
timbot1789: > The setDocContainerAclPermission would need an anchor point for the user to given permissions to other Pods. We can simply generate the Documents container with the initial ACL set for the user and the container.ttl.

Can we use the `PASS` container as that anchorpoint? Then just have any container underneath it inherit those permissions by default?  Then we only need to call the ACL setup function once, when the user registers, and we never have to think about it again. If that ACL changes, it's because the user intended to change it. Then if the client wants to revoke our permission to access their pod, they can just delete the one ACL. If they want to revoke permission to a sub container, they can use a different flow to add an ACL that overrides the default set at the PASS level.

timbot1789: I don't think we need to make document upload worry about permissions. Document upload already has enough to think about. I think if upload fails because the user is unauthorized, we can just make the user go to a different flow to get authorization, then try to upload again.

leekahung: Think it should be just limited to Documents. We don't want other users to have full access to PASS, just for the original user.

timbot1789: I don't understand what you mean. I agree the only people who should have access to a PASS container and its sub containers are the ones authorized by the owner of the pod. That's what's demonstrated in your video. The user you were uploading with didn't have access to that pod, so the upload what rejected.

I think you're talking about the question: how do we give the case manager permission to upload to the client's pod?

My point is that we can separate this from the question: how does the case manager actually upload to the client's pod?

This PR is about the latter. I think we can solve the former somewhere else. If you go to your client pod, go to the PASS container, and upload an ACL that gives your case manager read-write permissions to everything under that container, then this code should work.

leekahung: > If you go to your client pod, go to the PASS container, and upload an ACL that gives your case manager read-write permissions to everything under that container, then this code should work.

Would agree that setting ACL settings would be ideal in the Documents container rather than in the PASS container. It's similar to how Credentials are set up previously.

The main difference between Documents and Credentials should be that we don't have a `setDocContainerAclPermission` equivalent for Credentials to make changes to that ACL.

leekahung: We can checkout a branch from this branch and make changes related to permission setting and createDocumentsContainer?

leekahung:
Had finished discussion with Tim. Might be better to do permissions setting, search, and delete functionalities in follow-up PRs, focus just on allowing file uploads and updates in this PR. So users would still be able to upload and update file on their Pod.

Search, delete, and permission setting functionalities will be temporarily be broken.

src/components/Documents/DocumentTableRow.jsx
```diff
@@ -0,0 +1,68 @@
...
+    ) {
+      return;
+    }
+    await removeDocument(document.name);
+  };
+
+  return (
+    <StyledTableRow>
+      <StyledTableCell align="center">{name}</StyledTableCell>
+      <StyledTableCell align="center">{type}</StyledTableCell>
```
leekahung: For `type` think we could revert this back to the original string format with spacing. Would really just affect `types` with more than one word. Should be fine for the most part.

src/components/Form/CheckAclPermsDocContainerForm.jsx
```diff
@@ -28,8 +28,8 @@ import FormSection from './FormSection';
 const CheckAclPermsDocContainerForm = () => {
   const { session } = useSession();
   const { state, dispatch } = useStatusNotification();
-  const { selectedUser } = useContext(SelectUserContext);
-  const [username, setUsername] = useState('');
+  const { selectedUser } = useContext(SelectedUserContext);
+  const [username, setUsername] = useState(selectedUser.username);
```
leekahung: I see selectedUser defaults to the user's own podUrl if unselected. Think it's a great idea. It'll simply a lot of the functions we have to just use podUrl.

For the text displayed in "Client selected", think we could have it show "Personal Pod" instead of the podUrl. Think we could include an entry to the object so that it'll be `{podUrl, person: "Personal Pod" }` for the default selectedUser object or something of the like.

src/components/Form/UploadDocumentForm.jsx
```diff
@@ -32,58 +29,71 @@ import FormSection from './FormSection';
...
 
-  // Initialized state for file upload
-  const handleFileChange = (event) => {
-    if (event.target.files.length === 1) {
-      dispatch({ type: 'SET_FILE', payload: event.target.files[0] });
-    } else {
-      dispatch({ type: 'CLEAR_FILE' });
-    }
-  };
+  // Custom useField hook for handling form inputs
```
leekahung: Think we can remove this. No longer using useField here.

src/components/Form/UploadDocumentForm.jsx
```diff
@@ -32,58 +29,71 @@ import FormSection from './FormSection';
...
-    dispatch({ type: 'CLEAR_VERIFY_FILE' });
-    dispatch({ type: 'CLEAR_PROCESSING' });
+    setVerifyFile(false);
+    setDocType('');
+    setFile('');
+    setInputKey(!inputKey); // clears file by forcing re-render
     setDocDescription('');
     setDocType('');
-    setExpireDate(null);
+    setExpireDate('');
```
leekahung: The Dayjs value used in the date picker is a Dayjs object. Think it'll be more appropriate to use a null.

This is mainly an MUI validation issue. Should still work, after clearing it just triggers the error until a new date is selected.

https://github.com/codeforpdx/PASS/assets/14917816/774f7990-37d0-4201-ba9d-0af791e5775c

src/components/Form/UploadDocumentForm.jsx
```diff
@@ -144,7 +154,7 @@ const UploadDocumentForm = () => {
               color="primary"
               id="upload-doctype"
               name="uploadDoctype"
-              onChange={handleFileChange}
+              onChange={(e) => setFile(e.target.files[0])}
               required
             >
               Choose file
```
leekahung, line 158: It was missed in an earlier PR, but I found the issue now. In Line 161, the `accept` is missing a comma between .rtf and .gif. It'll miss those files as acceptable file extensions.

src/contexts/SelectedUserContext.jsx
```diff
@@ -0,0 +1,32 @@
...
+export const SelectedUserContext = createContext({});
+
+export const SelectedUserContextProvider = ({ children }) => {
+  const { podUrl } = useContext(SignedInUserContext);
+  const [selectedUser, setSelectedUser] = useState({ podUrl });
+
+  const selectedUserMemo = useMemo(
+    () => ({
+      selectedUser,
+      selectUser: async (user) => setSelectedUser(user || { podUrl })
```
leekahung: Mentioned in a comment earlier, think we could include an entry for `person: "Personal Pod"` in the default `{ podUrl }` is `user` is not selected from SelectedUserContext.

src/model-helpers/DocumentList.js
```diff
@@ -0,0 +1,97 @@
...
+  createSolidDataset,
+  saveFileInContainer,
+  getThing,
+  removeThing,
+  deleteFile
+} from '@inrupt/solid-client';
+
+import { makeDocIntoThing, parseDocFromThing } from './Document';
+
+const parseFromDataset = async (docsDataset, session) => {
```
leekahung: Think we can get away with dropping the async/await completely in this function. None of the get functions require using session, which does the fetch via session.fetch so we don't have to deal with any promises.

src/model-helpers/DocumentList.js
```diff
@@ -0,0 +1,97 @@
...
+  const docList = [];
+  const allDocThings = getThingAll(docsDataset);
+  await Promise.all(
+    allDocThings.map(async (userThing) => {
```
leekahung: Can switch to forEach after dropping the Promise.all

src/model-helpers/DocumentList.js
```diff
@@ -0,0 +1,97 @@
...
+      const docObject = await parseDocFromThing(userThing, session);
```
leekahung, line 20: parseDocFromThing() doesn't take in a session parameter.

leekahung: parseDocFromThing() doesn't take in a session as a parameter it looks like. Think we can remove session as a parameter completely from parseFromDataset() since that's the only place it's being used for.

src/model-helpers/DocumentList.js
```diff
@@ -0,0 +1,97 @@
...
+  );
+
+  return docList;
+};
+
+const saveToPod = async (session, { dataset, containerUrl }) => {
+  const newDataset = await saveSolidDatasetAt(`${containerUrl}doclist.ttl`, dataset, {
+    fetch: session.fetch
+  });
+  const newList = await parseFromDataset(newDataset, session);
```
leekahung: parseFromDataset() is taking in session as a parameter, but parseDocFromThing() doesn't utilize it. Can be removed along with the await.

src/routes/Forms.jsx
```diff
@@ -42,26 +39,14 @@ const Forms = () => {
...
-        size="small"
-        aria-label="Clear Client Button"
-        startIcon={<Remove />}
-        onClick={() => setSelectedUser('')}
-        sx={{ margin: '2rem 0 1rem' }}
-      >
-        Clear Client
-      </Button>
-      <Typography>Client selected: {selectedUser.person || 'No client selected'}</Typography>
+      <Typography>Client selected: {selectedUser.person || selectedUser.podUrl}</Typography>
```
leekahung: Mentioned about setting default selectedUser.person as "Personal Pod". Open to other options to indicate the user's own Pod.

src/components/Form/UploadDocumentForm.jsx
timbot1789: fixed

src/components/Form/CheckAclPermsDocContainerForm.jsx
timbot1789: fixed

src/components/Form/UploadDocumentForm.jsx
timbot1789: fixed

src/contexts/SelectedUserContext.jsx
timbot1789: That sounds more like display text than data to me. Added it to the form page.

src/model-helpers/DocumentList.js
timbot1789: fixed

src/model-helpers/DocumentList.js
timbot1789: fixed

src/model-helpers/DocumentList.js
timbot1789: fixed

src/components/Form/UploadDocumentForm.jsx
timbot1789: fixed

src/model-helpers/DocumentList.js
timbot1789: fixed

src/routes/Forms.jsx
timbot1789: fixed

src/routes/Forms.jsx
```diff
@@ -42,26 +40,19 @@ const Forms = () => {
         justifyContent: 'center'
       }}
     >
-      <Button
```
leekahung: Think we could keep the button for clearing selected users here. That way case workers could just unselect them here rather than having to go back to "Clients".

src/components/Documents/DocumentTableRow.jsx
timbot1789: updated

src/routes/Forms.jsx
timbot1789: fixed

leekahung (commented):
Think everything is working as intended in this PR. Permissions setting within PASS would be disabled briefly, developers and advance users could still mess with those settings manually on their Pod. Believe this would be fixed in the update with the user registration.

Other than that, small nitpick on JSDoc. Think we could add some JSDoc for functions that's being exported in Document.js and DocumentList.js in model-helpers like what's done in User.js and UserLIst.js? Could help catch some of the bugs related to parameters or function types like with parseFromDataset() in DocumentList.

Also, just noticed we have 2 parseFromDataset() functions. One in DocumentList.js and one in UserList.js. It's fine since they're internal functions for parsing different types of TTL files, but would need to consider renaming if going to be exported in the future like parseUserFromDataset and parseDocumentFromDataset.

Other than that, I'm good with merging.

timbot1789:
> Could help catch some of the bugs related to parameters or function types like with parseFromDataset() in DocumentList.

I'm not sure what tooling is necessary to use JSDoc in this way, but we don't have it. JSDoc is currently being used purely for documentation. It's not catching any bugs.

Either way, I added the docs

leekahung:
Update: Might be a bug from solidcommunity.net with regards to trying to add users to a clients list, but can't really confirm that right now due to the extreme server lag. (see the server timing out in the console.log before trying opencommons.net)

However has managed to get opencommons.net working instead to check the functionality and it seems to work okay. I'm not sure if the functionality in solidcommunity.net broke because it was unstable or if it was something else. If it's just a hiccup from the server over at solidcommunity.net, I think everything should be okay. Would likely be problem free if testing locally or on opencommons.net.

https://github.com/codeforpdx/PASS/assets/14917816/b40f0e73-7418-429d-b5c1-9ad80a3afae3

Old: Hey Tim, have just checked the functionality for the Client list again after clearing my Pod just to make sure everything is working properly. Something is breaking in the UserListContext. Been digging into UserList a bit and the functions it says are related to the error (see below), but having little luck.

Screen Shot 2023-06-19 at 17 06 54

src/model-helpers/Document.js
```diff
@@ -0,0 +1,168 @@
...
+ * from form submission (see {@link fileObjectType})
+ * @returns {Promise<WordArray>} Promise - Generates checksum for uploaded file
+ * using the SHA256 algorithm
+ */
+const createFileChecksum = async (file) => {
+  const text = await file.text(); // only hash the first megabyte
+  return sha256(text);
+};
+
+/**
```
leekahung, line 106: Think the JSDoc added will do for now. Can do some clean-up to make it more detailed in another PR.

leekahung (approved):
Think this looks okay as it. We know where in PASS is working and where in PASS is breaking. I'll approve.

PR #254, opened 2023-06-13
Issue 250/pass file viewer

### LINKED ISSUE #250: Add File Viewer to PASS
Currently, if a case manager wants to view a file on a client's pod, they must first download the file and then open it in a separate file viewer on their computer. This is a rather annoying workflow, as it requires the case manager to exit PASS. There's also a data privacy risk if the case manager doesn't clear the file from their downloads after viewing it.

We can solve both issues by adding a file viewer to PASS. The case manager would be able to open and read the file without exiting PASS, and PASS would be able to enforce any privacy cleanup or viewing requirements it deems appropriate. We can still have the download link as an alternative flow, but add a popup to it that reminds them to delete the file after downloading. 

**Additional context**
We'd want to support at least these files: PDF, PNG, JPG

Some other additional file types: TXT, HEIC, TTL, TIFF

HEIC is quite difficult to support, but would be really useful to add. It's the image format Iphone uses, and Apple won't release a spec for it. Support for it is generally poor (safari doesn't support it), because apple won't release a spec for it.

leekahung:
# PR Details

Have setup the functionality to utilize the browser's default file viewer. Using what is there already for DocumentTable with File URL to display the file as a local file Blob. The view may vary depending on the Browser, but would let the user use what they're most familiar with. The example below is for Chrome.

https://github.com/codeforpdx/PASS/assets/14917816/254111bb-1607-4f8a-a775-d8f2f74c24f9

Safari:

https://github.com/codeforpdx/PASS/assets/14917816/9333799a-5d09-4f58-b0a4-3deb7cf030a0

Edge:

https://github.com/codeforpdx/PASS/assets/14917816/3631938f-7c60-40b0-914e-52831f269f5f

FireFox:

https://github.com/codeforpdx/PASS/assets/14917816/62f68d16-68ca-44f8-9e65-a1338d3644ef


andycwilliams:
Haven't combed through the code yet. These are just my immediate thoughts.

In terms of functioning, it seems very nice. Anything we can do to help users not have to interact directly with Solid is good.

If we expect the user to end up with large amounts of files, then creating an option to view them as a list would be good. Nowadays it's common to have multiple viewing options, so a list, as well as tiles (similar to what's in the demo), would be very user-friendly.

We may also want to format any dates we present to the user. They don't really need anything other than the date and time. We can filter out the rest.

On the desktop version, maybe consider auto-expanding windows. That could make things slightly less annoying when dealing with larger files like PDFs, especially when there's plenty of space to use.

Maybe don't call the button 'Local' as it isn't as intuitive as, say, 'Open' in my opinion.

For 'New Window', we may want a confirmation dialog box. It can be annoying to accidentally click on a button only for it to take you to a different page or window.

What are the instances where we expect the user to want to select a new window? My presumption would be that they'd want to do everything in-app.

Really good feature though.

leekahung:
@timbot1789 @xscottxbrownx

Would probably need to be reworked after #249 is merged with the new pathing to files, but it should work for the existing version of Development. Let me know what you think.

xscottxbrownx:
Still digging into this... 

The idea is a must. Should be able to view the documents within PASS.
I'm with Andy and not sure a `new window` option is needed at all.
I'm also not sure the first modal `ShowDocumentsModal` is needed. 

---

### So, while quickly testing (I have lots more to do for a proper review) I noticed a few things:
1) The `Upload Document` seems to take a really long time and the `upload status` messaging is a little strange during upload (but this may be due to the specific files I am uploading as testers.)
2) When clicking the view files icon in `Clients`, they all change to disabled coloring. Probably an issue with state like I originally had with the pin icons.
3) Got breaking errors
Screenshot 2023-06-18 at 7 59 56 AM

I uploaded a random pdf to "Bank Statement" and that worked fine. I could view when clicking on icon in `Clients`, and I could view in the new `Users` tab.  
 
Then, I uploaded a different random pdf to "Passport" and uploading was fine (besides noted issues in 1 above.) But, when I went to view from either `Clients` or clicked `Show Documents` in the `Users` page - the above error occurred and browser was just blank/white.

#### Error is probably something to do with the files I was using to upload to it. 
I used different pdf files for each, but when I deleted the "Passport" file, it all worked fine again. Then, I uploaded a new "Passport" using the same pdf I used for "Bank Statement" and it worked fine.   But this is strange as "Bank Statement" pdf was actually more pages and a larger file size than the one I originally used for "Passport".

My pdfs don't seem to upload to "Driver's License" but no errors in console.  Again, could just be a file type/size issue.  We definitely need to get the form/input validations going hahah.


---

What was your thought with the `Users` page? I'm curious of it's purpose.  
Wouldn't the `Preview Documents` just live in current `Forms` page? - And ultimately just be refactored into an icon with an onClick when `Forms` becomes a usable `Documents` page like the `Clients` overhaul?
Just curious if we are creating/adding more than needed to the codebase here.

src/components/Clients/ClientListTableRow.jsx
```diff
@@ -101,6 +144,16 @@ const ClientListTableRow = ({ labelId, client, state, dispatch }) => {
         </Link>
       </StyledTableCell>
       <StyledTableCell align="center">{modifiedDate}</StyledTableCell>
+      <StyledTableCell align="center">
+        <IconButton disabled={state.processing} size="large" edge="end" onClick={handleShowFile}>
+          <FindInPageIcon />
```
xscottxbrownx: this seems like an icon that would be used for search.  

maybe use the "preview" icon instead?  And the outline version of whichever to match the other icons?

---

Also, all the icons of this type change when clicking one.

src/components/DocumentModals/ShowDocumentsModal.jsx
```diff
@@ -0,0 +1,131 @@
+// React Imports
```
xscottxbrownx: ## Do we need this?

Should clicking the `view documents` icon in client list open a modal, or just take you to a documents page? 

If yourself was in clients list, and you click `view documents` I would think take you to a page that lists all your documents (like `Clients` is now.)

If clicking a different user's `view documents` in clients list, I would think might be better as an accordion that lists all the files - then when click a single file from there, opens a modal to view.    Maybe this should be the same for yourself in the list too?

Maybe this just means the first iteration of clients list is just not good enough yet. Maybe when clicking on client it brings up modal of everything tied to that client with clickable links (like each document), or accordions to do the same roughly.

src/components/DocumentModals/ShowDocumentsModal.jsx
```diff
@@ -0,0 +1,131 @@
...
+              key={uuidv4()}
+              sx={{
+                padding: 5,
+                borderRadius: '10px',
+                backgroundColor: 'lightgray',
+                width: { xxs: '300px', xs: '320px' }
+              }}
+            >
+              <Box sx={{ display: 'flex', flexDirection: 'column', gap: 1 }}>
+                <Typography sx={{ fontWeight: 'bold' }}>{src.documentType}</Typography>
```
xscottxbrownx: center this title

src/components/DocumentModals/ShowDocumentsModal.jsx
```diff
@@ -0,0 +1,131 @@
...
+                <Typography
```
xscottxbrownx: I'm not sure what info here we actually need or is helpful to the user... maybe expiration date? Maybe upload date to know how recent it is? 

Not sure file name or description is needed (since we have the docType in the title.)

src/components/DocumentModals/ShowDocumentsModal.jsx
```diff
@@ -0,0 +1,131 @@
...
+                    textOverflow: 'ellipsis'
+                  }}
+                >
+                  File name: {src.filename}
+                </Typography>
+                <Typography>Upload date: {src.uploadDate.toLocaleString()}</Typography>
+                <Typography>Description: {src.description}</Typography>
+                <Typography>Expire Date (if applicable): {src.expireDate}</Typography>
+                <br />
+                <Typography sx={{ fontWeight: 'bold' }}>Show Document</Typography>
```
xscottxbrownx: center this if keeping, I'm not sure it's needed though. I think the buttons kind of express this.

src/components/DocumentModals/ShowDocumentsModal.jsx
```diff
@@ -0,0 +1,131 @@
...
+                  Preview
+                </Button>
+                <Button
+                  variant="contained"
+                  type="button"
+                  sx={{
+                    a: { color: 'white', textDecoration: 'none', width: '100%', height: '100%' }
+                  }}
+                >
+                  <a href={src.documentUrl} target="_blank" rel="noreferrer">
```
xscottxbrownx: Think could use MUI's `link` component here (https://mui.com/material-ui/react-link/), which would also allow for the styling to be applied on it in these lines, instead of on the parent button above.)

src/components/DocumentModals/ShowDocumentsModal.jsx
```diff
@@ -0,0 +1,131 @@
...
+                  type="button"
+                  onClick={() => handleShowDocumentLocal(src.documentUrl)}
+                >
+                  Preview
+                </Button>
+                <Button
+                  variant="contained"
+                  type="button"
+                  sx={{
+                    a: { color: 'white', textDecoration: 'none', width: '100%', height: '100%' }
```
xscottxbrownx: think can just use the `fullWidth` attribute: https://mui.com/material-ui/api/button/

src/AppRoutes.jsx
```diff
@@ -48,6 +44,7 @@ const AppRoutes = () => {
       />
       <Route element={<ProtectedRoute isLoggedIn={session.info.isLoggedIn} />}>
         <Route path="/PASS/clients" element={<Clients />} />
+        <Route path="/PASS/user" element={<User />} />
```
xscottxbrownx: is this whole new page/route needed? Seems like what is on it would fit fine in the `forms`.   There may be future thoughts about it that I'm unaware of though.

src/components/Clients/ClientListTableRow.jsx
```diff
@@ -1,20 +1,24 @@
...
 import DeleteOutlineOutlinedIcon from '@mui/icons-material/DeleteOutlineOutlined';
 import PushPinIcon from '@mui/icons-material/PushPin';
 import PushPinOutlinedIcon from '@mui/icons-material/PushPinOutlined';
 // Utility Imports
-import { runNotification } from '../../utils';
+import { runNotification, getDocTTLs } from '../../utils';
 // Context Imports
 import { SelectUserContext, UserListContext } from '../../contexts';
+// Component Imports
+import { ShowDocumentsModal } from '../DocumentModals';
```
xscottxbrownx: is this needed?   see comments for `ShowDocumentsModal.jsx`

src/components/Clients/ClientListTableRow.jsx
```diff
@@ -71,6 +75,45 @@ const ClientListTableRow = ({ labelId, client, state, dispatch }) => {
     setPinned(!pinned);
   };
 
+  // EXPERIMENTAL - Event handler for displaying Documents from PASS
+  const { session } = useSession();
+  const [fileSrc, setFileSrc] = useState([]);
+  const [showDocument, setShowDocument] = useState(false);
+
+  const handleShowFile = async () => {
```
xscottxbrownx: not necessary, but had a thought to pull out the code block inside the `if`s of an error into it's own function to reduce repetitive code.  

and COULD make the `allDocumentData.message` slightly more clear by storing in a variable named `errorMessage`

which would result in something like this:

```js
const handleFailure = (reason) => {
  runNotification(`Operation failed. Reason: ${reason}`, 5, state, dispatch);
  dispatch({ type: 'CLEAR_PROCESSING' });
};

const handleShowFile = async () => {
  runNotification('Fetching documents from Pod...', 5, state, dispatch);
  dispatch({ type: 'SET_PROCESSING' });

  const allDocumentData = await getDocTTLs(session, client.podUrl);
  const errorMessage = allDocumentData.message;

  if (errorMessage?.includes('Unauthorized')) {
    return handleFailure('No permission to files.');
  }

  if (errorMessage?.includes('No documents found')) {
    return handleFailure('No files found.');
  }

  const permittedData = allDocumentData.filter(item => typeof item !== 'number');

  if (permittedData.length === 0) {
    return handleFailure('No permitted files found.');
  }

  setFileSrc(permittedData);
  setShowDocument(!showDocument);

  setTimeout(() => {
    dispatch({ type: 'CLEAR_PROCESSING' });
  }, 5000);
};
```

src/components/DocumentModals/ShowDocumentLocalModal.jsx
```diff
@@ -0,0 +1,54 @@
...
+// Material UI Imports
+import Button from '@mui/material/Button';
+import Dialog from '@mui/material/Dialog';
+import DialogActions from '@mui/material/DialogActions';
+import DialogContent from '@mui/material/DialogContent';
+import DialogTitle from '@mui/material/DialogTitle';
+import ClearIcon from '@mui/icons-material/Clear';
+
+/**
+ * ShowDocumentsModal Component - Popup modal that shows client documents if gotten
```
xscottxbrownx: `ShowDocumentLocalModal Component`*

src/components/DocumentModals/ShowDocumentLocalModal.jsx
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+const ShowDocumentLocalModal = ({ showModal, setShowModal, fileBlobUrl }) => (
+  <Dialog
+    fullScreen
+    open={showModal}
+    aria-labelledby="dialog-title"
+    onClose={() => setShowModal(false)}
+  >
+    <DialogTitle id="dialog-tile">Documents</DialogTitle>
```
xscottxbrownx: should this be the file name or docType instead maybe?

src/components/DocumentModals/ShowDocumentsModal.jsx
```diff
@@ -0,0 +1,131 @@
...
+            display: 'flex',
+            flexWrap: 'wrap',
+            alignItems: 'center',
+            justifyContent: 'center',
+            gap: 5
+          }}
+        >
+          {fileSrc.map((src) => (
+            <Box
+              key={uuidv4()}
```
xscottxbrownx: what is this?  I've been meaning to ask for a while what this library was

src/components/DocumentModals/ShowDocumentsModal.jsx
```diff
@@ -0,0 +1,131 @@
...
+                <br />
+                <Typography sx={{ fontWeight: 'bold' }}>Show Document</Typography>
+                <Button
+                  variant="contained"
+                  type="button"
+                  onClick={() => handleShowDocumentLocal(src.documentUrl)}
+                >
+                  Preview
+                </Button>
+                <Button
```
xscottxbrownx: Not sure this is needed...

src/components/DocumentModals/index.js
```diff
@@ -0,0 +1,11 @@
+import ShowDocumentLocalModal from './ShowDocumentLocalModal';
```
xscottxbrownx: Think maybe name of folder and `@namespace` might be better as ViewDocuments or something else, because there will be plenty of document modals used in `Forms`/`Documents`.  May be better for readability if this was something else.

src/routes/User.jsx
```diff
@@ -0,0 +1,95 @@
+// React Imputs
```
xscottxbrownx: Same comment as `AppRoutes`... is this new page/route needed? Does this not fit in `Forms`?

src/routes/User.jsx
```diff
@@ -0,0 +1,95 @@
...
+ * @returns {React.ReactElement} The user page
+ */
+
+const User = () => {
+  const location = useLocation();
+  localStorage.setItem('restorePath', location.pathname);
+
+  const { state, dispatch } = useStatusNotification();
+
+  // EXPERIMENTAL - Event handler for displaying Documents from PASS
```
xscottxbrownx: seems to be extremely similar to function within `ClientListTableRow.jsx`, could we refactor/pull out and make work for both?

src/routes/User.jsx
```diff
@@ -0,0 +1,95 @@
...
+    setFileSrc(dataWithFiles);
+    setShowDocument(!showDocument);
+
+    setTimeout(() => {
+      dispatch({ type: 'CLEAR_PROCESSING' });
+    }, 5000);
+  };
+
+  return (
+    <FormSection
```
xscottxbrownx: Again, I feel like I'm missing the point (or future point) of this `User` page.  If this is the only thing we don't need it as a form and button combo.  Just run the function and display the documents (like `Clients`.)

xscottxbrownx (commented):
Think some of this requires more info on thoughts or a discussion, but great work overall. A must have feature I feel.

leekahung:
> 1) The `Upload Document` seems to take a really long time and the `upload status` messaging is a little strange during upload (but this may be due to the specific files I am uploading as testers.)

This I'm not too sure. The server on solidcommunity.net has been slow as of late, but it also just be the functionality. Thinking of checking using opencommons.net or localhost.

> 2) When clicking the view files icon in `Clients`, they all change to disabled coloring. Probably an issue with state like I originally had with the pin icons.

This was intended. That way, it'll focus on one user’s set of documents.

> 3) Got breaking errors
> Screenshot 2023-06-18 at 7 59 56 AM
> 
> I uploaded a random pdf to "Bank Statement" and that worked fine. I could view when clicking on icon in `Clients`, and I could view in the new `Users` tab.  
>  
> Then, I uploaded a different random pdf to "Passport" and uploading was fine (besides noted issues in 1 above.) But, when I went to view from either `Clients` or clicked `Show Documents` in the `Users` page - the above error occurred and browser was just blank/white.
> 
> #### Error is probably something to do with the files I was using to upload to it. 
> I used different pdf files for each, but when I deleted the "Passport" file, it all worked fine again. Then, I uploaded a new "Passport" using the same pdf I used for "Bank Statement" and it worked fine.   But this is strange as "Bank Statement" pdf was actually more pages and a larger file size than the one I originally used for "Passport".

That I'm not sure. I've followed a similar userflow before, but have not encountered that error. Would like to test it again with your files to see if I get the same errors.

> My pdfs don't seem to upload to "Driver's License" but no errors in console.  Again, could just be a file type/size issue.  We definitely need to get the form/input validations going hahah.

I think it might have to due with file type. Have only tested uploading driver's license with a picture of a barcode since the scanner has been implemented. Upload for other file types might be buggy.

> What was your thought with the `Users` page? I'm curious of it's purpose.

The Clients page is something that's only accessible to case managers or case workers. The User page was meant to be something for any user, a kind of user portal to view things in like documents.
  
> Wouldn't the `Preview Documents` just live in current `Forms` page?

I do see some forms from Forms moving into User or used as modals in User.

src/components/Clients/ClientListTableRow.jsx
leekahung: I've checked the icons on MUI. This one feels better for the role. Think we could use a straight-up magnifying glass for search.

src/components/DocumentModals/ShowDocumentsModal.jsx
leekahung: If it's just previewing documents, I think a modal should suffice. There'll be more relevant information about the client on the Clients page

src/components/DocumentModals/ShowDocumentsModal.jsx
leekahung: Think Upload date could be included

src/components/DocumentModals/ShowDocumentsModal.jsx
leekahung: Think this is fine as is

src/components/DocumentModals/ShowDocumentsModal.jsx
leekahung: This is fine as is I think

src/AppRoutes.jsx
leekahung: Yes, it's going to be a full page for the user using PASS.

src/components/Clients/ClientListTableRow.jsx
leekahung: See comments related to the modal.

src/components/Clients/ClientListTableRow.jsx
leekahung: Think we can add a parameter for time in handleFailure. I'll implement it

src/components/DocumentModals/ShowDocumentLocalModal.jsx
leekahung: Just Documents

src/components/DocumentModals/ShowDocumentsModal.jsx
leekahung: It's a function for generating unique keys if you do not have one but need something.

src/components/DocumentModals/ShowDocumentsModal.jsx
leekahung: This is mainly for users who wish to see a full separate window for documents on desktops.

src/components/DocumentModals/index.js
leekahung: Thinking of just calling it modals and moving everything else in here in the future

src/routes/User.jsx
leekahung: The opposite, actually. Think we could move some forms in user instead like upload and delete documents.

src/routes/User.jsx
leekahung: Yeah, think that would be good. I'll refactor these

src/routes/User.jsx
leekahung: User would expand, so it'll be fine

src/routes/User.jsx
xscottxbrownx: Ok, so when building the list of all the user's documents (let's just say it looks like the client list for an easy visual), that should go in `Users`?   This preview would happen onClick of a document.  Then there'd be other things from `forms` like search documents, delete document, etc ?

leekahung: I see the forms for upload, etc. on the User page itself rather than on the modal. The modal will just be for the file viewer.

src/components/DocumentModals/ShowDocumentsModal.jsx
andycwilliams: Agreed that a modal is all that should be needed here

src/components/DocumentModals/ShowDocumentsModal.jsx
leekahung: Yeah, I see both upload ate and expiration data included.

src/components/DocumentModals/ShowDocumentsModal.jsx
leekahung: Getting this switched to MUI's Link.

src/routes/User.jsx
leekahung: Refactored to functions inside file-viewer-helper.js in utils/frontend

src/components/Clients/ClientListTableRow.jsx
leekahung: Had modified how error message comes out of `handleShowFile` in the refactoring, but the effect should be the same as this. See changes to `handleShowFile` in `User.jsx` and `ClientListTableRow.jsx` and the new `runCheckFiles` in `file-viewer-helper.js`

src/components/DocumentModals/ShowDocumentsModal.jsx
timbot1789: It would be better to use file names or some other value on the `src` as the keys. Then the key can be used by screen readers or other parts of the code base, or used in testing. As is, this will generate a meaningless number every single time the code is run.

src/components/DocumentModals/ShowDocumentsModal.jsx
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * ShowDocumentsModal Component - Popup modal that shows client documents if gotten
+ * explicit permission via ACL
+ *
+ * @memberof DocumentModals
+ * @name ShowDocumentsModal
+ */
+
+const ShowDocumentsModal = ({ showModal, setShowModal, fileSrc }) => {
```
timbot1789: What's the difference between this and `ShowDocumentsLocalModal`?

src/utils/barcode/barcode-scan.js
```diff
@@ -83,15 +83,16 @@ const csvToJson = (csvData) => {
...
 const getDriversLicenseData = async (file) => {
-  const image = await readImageFile(file);
-  const decoded = await decodeBarcode(image);
-  let returnedData = '';
-  if (decoded) {
-    const jsonData = csvToJson(decoded);
-    returnedData = jsonData;
-  } else {
-    returnedData = new Error('Unable to decode file');
+  let returnedData = {};
```
timbot1789: What are these changes?

src/routes/User.jsx
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * User - Page that displays the User's stored Documents
+ *
+ * @memberof Pages
+ * @name User
+ * @returns {React.ReactElement} The user page
+ */
+
+const User = () => {
```
timbot1789: Is this needed? I feel like it displays the same information already present in clients and forms. Either way, I feel like it should have a different name. As is, it's confusing for both the developer and the case manager. The case manager won't know who the 'user' is, and developers may confuse this page with the `User` model object (which is supposed to refer to a generic user, not specifically the signed in user).

Maybe we can pass this idea through UX a bit more and revisit it in a later PR.

src/utils/frontend/file-viewer-helper.js
```diff
@@ -0,0 +1,22 @@
+import { getDocTTLs } from '../network/session-core';
+
+const runCheckFiles = async (session, podUrl) => {
```
timbot1789: I'm not sure I understand what this function is for. The server handles checking permissions. In fact, a well-implemented server won't give you enough information to tell if you're unauthorized, or if the document you're trying to fetch doesn't exist.

Just fetch all the documents you can from the server, and display whatever you get back. No need to pass stuff through another check.

src/utils/network/session-core.js
```diff
@@ -560,3 +567,76 @@ export const createInbox = async (session, podUrl) => {
...
+/**
+ * A function that shows the documents from Solid Pod from user into PASS
+ *
+ * @function getDocTTLs
+ * @param {Session} session - Solid's Session Object {@link Session}
+ * @param {URL} podUrl - Pod URL of user to show documents from
+ * @returns {Promise<object[]>} - List of parsed objects and/or error status codes
+ */
+
+export const getDocTTLs = async (session, podUrl) => {
```
timbot1789: A lot of this duplicate functionality of what's being written in https://github.com/codeforpdx/PASS/pull/254

We probably don't want to be implementing the same functionality in multiple places in the code base.

timbot1789 (commented):
I feel like this PR is getting pretty big, and is losing some focus.

Currently it:

- Implements 2 file viewers
- Refactors document loading/parsing
- Adds a new page and route

Would it be possible to limit this PR to just the file viewer? Just create a component that takes a file URL from an already loaded document object and displays the file? We should be able to just drop that into the search document component in the forms page and call it good.

src/components/DocumentModals/ShowDocumentLocalModal.jsx
```diff
@@ -0,0 +1,54 @@
...
+    <DialogContent
+      sx={{
+        display: 'flex',
+        flexWrap: 'wrap',
+        alignItems: 'center',
+        justifyContent: 'center',
+        gap: 5
+      }}
+    >
+      <iframe src={fileBlobUrl} title="File Preview Local" height="100%" width="100%" />
```
timbot1789: nitpick: this kicks rendering of the file to the browser. Probably fine, but not all browsers may be able to render the same file types. And certain ad blockers disable iframes like this.

If we get enough users for this to become an issue though, our app's in a good spot.

src/components/DocumentModals/ShowDocumentsModal.jsx
leekahung: Think that would be good. Could make sure it's fully unique with the url path to the file.

I'll give that a go in place of the uuid.

src/routes/User.jsx
leekahung: I'm not sure what would be a better word for the user's own Pod storage. That's what the page would be for eventually while we move some functionalities over.

I'm thinking along the line of 'Personal Pod' or 'Own Pod', but 'User' is the best I could think of at the moment. I'm up for suggestions.

src/components/DocumentModals/ShowDocumentsModal.jsx
leekahung: Element wise, they should be the same but use different handler functions to show files. Think I would refactor the component out.

src/utils/barcode/barcode-scan.js
leekahung: This is to make sure that the scanner takes in only image files, there seems to have been an issue when the function tried to process non-text files. Think this is likely due to the fact that ZXing was expecting image files to begin with, but it would be something else.

Altering the function to this seems to make the run more "smoothly" when using docType "Driver's License" whether it'll be for image files like pngs or gifs for the barcode or with non-image files like pdfs or txts.

src/utils/frontend/file-viewer-helper.js
leekahung: Think I'll rename this to handleErrorsMessages. The function is used for that primarily based on what is being returned from the getDocTTLs function

src/components/DocumentModals/ShowDocumentsModal.jsx
leekahung: Actually, just checked again. I mistaken this with the handleShowFile function.

The main difference between `ShowDocumentsLocalModal` and `ShowDocumentModal` is that `ShowDocumentLocalModal` shows the document in question as a modal from the `ShowDocumentsModal`. Think I'll rename `ShowDocumentLocalModal` to `ShowDocumentPreviewModal`.

src/utils/frontend/file-viewer-helper.js
leekahung: Actually, I should rename this as `runShowFiles` instead.

src/utils/network/session-core.js
leekahung: I believe only this function fetches for `{ uploadDate, filename, documentType: docType, expireDate, description, documentUrl }` at this time from `document.ttl`? I do find functions that builds and updates `document.ttl` in session-core and session-helper.

Do you mean to refactor the try portion of the codebase out as a helper function?

src/routes/User.jsx
andycwilliams: 'Personal Pod' is more apt in my opinion. Maybe 'Your Pod', 'Your Info', or something along those lines.

leekahung:
> I feel like this PR is getting pretty big, and is losing some focus.
> 
> Currently it:
> 
> * Implements 2 file viewers
> * Refactors document loading/parsing
> * Adds a new page and route
> 
> Would it be possible to limit this PR to just the file viewer? Just create a component that takes a file URL from an already loaded document object and displays the file? We should be able to just drop that into the search document component in the forms page and call it good.

UPDATED: I think what we have now would probably be the extent of things, don't think it'll expand beyond what is mentioned in the bulletpoints.

As for the two viewers, it's not too hefty. One is a modal with an iframe (PREVIEW) and one is a link to a new window (NEW WINDOW). I haven't really thought about ad-blockers and the such potentially blocking iframes, so we'll probably have to fall back on just the one method of new window, so I think removing the PREVIEW option would be good.

Considered using window.open() as a solution, but if we're using window.open(), we could just use that instead of a Link since the file Blob is already extracted from Solid.

As for the refactoring of the Driver's License document loading/parsing, that's more of a bug fix since there is a problem with that. Think we could move that to another PR since the fixes implemented here do resolve that.

For the last bullet point, the functionality being used in two places are necessary since the functionality inside "Clients" would only be accessible to case workers.

As "developers" we have access to "Clients", but normal users wouldn't be able to use that. So a common "User" page has been is created. That way users can check documents in their own Pod. Eventually it'll expand out to incorporate parts of what's in "Forms" right now like file upload or file deletion, but it should be pretty bare at the moment. However, to move the flows into existing ones like search documents mentioned by Tim is a good idea.

I'll be moving towards that for an updated iteration, so I'll be removing "User" until further notice and removing the view file functionality from "Clients" for now.

I think the cross-pod search in "Forms" would ultimately be the form used in "Clients" and search documents in "Forms" would be what's used in "User". What we'll end up calling "User" I'm open to suggestions. Andy have already mentioned "Your Pod" and "Your Info".

src/utils/barcode/barcode-scan.js
leekahung: Relegating these fixes to PR #264.

leekahung:
Got a second iteration up based off of feedback. For the changes, I've:

- Moved bug fixes for file upload related to Driver's License into its own dedicated PR.
- Simplified the modals for file viewer to just ShowDocumentsModal
- Functionality for ShowDocumentsModal utilizes window.open() instead of iframes
- Updated the search functionality from getDocuments to run the functionality of getDocTTLs instead of just returning a url link to the container
- Removed "User" for now until we decide what to move out of "Forms"
- Relegate file viewing to "Forms" temporarily instead of in "Clients" and "User" ("Clients" will likely be using the cross pod search functionality to search for other user's Pods, while "User" will likely just be using the search documents for documents on their own Solid Pod)

src/utils/network/session-core.js
leekahung: Merged getDocTTLs with getDocuments which currently just returns a URL of the container

src/utils/frontend/file-viewer-helper.js
leekahung: Helper function is removed in favor of an updated getDocuments function

src/routes/User.jsx
leekahung: Will consider this when we make the "User" page again; Will temporarily remove this for now

src/components/DocumentModals/ShowDocumentsModal.jsx
leekahung: Simplified Modals into one modal

src/routes/User.jsx
leekahung: Temporarily removing "User" until things are more flushed out

src/routes/User.jsx
leekahung: Removing function in favor of an updated getDocuments function

src/components/DocumentModals/index.js
leekahung: Have updated the directory to just Modals, thinking of having other modals moved into here in the future like LogoutModal, AddClientModal, etc.

src/components/DocumentModals/ShowDocumentsModal.jsx
leekahung: Have just simplified this to just one button/option (see new iteration)

src/routes/User.jsx
xscottxbrownx: > I'm not sure what would be a better word for the user's own Pod storage. That's what the page would be for eventually while we move some functionalities over.

Can you explain what you mean by this? Is the user's own Pod storage anything more than documents?

leekahung: "User" would eventually include upload, delete, fetch, and set permissions to documents. Each user of PASS would be able to access it there, and it'll only be related to their own Pod.

For now, I'm thinking of just leaving the cross Pod functionalities to "Clients" so that only case workers could work on other people's Pods.

Might consider some sort of cross Pod access in the future if the user wants to move things into a new Pod akin to transfer account, but that would be a feature way further down the road or it'll be in someplace else like settings.

xscottxbrownx:
## My 2 cents...

#### This should probably be the last "new functionality" introduced until we all have the same vision/instructions as to how to map out this app from a user's perspective.  

---

### Do we think an individual user (non case-worker/organization) won't need a page to view their connections (people they've given perms to?)  That will only be displayed in their documents section? - which I vote to call "My Documents".    

- From a case-worker/organization standpoint I could see "Clients" and "Client Documents", then "My Documents".
- From an individual user standpoint I could see "Connections" and "Documents" or "My Connections" and "My Documents", unless we don't need a connections/clients page for the individual.

---

#### I honestly planned on building out a basic "Documents"/"Forms"/"My Documents" page similar to "Clients", so it would be easier to envision, but I'm a little lost on what we want/need/are going with now. 

Obviously we can make the UI whatever, doesn't have to be tables, but copying and editing code from Clients can easily get us a "Documents" page with the list of the user's documents and links to view the file itself with this file viewer (similar to what you had, but no need for the modal that contained all the different files - they will already be listed.)  Also with the linked "add document" functionality we already have (button modal like clients), the linked "delete document" functionality we already have (trash icon with confirmation modal like clients), etc.

leekahung:
> ### Do we think an individual user (non case-worker/organization) won't need a page to view their connections (people they've given perms to?) That will only be displayed in their documents section? - which I vote to call "My Documents".
> * From a case-worker/organization standpoint I could see "Clients" and "Client Documents", then "My Documents".
> * From an individual user standpoint I could see "Connections" and "Documents" or "My Connections" and "My Documents", unless we don't need a connections/clients page for the individual.

From an individual's perspective, kind of see two different approaches being used, but I can see it be used in concert.

I think having an address book to like "Clients" for users in the form of "Connections" would be pretty good idea. It'll probably just be a less powerful version of "Clients" that mainly deals with permission setting to documents and tracking people who's in contact with the user's Pod. It'll include information similar to "Clients", so it'll have name and webId like before, but could include organization name, and organization webId. So this acts as a sort of central hub.

An alternative to this, would be to include input fields for who's the uploader of the file would be. So individual users who uploads would just have their webId and name in the uploader fields, while cross-pod uploads would have the webId and name of the person who's performing the cross-pod uploads in the uploader field and potentially their organization name and webId. This is on top of Tim's "verify file" feature which would assign a unique `signature.ttl` to the document container and also points to the original uploader (the "verify file" feature, like "Clients" would likely only be seem by case workers/managers).

From "My Documents" (which I believe is a reference to what I'm temporarily calling "User"), the individual user could see who the original uploader is for that document and set permissions there.

So I could see either a ["Connections", "My Documents", "Messages"] layout for navigation for individual users or ["My Documents", "Messages"]. I'm open to both.

For case workers/managers it'll be ["Clients", "My Documents", "Messages"] for navigation.

src/routes/User.jsx
leekahung: Am considering moving permission setting features to if we create a "Connections" page instead of in "User"/"My Documents" as a possible direction, see https://github.com/codeforpdx/PASS/pull/254#issuecomment-1597524146

src/components/Form/SetAclPermsDocContainerForm.jsx
```diff
@@ -34,7 +34,7 @@ const SetAclPermsDocContainerForm = () => {
   const { session } = useSession();
   const { state, dispatch } = useStatusNotification();
   const { selectedUser } = useContext(SelectedUserContext);
-  const [username, setUsername] = useState(selectedUser.username);
+  const [username, setUsername] = useState('');
```
leekahung, line 37: Using empty string allows the application to reset username when clearing client.

https://github.com/codeforpdx/PASS/assets/14917816/bea20026-4706-4d9f-9722-1196b5d5fd92

timbot1789 (approved):
Not sure why opencommons doesn't allow new tabs to load documents, but this workaround is fine. Should be easy to drop that blob url in an iframe too and preview the document in PASS itself.

andycwilliams (approved):
Looks good to me. I don't have anything to add that hasn't been said already.

PR #256, opened 2023-06-15
Issue 255 - Fixing update/upload new file bug

### LINKED ISSUE #255: Update file/Upload new file bug
See https://github.com/codeforpdx/PASS/pull/249#discussion_r1230327778 for context, but essentially, the update/upload new file functionality has been buggy for a while.

Since we're starting to refactor that functionality in PR #249, it'll be best to make sure this functionality is properly implemented.

leekahung:
Relevant to PR #249 where the full refactor to this functionality will take place. Changes here will ultimately be changed to the version implemented there. The old implementation of upload new file/update file has been buggy for a while, but has not been revisited. Now that we're starting to refactor session-core with uploadDocument being one of the targets, it's a good time to make sure the functionality under the hood is implemented correctly so that it could be used as the starting point for optimizations and subsequent refactors.

This PR fixes that issue by implementing what originally intended for that flow (see [#249 (comment)](https://github.com/codeforpdx/PASS/pull/249#discussion_r1230327778) for context).

First file upload works like before since the container initially starts empty, so `updateDocument` was not called when `uploadDocument` is triggered. A new Thing has been generated for document.ttl at `#document`. This Thing will simply be a reference to itself rather than the document files like with container.ttl. All subsequent Things will be named something else like `#document1` or `#document2`, etc. But a more robust naming scheme would need to be flushed out in case we're going to keep this feature during the refactor in PR #249.

Screen Shot 2023-06-15 at 0 03 30

A second file upload triggers `updateDocument` since container is no longer empty. But unlike when the container is empty, all subsequent calls for `updateDocument` triggers confirmation window which enables the user to cancel if desired.

Assuming the user continues and not cancel the operation, since the file name for the second upload is different, `uploadNewFile` within `updateDocument` is triggered instead of `updateFile`. Compared to `uploadDocument` which generates a new `document.ttl` after uploading a file, `uploadNewFile` instead appends to the existing `document.ttl`.

Screen Shot 2023-06-15 at 0 05 18

Uploading an updated version of the first file will trigger `updateFile` when uploading and the old file will be overwritten by the new one. The Thing related to the first file will also be updated based on what was filled out and a new dateModified field is included to show that it has been updated.

Screen Shot 2023-06-15 at 0 11 02

Screen Shot 2023-06-15 at 0 16 46

Other changes include edits to the message for new file upload and file updates so that the message is now "Uploading new/updated content on Pod..." in `FormSubmissionHelper` and a new parameter for `createResourceTtlFile `so that a unique name could be used for the Thing being generated instead of being hard-coded to 'document'.

### DIFF
--- a/src/utils/frontend/FormSubmissionHelper.js
+++ b/src/utils/frontend/FormSubmissionHelper.js
@@ -48,11 +48,11 @@ const makeHandleFormSubmission =
       clearInputFields(event);
     } catch (e) {
       try {
-        runNotification('Updating on Pod...', 3, state, dispatch);
+        runNotification('Uploading new/updated content on Pod...', 3, state, dispatch);
 
         await updateDocument(session, uploadType, fileObject, crossPodUsername);
 
-        runNotification(`File updated on Pod.`, 5, state, dispatch);
+        runNotification(`File uploaded on Pod.`, 5, state, dispatch);
         clearInputFields(event);
       } catch (error) {
         runNotification(`Operation failed. Reason: ${error.message}`, 5, state, dispatch);
--- a/src/utils/network/session-core.js
+++ b/src/utils/network/session-core.js
@@ -17,7 +17,6 @@ import {
   hasTTLFiles,
   setDocAclForUser,
   createResourceTtlFile,
-  updateTTLFile,
   getUserProfileName,
   saveMessageTTL,
   parseMessageTTL,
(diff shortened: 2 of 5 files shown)


leekahung:
Had been following and reviewing the progress of #249. It seems to be handling the cases touched by this PR. Can close this in favor of #249.

Issue #251, opened 2023-06-12
Update the inactivity message and add force logout on inactivity


timbot1789:
# Update July 7, 2023 [Ka Hung]

We have implemented an extension for the inactivity message from 3 to 30 minutes. If the user is not active after 30, user is likely no longer active, we can forcefully log them out without issue. We just need to implement that as stated in original post.

> 3. Make the required period of inactivity longer (30 mins?) and forcibly logout the user after that time.

We can add an additional delay to warn users of impending logout (i.e. 5 minutes) before actually logging them out. Timing of warning message and forced logout can be discussed.

---

## Original Post

**Is your feature request related to a problem? Please describe.**
Currently, PASS shows an inactivity message after about 3 minutes. This is a rather short time. It's quite common for users to leave to grab a coffee, or just get caught up in a simple conversation and have the message appear. The message itself also doesn't do anything. Normally, these types of messages are used in sensitive systems where you are forced to log out after a period of inactivity. However, this popup does not do that.

This is also going to be annoying and intrusive in live demos, as there will certainly be periods of a few minutes in the demo where we are answering questions for 3 minutes.

**Describe the solution you'd like**
A few possible solutions:
1. Remove the inactivity message
2. Make the message less intrusive (maybe a warning banner at the top of the window?)
3. Make the required period of inactivity longer (30 mins?) and forcibly logout the user after that time.

**Additional context**
It's unclear at this stage what value the inactivity message adds to PASS. It is likely that PASS will be storing sensitive information, and will need to force user logout. However, we're not at that stage yet. There is also value in having a standard method of showing modals. An experiment in this activity message is helpful in that regard. At least we should bump the inactivity timer to ~30 minutes.

xscottxbrownx:
I made a plea for number 3 a while back.  It seemed pointless to not have it automatically log the user out after the message was shown for certain amount of time.

I'm not positive on what the correct timeframes are (how long until message pops up, and how long once it's displayed until user is logged out), but user should be logged out or message doesn't serve any purpose.

leekahung:
I think we can push for number 3 as well and extend the time for the notification.

At longer time scales, I think it's safe to assume that the person is no longer using the app, so forcefully log them out then would be fine. Before with the 3 minute time scale, a reminder would be the better choice.

andycwilliams:
The inactivity timer has been extended to 30 minutes as of #260

andycwilliams:
[...]

PR #260, opened 2023-06-18
Issue 207/inactivity message styling updates

### LINKED ISSUE #207: Refactor InactivityMessage to MUI
The `InactivityMessage.jsx` component currently uses custom CSS. In order to maintain consistency, as well as reduce redundancy, it ought to be updated with the MUI styling chosen for this project.

Refactor the existing code in keeping with the existing style (i.e. other components, utilizing the `theme.js` stylesheet, and updating it as necessary).

andycwilliams:
Updating styling of `InactivityMessage.jsx` to MUI. The dialog box more closely matches the styling of other components (such as `LogoutModal.jsx`), while leaving specifically the button styling largely the same. That is, they are different variants but both colored green.

New styling:

Previous styling:

`LogoutModal.jsx`:

TODO: Determine the styling we want for modals/dialog boxes. For example, several other components have only `contained` buttons, others have only `outlined` ones (like `LogoutModal.jsx`), and still others (like this one and `AddClient.jsx` ) have both.

Is this something we want to keep, do we wish to have a single theme to apply universally, or something else?

Additionally, we will likely want to recolor the `Log Out` button to `error` coloring rather than `success`. Buttons that log out the user should be more easily distinguishable.

Creating as a draft until further discussion.

Closes #207


leekahung:
Could we up the timing for the pop-up to be 30 minutes?

We could think about whether to make that trigger logout after that could be handled in a different PR.

leekahung:
Oh, one more thing to note before I dig into the code. Could we use what we have for red for log out.

It'll be the better choice, like using red for 'no' or 'cancel'. And swap the logout icon for what we're using in the navbar for logout that way it'll be consistent.

xscottxbrownx:
## Thanks for making this a draft an opening for discussion.

### I also believe there should be consistency between all the modals.  My thought process was whichever button would continue the userflow - would be green, and the button that canceled the userflow - would be red.

So, in the case of `LogoutModal`, the modal shows when a user clicks "logout" and is likely they want to logout. Therefore, I made "yes" green and "no" red.  
You'll see I did the same for `AddClientModal` (even though those buttons don't work as intended yet.)

---

### Now in the case of `InactivityMessage`, this logic doesn't really work as the user didn't do anything to get to it.
So, 2 thoughts come to mind:
- We could say by 30 mins of inaction, the user *is* saying they no longer want to use PASS (and most likely *want* to logout) - therefore `continue` is red, and `logout` is green.
- Since the user did nothing to get to this option, we could say they don't want to logout - therefore `logout` is red (and I propose listing it on the left), and `continue` is green (and I propose listing it on the right for consistency with others.)

---

### Specific text of this modal
Personally I don't really like the `continue` coupled with the specific wording of the question.  It asks if you would like to log out - so "continue" can also mean "continue to logout."  **BUT** the text button could be answering the modal title question of `Continue session?`    
### Again, I think we need to make this modal clearer - we are asking two opposing questions in the same popup:
 1) Do you want to continue your session?
 2) Do you want to logout?
 
### So, maybe we pick one or the other:
 1) **Modal Title:** Continue session?
 **Modal Body:** You've been inactive for 30 minutes. Would you like to continue using PASS? (By clicking "No" or not answering within 5 minutes, you will be logged out.)
 2) **Modal Title:** Logout?
 **Modal Body:** You've been inactive for 30 minutes. Would you like to logout from PASS? (By clicking "Yes" or not answering within 5 minutes, you will be logged out.)

xscottxbrownx:
> TODO: Determine the styling we want for modals/dialog boxes. For example, several other components have only `contained` buttons, others have only `outlined` ones (like `LogoutModal.jsx`), and still others (like this one and `AddClient.jsx` ) have both.
> 
> Is this something we want to keep, do we wish to have a single theme to apply universally, or something else?

For `LogoutModal` and `AddClientModal` I used outlined buttons, but I do like the thought of mixed in these components. 

My thought is what would BREAK the userflow (what button choice the user most likely DOES NOT want to choose) to be `outlined`, and what would CONTINUE the userflow (what button choice the user most likely DOES want to choose) to be `contained`.

To move forward with this thought ⬆️, we have to make the modal text/messaging more clear first.

---

**Edit:**  I actually really like this idea of mixed buttons - outline for break userflow option, and contained for continue userflow option.  Break userflow option on left, continue userflow option on right.

I also just noticed it's what gitHub does as well:
Screenshot 2023-06-18 at 7 12 30 AM

xscottxbrownx:
> Oh, one more thing to note before I dig into the code. Could we use what we have for red for log out.
> 
> It'll be the better choice, like using red for 'no' or 'cancel'. And swap the logout icon for what we're using in the navbar for logout that way it'll be consistent.

If we go with this (coloring `logout` red), then I propose we change the question in the body of the modal to "Would you like to continue using PASS?" and swapping the button order.

---

If we change the icon, is this the exception (meaning logout is always special and gets its icon?)
Or do we need to change icons on all buttons that make sense (like "add client" could be the + icon?)

My current vote on this is **no**, keep it universal:
- Red with "x" for things that break the userflow (most likely what user DOES NOT want to do)
- Green with "✓" for things that continue the userflow (most likely what user DOES want to do.)

xscottxbrownx:
Does this PR also address Issue #251 as well?   Or should we consider that something separate with an automatic logout feature if the user doesn't respond to this modal in certain amount of time?

leekahung:
> Does this PR also address Issue #251 as well?   Or should we consider that something separate with an automatic logout feature if the user doesn't respond to this modal in certain amount of time?

I think we should consider the auto logout functionality from inactivity as a separate issue, but yes, this should be related to issue #251.

At 30 minutes, I think a shorter 2 minute count down to trigger logout might would be better, but I'm good with 5 as well.

leekahung:
> > Oh, one more thing to note before I dig into the code. Could we use what we have for red for log out.
> > 
> > It'll be the better choice, like using red for 'no' or 'cancel'. And swap the logout icon for what we're using in the navbar for logout. That way, it'll be consistent.
> 
> If we go with this (coloring `logout` red), then I propose we change the question in the body of the modal to "Would you like to continue using PASS?" and swapping the button order.

Yes, I think swapping the ordering would be good. That way, we could keep the ordering for exiting userflow on the left and continuing on the right and change the phasing as you mentioned.

> If we change the icon, is this the exception (meaning logout is always special and gets its icon?)
> Or do we need to change icons on all buttons that make sense (like "add client" could be the + icon?)

Yeah, thinking of making logout a special case, everything else could remain x's and checks.

leekahung:
> > > Oh, one more thing to note before I dig into the code. Could we use what we have for red for log out.
> > > 
> > > It'll be the better choice, like using red for 'no' or 'cancel'. And swap the logout icon for what we're using in the navbar for logout. That way, it'll be consistent.
> > 
> > If we go with this (coloring `logout` red), then I propose we change the question in the body of the modal to "Would you like to continue using PASS?" and swapping the button order.
> 
> Yes, I think swapping the ordering would be good. That way, we could keep the ordering and phrasing for exiting userflow on the left and continuing on the right.
> 
> > If we change the icon, is this the exception (meaning logout is always special and gets its icon?)
> > Or do we need to change icons on all buttons that make sense (like "add client" could be the + icon?)
> 
> Yeah, thinking of making logout a special case, everything else could remain x's and checks.

andycwilliams:
Here's the latest styling: 


Going to keep this as a draft until we have a theme for this modal firmly decided.

xscottxbrownx:
> Going to keep this as a draft until we have a theme for this modal firmly decided.

I'll look at the code today, but I like this... I would just suggest changing the green button to the same green we've been using. That seems to be something different.    After that, from a "looks" perspective I think this is solid.

src/components/Notification/InactivityMessage.jsx
```diff
@@ -2,10 +2,15 @@
...
+import Dialog from '@mui/material/Dialog';
+import DialogActions from '@mui/material/DialogActions';
+import DialogContent from '@mui/material/DialogContent';
+import DialogContentText from '@mui/material/DialogContentText';
+import DialogTitle from '@mui/material/DialogTitle';
+import CheckIcon from '@mui/icons-material/Check';
+import LogoutIcon from '@mui/icons-material/Logout';
 
 /**
  * Inactivity Notification Component - Component that displays a popup modal
```
xscottxbrownx, line 16: change jsdoc to also say "30" now instead of "3"

src/components/Notification/InactivityMessage.jsx
```diff
@@ -67,76 +72,44 @@ const InactivityMessage = () => {
...
+              color="error"
+              endIcon={<LogoutIcon />}
+              onClick={handleLogout}
+            >
+              Log Out
+            </Button>
+          </LogoutButton>
+          <Button
+            variant="contained"
+            color="success"
```
xscottxbrownx: we should probably change this to match all the other modal/green colors.  Or we could change the rest to match this I guess.

xscottxbrownx (commented):
Think code looks good too. Just a small change to the jsdoc comments, since the timing change.   

I'd approve after these changes, or at least a discussion if not changing.  Basically, as long as team addresses these two things I think this is ready to merge.

src/components/Notification/InactivityMessage.jsx
```diff
@@ -67,76 +72,44 @@ const InactivityMessage = () => {
...
+        aria-labelledby="inactivity-message-title"
+        aria-describedby="inactivity-message-description"
+      >
+        <DialogTitle id="inactivity-message-title">Continue session?</DialogTitle>
+        <DialogContent id="inactivity-message-description">
+          <DialogContentText>
+            You have been inactive for a while now. Would you like to continue using PASS?
+          </DialogContentText>
+        </DialogContent>
+        {/* TODO: In future PR, add countdown timer to automatically log user out if they do not select continue */}
```
leekahung: Could probably move this comment above the handleLogout function which is being triggered.

leekahung (commented):
I agree with Scott. Think both the code and visuals looks good. Think it should be ready for merging. Just a small note on the placement of the TODO comment, but that's just nitpick.

leekahung (approved):
Alright, think that should cover all the bases here. Since PR #261 was merged as well, remember to update the branch first before the merge.

PR #261, opened 2023-06-18
237/delete env

timbot1789:
Remove .env file from repository. Keeping that around can mess with different deployment environments in hard to predict ways.

NOTE: ACTION MUST BE TAKEN TO MAINTAIN CURRENT BEHAVIOR

1. developers will need to add a `.env` file to their root directory. They can copy the `dev.env` file found in the `env_templates` directory to do so
2. The environment variables `VITE_SOLID_IDENTITY_PROVIDER_PRODUCTION` and `VITE_SOLID_IDENTITY_PROVIDER_DEV` have been condensed into `VITE_SOLID_IDENTITY_PROVIDER`. Environments (code for PDX pages) need to be updated accordingly.


leekahung (commented):
Think this looks good. Tested the branch out myself and it seems to work just fine. I believe this is ready for merging?

Update: Have just updated Environment to use VITE_SOLID_IDENTITY_PROVIDER

leekahung (approved):

xscottxbrownx:
I would love it if it was set to "solidcommunity.net" so I don't have to copy that into the login component anymore. 🤣

xscottxbrownx:
I'm a little confused on what we are doing here...

Is the goal just to remove the `.env` file from GitHub?    Seemingly so, since directions are for devs to create a `.env` file.

Couldn't we just delete the file here and add `.env` to the `.gitignore` ?    Is that what this PR effectively is doing?

leekahung:
> Couldn't we just delete the file here and add `.env` to the `.gitignore` ? Is that what this PR effectively is doing?

It's because we have an `.env` file on GitHub, GitHub continues to track it for changes. So having this branch remove the `.env` and having developers copy a new one locally from `env_templates/dev.env` will let `.gitignore` kick in.

timbot1789:
> I would love it if it was set to "solidcommunity.net" so I don't have to copy that into the login component anymore. rofl

This will happen after merge if you copy `templates/dev.env` to `.env`. Part of the reason I made this fix is because you were having that issue :)

PR #262, opened 2023-06-18
Remove user activity

timbot1789:


xscottxbrownx:
### What was the reasoning for this whole thing again?

Was it just no real utility to show this in the table?   

Would there be any other reason now or in future to need to know this "last activity" time, even if not to display to user but to keep the functionality for other dev purposes?

timbot1789:
> ### What was the reasoning for this whole thing again?
> Was it just no real utility to show this in the table?
> 
> Would there be any other reason now or in future to need to know this "last activity" time, even if not to display to user but to keep the functionality for other dev purposes?

No immediate utility in the table, and a pretty substantial performance degradation. Probably what we need instead here is a Client Profile stored on the client's pod that we fetch, but that's not really what this data point does.

xscottxbrownx:
I don't see any issues.  I would like @leekahung 's eyes on the modal-helpers files though.

leekahung (approved):
I think it should be fine.

We're going to move over the storing of the last active date inside the user's profile card. Could get the time from there once that completes.

leekahung:
Think we can merge this after the branch updates. @timbot1789 

Let me know if you think it's ready. 👍

PR #264, opened 2023-06-19
Issue-263/Fix bug license upload

leekahung:
Mostly a file upload bug fix for docType "Driver's License".

Scott has noticed how user's couldn't upload files that are not image files when testing out the file viewer. Had fixed the underlying function by adding conditions for when the function is triggered. If the file is an image file, it'll process the files like before and give a license specific document.ttl. Otherwise, it'll give out the normal one like with the other docTypes.

(See fixes)

https://github.com/codeforpdx/PASS/assets/14917816/69bc109a-6f85-43b4-9dfd-5c0cfa8c4971

A secondary feature of this would be a fallback for when the image fails to process. Perhaps the image file is corrupted or the image is poorly taken, it'll switch back to the same document.ttl file for now using the input fields from "Upload Documents". We could create a more Driver's License specific form for upload to produce the specialized document.ttl in case of failure to process, but that could be done in another issue/PR.


xscottxbrownx:
Just to make sure...

Is this the functionality we want?  Do we want anything other than images?

leekahung:
> Just to make sure...
> 
> Is this the functionality we want? Do we want anything other than images?

Yeah, perhaps there's a document related to a Driver's License like an application file someone wish to upload (i.e., pdf). But that's just for now since we only have three categories. I could see it expand now that we starting to implement multi-file uploads.

A secondary feature of this is a fallback document.ttl file in case the scanner fails to run. For now we'll be using the ones we use for other docTypes. In the future, we might want to use something more relevant to driver's license.

leekahung:
I believe this issue was resolved in PR #249. Closing this in favor of that PR.

PR #265, opened 2023-06-20
Issue 217/consolidate inbox outbox

### LINKED ISSUE #217: Consolidate Inbox and Outbox Pages into Messages
At the moment, Inbox and Outbox are independent routes to PASS, but can be consolidated to just one page called Messages.

Consolidate Inbox and Outbox pages into Messages. Users would then be able to toggle between inbox and outbox components without routing using some sort of button or internal navigation within Messages.

leekahung:
This PR is mainly about consolidating `Inbox` and `Outbox` into a single `Messages` Page. With this, `Inbox` and `Outbox` would be re-routing to `Messages` with and `Inbox` and `Outbox` being used as toggles for displaying their respective components as oppose to sending the user to another page.

https://github.com/codeforpdx/PASS/assets/14917816/e7487a20-98b3-4b7c-bb6b-52f6ed5ee7b6

Considering how similar the two message component are they're now refactored as the new `MessageFolder` component, which changes based on `folderType`. For an easy transition to Material UI from styled-components, I've moved all relevant stylings into `MessageStyles.js`. Have moved a few styled elements into MUI components. The "New Message" button moved outside of `Inbox`.

Included minor edits to Layouts and other components so that it could fit within 1 x 3 grid. Should have minimal visual effect on what we have currently, but makes styling for the children component within Layout a lot easier.

https://github.com/codeforpdx/PASS/assets/14917816/617d965c-a179-4d14-8030-b699fa3fa3ed


src/components/Messages/MessageStyles.js
```diff
@@ -0,0 +1,156 @@
+import styled from 'styled-components';
```
xscottxbrownx, line 1: Just to be clear...

are you making more styled components for this feature, or only moving the pre-existing styled components out from all the separate components and into their own file (`MessageStyles.js`)?

src/components/Messages/Pagination.jsx
```diff
@@ -32,7 +47,8 @@ const PaginatedMessages = ({ messages }) => {
 
   return (
     <>
-      <Messages currentMessages={currentMessages} />
+      {currentMessages &&
+        currentMessages.map((message) => <MessagePreview key={uuidv4()} message={message} />)}
```
xscottxbrownx: Not an issue for 2 characters, but COULD remove the parentheses around `message` in the map function.

src/components/NavBar/NavbarLinks.jsx
```diff
@@ -1,5 +1,5 @@
 // React Imports
-import { React, useState } from 'react';
+import React, { useState } from 'react';
```
xscottxbrownx, line 2: Not sure if I've asked this yet, but why does this codebase/project seem to require `import React from 'react'`?

This hasn't been needed since v17.   

Is there something about vite or the linting process for this project?

src/routes/Messages.jsx
```diff
@@ -0,0 +1,104 @@
...
+    messagesInSolid.sort((a, b) => b.uploadDate - a.uploadDate);
+    if (folderType === 'Inbox') {
+      setInboxList(messagesInSolid);
+    } else {
+      setOutboxList(messagesInSolid);
+    }
+    setLoadMessages(false);
+  };
+
+  // Re-sorts inbox messages upon updates
```
xscottxbrownx, line 54: Can we refactor these two separate `useEffect`s into one maybe?  Not sure if this is correct, so you'd have to check, but something like this:

```js
// Re-sorts appropriate messages upon updates
useEffect(() => {
  setLoadMessages(true);
  
  // Re-sort inbox messages if inboxList changes
  if (inboxList) {
    let inboxCopy = [...inboxList];
    inboxCopy.sort((a, b) => b.uploadDate - a.uploadDate);
    setInboxList(inboxCopy);
  }

  // Re-sort outbox messages if outboxList changes
  if (outboxList) {
    let outboxCopy = [...outboxList];
    outboxCopy.sort((a, b) => b.uploadDate - a.uploadDate);
    setOutboxList(outboxCopy);
  }

  setLoadMessages(false);
}, [inboxList, outboxList]);
```

src/routes/Messages.jsx
```diff
@@ -0,0 +1,104 @@
...
+    setOutboxList(outboxCopy);
+    setLoadMessages(false);
+  }, [outboxList]);
+
+  const [boxType, setBoxType] = useState('inbox');
+
+  return (
+    <>
+      <Tabs value={boxType} sx={{ padding: '10px 30px' }}>
+        {routesArray.map((item) => (
```
xscottxbrownx: Not necessary, but can remove parentheses around `item` in map.

src/routes/Messages.jsx
```diff
@@ -0,0 +1,104 @@
...
+          <Tab
+            key={`${item.label}Tab`}
+            value={item.label.toLowerCase()}
+            label={item.label}
+            onClick={() => setBoxType(item.label.toLowerCase())}
+          />
+        ))}
+      </Tabs>
+
+      {boxType === 'inbox' ? (
```
xscottxbrownx: Maybe refactor this section to:

```js
        <MessageFolder
          folderType={boxType === 'inbox' ? "Inbox" : "Outbox"}
          handleRefresh={handleMessageRefresh}
          loadMessages={loadMessages}
          messageList={boxType === 'inbox' ? inboxList : outboxList}
        />
 ```

File: src/utils/network/session-core.js [hunk 7]
```diff
@@ -169,16 +169,16 @@ export const getMessageTTL = async (session, boxType, listMessages, podUrl) => {
 };
 
 /**
- * Function that sends a message to another user's Pod inbox and saves a copy in
- * user's own inbox
+ * Function that sends a message to another user's Pod inbox as a TTL file and
```
xscottxbrownx, line 172: cut out one `and`

src/components/Messages/MessageFolder.jsx
```diff
@@ -0,0 +1,60 @@
...
+ * @memberof Messages
+ * @name MessageFolder
+ * @param {messageFolderProps} Props - Component props for MessageFolder
+ */
+
+const MessageFolder = ({ folderType, handleRefresh, loadMessages, messageList }) => {
+  const [showForm, setShowForm] = useState(false);
+
+  return (
+    <section
```
xscottxbrownx: Maybe change `div`s to `<Box>` with `sx` prop to replace `style`.  
And also `section` to `<Box component="section">` with `sx` prop to replace `style`.

Just to cut down on how much refactoring there is to MUI in future.

xscottxbrownx (changes requested):
Last thought:

I'm not sure having the `New Message` button inside `Inbox` is the most logical - but that wasn't made in this PR and can definitely stay.  But, I feel that user should be able to make a new message from anywhere in `Messages` route/page.

I think I'd prefer `Messages` to be split more.  Like instead of tabs, having a sidebar of sorts with:
1. the `SendOutlined` icon and Create Message
2. the `MarkunreadOutline` icon and Inbox
3. the `ForwardToInboxOutline` icon and Sent (renamed from Outbox)

This would be the navigation of `Messages`, all clickable to render the rest of screen. Basically a new layout of messages with these 3 things as an aside to the main.

xscottxbrownx:
Since you are already redoing Navbar Links to have `Messages`, want to rename `Forms` to `Documents` for now?

src/routes/Messages.jsx
timbot1789: I actually disagree. I like having them separate.

I'm not sure that line `if (inboxList)` checks for a list of changes. I'm pretty sure `inboxList` is just the whole array. In which case the inbox sort will always run no matter what. Would be happy to be corrected on that though.

Either way, keeping the hooks separate means that React handles what changed and we don't have to think about it, so we don't need those `if` statements.

xscottxbrownx: You are correct on both parts: A) the `if` condition is not correct. B) probably *is* better separate.

src/components/Messages/MessagePreview.jsx
```diff
@@ -1,14 +1,25 @@
...
+ */
 
 /**
  * Message Preview Component - Component that displays message previews from
  * user's Inbox container
  *
  * @memberof Inbox
  * @name MessagePreview
+ * @param {messageFolderProps} Props - Component props for MessagePreview
  */
```
timbot1789: I don't know if it's just my setup, but these blank between the JSDoc description and the function prevent my JSDoc validation from running. Can we remove them?

src/components/NavBar/NavbarLinks.jsx
timbot1789: Personally, I wasn't aware this was updated, so that's good to know.

However, at least my linter throws errors when I don't import it.

I'll see if I can figure out whatever's wrong there.

src/components/Messages/MessageStyles.js
leekahung: They're just moved to a separate file until they're phased out with Material UI, kind of like the classes in styles.css.

src/components/NavBar/NavbarLinks.jsx
leekahung: For this, I'm basically following the linter. It seems to complain when you try to make .jsx files without it. See below:

Screen Shot 2023-06-20 at 11 39 02

src/routes/Messages.jsx
leekahung: I'll change it to this

timbot1789 (commented):
Overall, I think this is a good UI refactor. I think if you just respond to my and Scott's comments I have no issue approving.

However, the more I look at messages, the less it looks like email and the more it looks like a chat app where you have basically one conversation with each person. IMO it would be good to touch base with UI/UX about which system would be more useful.

A simple conversational chat system where we have one single conversation with each individual would be much easier to implement. Then we wouldn't need inbox or outbox tabs, just the chat with each client. We wouldn't need a new message field either. Just a text box at the bottom of the convo and a button that says 'send'. We could also integrate that with an instant messaging service and send text messages to a client if they have a phone. Twilio provides this service, although they're pricey.

A full email system would be more robust, but will be much harder to write properly. Email has a lot of weird features.

src/components/Messages/MessageFolder.jsx
leekahung: Got this moved to using Boxes from MUI.

src/components/Messages/MessagePreview.jsx
leekahung: I can do that.

andycwilliams:
> However, the more I look at messages, the less it looks like email and the more it looks like a chat app where you have basically one conversation with each person. IMO it would be good to touch base with UI/UX about which system would be more useful.

I'd be really interested to hear what the caseworkers think about this. Maybe they really don't need anything more robust than a chat system. Although if we do opt for a more standard email system, we don't necessarily have to include every feature.

src/components/Messages/Pagination.jsx
leekahung: Prettier seems to insist the parentheses in the map. It reappears after running Prettier.

src/routes/Messages.jsx
leekahung: Similar to the earlier case with map and Prettier. Prettier includes the parentheses during formatting even after removing them.

src/utils/network/session-core.js
leekahung: Got this changed to `and`

src/layouts/Layout.jsx
leekahung, line 1: Included minor edits to Layouts and other components so that it could fit within 1 x 3 grid. Should have minimal visual effect on what we have currently, but makes styling for the children component within Layout a lot easier with the 1fr.

https://github.com/codeforpdx/PASS/assets/14917816/617d965c-a179-4d14-8030-b699fa3fa3ed

timbot1789:
> I'd be really interested to hear what the caseworkers think about this. Maybe they really don't need anything more robust than a chat system. Although if we do opt for a more standard email system, we don't necessarily have to include every feature.

Building out a fully featured email system is going to be a really big ask. Months to years of work. Easiest way would probably be to bolt on an email server onto the solid server and just use email as is. But that's not a small ask either. I'd be hesitant to sell that as a core feature if we can avoid it.

src/components/Messages/Pagination.jsx
xscottxbrownx: Alright, seems a weird setup and coupled with the must use `import React from 'react'` in linting.   I use both prettier and eslint on my personal projects and neither of these occur 🤷🏻‍♂️

It's not a problem, just something I was meaning to ask for a while - @timbot1789 (you responded somewhere about this too.)

src/components/Clients/ClientListTableRow.jsx
```diff
@@ -95,7 +95,7 @@ const ClientListTableRow = ({ labelId, client, state, dispatch }) => {
           rel="noreferrer"
           style={{ textDecoration: 'none', color: theme.palette.primary.dark }}
         >
-          {client.webId}
+          Link to Pod Profile
```
xscottxbrownx, line 98: Fine with this for now, but I'd prefer the name is just a link to this - not another whole column like now.  But, that's for another PR.

xscottxbrownx (approved):

src/components/Clients/ClientListTableRow.jsx
andycwilliams: Either way, we may want to add an "open in new tab" icon so the user knows it'll exit the app.

e.g.
`import OpenInNewIcon from '@mui/icons-material/OpenInNew';`

leekahung:
Alright, I think I've touched most if not all the comments in this PR. Think it's ready for merging.

PR #270, opened 2023-06-20
Issue 267/loading documents spinner

### LINKED ISSUE #267: Create a "loading documents" component
## Issue:
Currently, there is no indication that the table of documents is loading for the user

---

## Proposed solution: 
1. Use a similar loader to the `Clients` for consistency for now

xscottxbrownx:
# This PR closes #267 

This PR:
1. Pulls the "loading clients" feature out of `Clients` and into it's own component `Loading` in the `Notifications` folder.
2. Imports that into `Clients` to display when the client table is loading
3. Imports that into `Forms` to display it when the document table is loading

---

### To successfully use `Loading`:

Make sure to use the `loadingItem` prop with a string.   i.e. `<Loading loadingItem="clients" />`

### DIFF
--- a/src/components/Documents/DocumentTable.jsx
+++ b/src/components/Documents/DocumentTable.jsx
@@ -1,15 +1,17 @@
 // React Imports
 import React, { useContext } from 'react';
-// Inrupt Library Imports
+// Material UI Imports
 import Table from '@mui/material/Table';
 import TableBody from '@mui/material/TableBody';
 import TableContainer from '@mui/material/TableContainer';
 import TableHead from '@mui/material/TableHead';
 import TableRow from '@mui/material/TableRow';
 import Paper from '@mui/material/Paper';
+// Component Imports
 import { DocumentListContext } from '../../contexts';
 import { StyledTableCell } from '../Table/TableStyles';
 import DocumentTableRow from './DocumentTableRow';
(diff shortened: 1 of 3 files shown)


xscottxbrownx:
I now finally see how I requested PRs into `Master` instead of `Development` all those previous times.

I always change it to `Development` branch first, but when I finally type everything up and go to the bottom, the button says "Draft PR".  When I switch to "PR" to create an actual real PR, it changes the branch to `Master`.

andycwilliams:
Seems like a simple enough PR. I like it.

The one thing is I'd maybe rename the file from `Loading` (and therefore the import) to something more specific, like `LoadingAnimation`. That way its purpose is clearer.

Also, while looking through more MUI options, I found what they call a skeleton loading animation:
https://mui.com/material-ui/react-skeleton/

The linear animation looks good and I have no problem keeping that. I just wonder if a skeleton would be better suited to large tables of data. It's also customizable, so we can change the size and specifics if we want to (e.g. have it use one animation per row of existing data).

I think it would suit both clients and documents, so there wouldn't be a need to create a separate animation.

timbot1789 (approved):

leekahung (approved):
Yeah, this looks good to me. I approve too.

PR #272, opened 2023-06-21
Issue 218/addclient styling update

### LINKED ISSUE #218: Update the AddClient and AddClientModal so buttons work as intended & styled better
### Issue:

**Currently `AddClient` lives inside the `AddClientModal` and the `AddClientModal` buttons do not do anything.**

---

### Direction:

- Most likely fully refactor `AddClient` into `AddClientModal` and delete `AddClient`
- Link the modal button to the proper function to add the client
- Style better using MUI

Screenshot 2023-06-06 at 12 52 19 PM

andycwilliams:
Opening as a draft until a styling theme is determined.

This PR refactors `AddClient.jsx` and `AddClientModal.jsx` into a single component. Buttons and styling are now consolidated and the form now looks cleaner.

However, I want to discuss styling (and ensure that it is fully functional) before submitting it for review. Some of the choices I have made are experimental, to try things out and see how much styling is redundant.

To test, I have changed the `Paper` in `FormSection.jsx` to have an elevation of 0 because otherwise it conflicts with the way modals are styled by default.

However, as a result, the other forms on the Documents page have no border. If I'm not mistaken, those will also be refactored out of this page and put into modals. If we want to, we could add a border back into them individually if we so desire.

I've also commented out `onClick={() => setShowModal(false)}` on the `Add Client` button as a stopgap measure, as it was closing the dialog box instantly. Haven't had as much time to work on that.

These are just some ideas. Let me know what direction you think we should go in.

Closes #218


xscottxbrownx:
### Ok, my thoughts...

#### As is, this is a great update. 

Buttons now function as you would expect. 
Fields all work and the required validation works as expected. 
Styling matches what other modals/themes we have going on.

---

I'm not too concerned with the `<Paper>` styling for the forms/documents - as I feel those will all turn into modals linked to a button or icon of some sort.

---

I personally feel like the modal should close after clicking "Add Client", BUT the status messaging that is currently in the Add Client modal should be shown to user still. Requiring either another component or something, OR putting that specific status messaging on the table itself. 

---

All in all, I'd be interested in moving forward with this the way it is now honestly.  But would prefer a quick discussion with others on how to handle modal closing/status messaging.  I think a solid direction for that would make this perfect.

leekahung:
We could drop some MUI components to simplify the styling. For example, `Container` and `Paper` in `FormSection` can be dropped and the properties we want to keep like midWidths could be moved to the remaining `Box`. We can also use `boxShadow` in `Box` to mimic the elevation from `Paper`.

https://github.com/codeforpdx/PASS/assets/14917816/b5cf4921-737c-4744-b9e5-58048be9c348

Of course, we'll have to change the `gap` of the upper most `Box` in `Documents`, but that should be doable.

andycwilliams:
Just a quick update. Want to make sure everything is on the right track. Can't take a screenshot at the moment.

Adding a shadow to `Box` in `FormSection` does appear in the `AddClientModal`, at least the way I've implemented it.

leekahung:
Yeah, everything looks like it's headed in the right direction from the visuals.

andycwilliams:
If we're good on merging, I can do that now

xscottxbrownx:
> I personally feel like the modal should close after clicking "Add Client", BUT the status messaging that is currently in the Add Client modal should be shown to user still. Requiring either another component or something, OR putting that specific status messaging on the table itself.

I'd prefer this discussion to hash out and have a direction here, **BUT good with merging as is because it's only better from what it was.** The modal not auto-closing was already the case, so this isn't a backwards step or anything!

xscottxbrownx (approved):
> I personally feel like the modal should close after clicking "Add Client", BUT the status messaging that is currently in the Add Client modal should be shown to user still. Requiring either another component or something, OR putting that specific status messaging on the table itself.

I'd prefer this discussion to hash out and have a direction here, **BUT good with merging as is because it's only better from what it was.** The modal not auto-closing was already the case, so this isn't a backwards step or anything!

andycwilliams:
There don't seem to be any issues with this PR, so I will merge it. In a future one, we should address Scott's concerns.

PR #274, opened 2023-06-22
Issue 271/pass profile page

### LINKED ISSUE #271: Create Profile Page to view/edit Pod profile card
Currently, users have to go to their webIds to view the data related to their user profile for their Pod. However, even then it's not exactly readable for the average user.

We can use what we have in the Navbar for Profile to navigate us to "/PASS/profile".

It'll only require minor tweaks to `NavMenu` and `NavbarLinks`, a new route to AppRoutes for "/PASS/profile" and a new `Profile` page. To test out this new feature, I plan to make this simple, like viewing and updating the user's name. We can further expand from this to include other fields like organization in future issues/PRs as well as for fetching profile information from webId when loading into the page.

https://github.com/codeforpdx/PASS/assets/14917816/0d7dcf4f-6a01-4775-9e95-f4222fd29a47

leekahung:
# PR Details

This PR is about creating a new `Profile` Page for users to view and edit their personal information on their webId's profile card without needing to see or touch the TTL file directly.

The page could be accessed using the existing "Profile" link on the top right right corner and would have the route of "/PASS/profile" instead of the profile card itself.

https://github.com/codeforpdx/PASS/assets/14917816/03858b55-33b9-47a8-9ba9-54aa4bbcd44a

Note that a user doesn't come with a name for some servers because Solid does not mandate a name in its protocol, just the webId. But this will allow us to create a profile card follows some of the recommendations set in the following spec: https://github.com/solid/solid-spec/blob/master/solid-webid-profiles.md#minimum-recommended-profile-information

At the minimum, we're going to have the `foaf:name`, `foaf:nick`, and `foaf:img` fields included for name, nickname, and profile picture respectively. For this PR, we have left out `pim:storage` and `cert:key`.

Having `foaf:name` would at the very least allow some utility for when functions searches for the user's name from a profile card. If it's blank or null because the Solid provider didn't provide an `foaf:name` initially (see image below), you can now update that field.

Screen Shot 2023-06-29 at 13 19 57

While the specification mentioned we may fall back on webId if `foaf:name` is missing, I think it'll be better to have this filled.

Going back to `Profile`, there is still the full webId if users still wish to see it. Note, this profile card is public, so any consideration of additional fields in here must take that in mind.

## Future Consideration

Considering building a private card (i.e. `private_card.ttl` in `/PASS`) to store sensitive information in the future. The forms in Profile could then be separated into "Public" and "Private", both listed under "Profile Information", but "Public" will be from `/profile/card`, while "Private" will be from `/PASS/private_card.ttl`


src/components/Profile/ProfileInputField.jsx
```diff
@@ -0,0 +1,70 @@
...
+        <>
+          <Typography>{inputName}: </Typography>
+          <Input value={inputValue} onChange={(e) => setInputValue(e.target.value)} />
+        </>
+      ) : (
+        <Typography>
+          {inputName}: {inputValue}
+        </Typography>
+      )}
+      <Button
```
xscottxbrownx: Should we make the `Cancel` the red version like we've been doing in modals? And/Or putting in the "X" icon?

Should we make the `Update` the green version like we've been doing in modals? And/Or putting in the "checkmark" icon?

---

I'm not sure either would be necessary, just a thought.

src/components/Profile/ProfileInputField.jsx
```diff
@@ -0,0 +1,70 @@
+// React Imports
+import React from 'react';
+// Material UI Imports
+import Box from '@mui/material/Box';
+import Button from '@mui/material/Button';
+import Input from '@mui/material/Input';
+import Typography from '@mui/material/Typography';
+
+/**
+ * messageFolderProps is an object that stores the props for the MessageFolder
```
xscottxbrownx: need to update this comment to name this prop/component

src/components/Profile/ProfileInputField.jsx
```diff
@@ -0,0 +1,70 @@
...
+ * @property {boolean} editInputValue - Boolean for editing input value
+ * @property {(value: React.SetStateAction<null>) => void} setInputValue - Set
+ * function for inputValue
+ * @property {() => void} handleEditInput - Handler function for editing input
+ * @property {() => void} handleCancelEdit - Handler function for cancelling edit
+ * @memberof typedefs
+ */
+
+/**
+ * New Message Component - Component that allows user to write
```
xscottxbrownx: This one too

src/components/Profile/ProfileInputField.jsx
```diff
@@ -0,0 +1,70 @@
...
+  setInputValue,
+  handleEditInput,
+  handleCancelEdit
+}) => (
+  <Box sx={{ display: 'flex', alignItems: 'center', gap: '10px' }}>
+    <Box sx={{ display: 'flex', alignItems: 'center', gap: '10px' }}>
+      {editInputValue ? (
+        <>
+          <Typography>{inputName}: </Typography>
+          <Input value={inputValue} onChange={(e) => setInputValue(e.target.value)} />
```
xscottxbrownx: Getting a warning about this...

Screenshot 2023-06-24 at 8 17 42 AM

it doesn't give *this* warning, but I believe should have some sort of label tied to this too. maybe `<Typography>` can take an `htmlFor` that is same as `<Input>` `id`??

src/components/NavBar/NavMenu.jsx
```diff
@@ -53,14 +53,13 @@ const NavMenu = ({ menuId, openMenu, setOpenMenu, anchorEl, setAnchorEl, setShow
           </Button>
         </MenuItem>
         <MenuItem>
```
xscottxbrownx, line 63: Not really the scope of this PR, but seems that each `<MenuItem>` is clickable but does nothing but close the menu... user needs to make sure to actually be on the button within the `<MenuItem>`.  Something should look to address in future for better UX.

src/utils/network/session-core.js
```diff
@@ -293,3 +305,61 @@ export const getBlobFromSolid = async (session, fileUrl) => {
...
+
+  const profileInfo = { profileName, organization };
+
+  // TODO: include more fields to the object like organization, address, etc.
+  // when expanding this feature
+  return { profileInfo, profileDataset, profileThing };
+};
+
+/**
+ * A function fetches the user's profile information from their webId's profile
```
xscottxbrownx: change comment to match. Says "fetches", but think this is "updates"

src/components/Profile/ProfileInputField.jsx
```diff
@@ -0,0 +1,70 @@
...
+        <Typography>
+          {inputName}: {inputValue}
+        </Typography>
+      )}
+      <Button
+        variant="outlined"
+        type="button"
+        onClick={editInputValue ? handleCancelEdit : handleEditInput}
+      >
+        {editInputValue ? 'Cancel' : 'Edit'}
```
xscottxbrownx: Also think I'd prefer the "edit button" to be the `EditOutlined` icon instead, but that would make the code a little more to handle.   I like the code with the button and ternary to handle it.  But from a UI standpoint, I'd personally prefer the icon.  Not a deal breaker by any means, definitely approve of button instead.

xscottxbrownx (changes requested):
Overall pretty solid. Some minor changes requested.

No idea what else we would need to live in this `Profile`, but we can add as we figure it out.

xscottxbrownx:
Oh, and need to merge latest changes into this branch.  Doesn't have the table placeholders update at least.

src/utils/network/session-core.js
leekahung: Fixed

src/components/Profile/ProfileInputField.jsx
leekahung: Included `htmlFor` and `id`, but swapped `<Typography>` for `<InputLabel>`

src/components/Profile/ProfileInputField.jsx
leekahung: I can do that. Having these updated with icons and colors.

src/components/Profile/ProfileInputField.jsx
leekahung: ~~This is a typedef which is used to define the prop~~

leekahung: Oh, you're seeing the older version of this, nevermind.

src/components/Profile/ProfileInputField.jsx
leekahung: Has been updated

src/components/Profile/ProfileInputField.jsx
xscottxbrownx: > Oh, you're seeing the older version of this, nevermind.

No, the `typedef` and below is fine. It's the comment above, this is not the `MessageFolder` or its props. This is all for `Profile`.

src/utils/network/session-core.js
```diff
@@ -293,3 +305,72 @@ export const getBlobFromSolid = async (session, fileUrl) => {
...
+ * card
+ *
+ * @function fetchProfileInfo
+ * @param {Session} session - Solid's Session Object {@link Session}
+ * @returns {Promise<object>} Object - The object containing the information related
+ * to the person on their profile card, the profile dataset, and the profile Thing
+ */
+export const fetchProfileInfo = async (session) => {
```
timbot1789: Can we move these profile functions into separate files? Like how we have for documents and user list. Dumping all the business logic into a single monolithic file like `session-core` is pretty bad practice.

src/utils/network/session-core.js
```diff
@@ -293,3 +305,72 @@ export const getBlobFromSolid = async (session, fileUrl) => {
...
+  const profileDataset = await getWebIdDataset(session.info.webId);
+  const profileThing = getThing(profileDataset, session.info.webId);
+
+  const profileName = getStringNoLocale(profileThing, RDF_PREDICATES.profileName);
+  const memberOf = getStringNoLocale(profileThing, RDF_PREDICATES.memberOf);
```
timbot1789: According to schema.org, the object of a `memberOf` predicate isn't a string. It's an Organization (which is a child of thing), so this schema is incorrect.

https://schema.org/Organization

timbot1789 (changes requested):
A few changes requested.

Have you shown this page to Gabby or the rest of UX? It seems pretty similar in concept to a lot of the mockups they have. Would be nice to see one of those used in the implementation. I think basing this PR off one of the forms they have in the mockups would make it more useful.

src/routes/Profile.jsx
```diff
@@ -0,0 +1,131 @@
...
+ */
+const Profile = () => {
+  const location = useLocation();
+  const { session } = useSession();
+
+  localStorage.setItem('restorePath', location.pathname);
+
+  const [profileName, setProfileName] = useState(null);
+  const [memberOf, setMemberOf] = useState(null);
+  const [editProfileName, setEditProfileName] = useState(false);
```
timbot1789: I feel like this state could be local to the `ProfileInputField`. Then you wouldn't need to pass it in or define a bunch of booleans up here.

src/components/NavBar/NavbarLinks.jsx
```diff
@@ -105,7 +105,7 @@ const NavbarLinks = () => {
 
       {/* will show on larger screen (tablet/desktop view) */}
       <Tabs
-        value={location}
+        value={location === 'profile' ? false : location}
```
timbot1789: A more robust test would be to change `routesArray` to `tabsArray`, and then check if the location we're navigating to is in the tabs array. Then we don't have to add a different check for every non-tab route.

src/components/Profile/ProfileInputField.jsx
```diff
@@ -0,0 +1,82 @@
...
+  inputName,
+  inputValue,
+  editInputValue,
+  setInputValue,
+  handleEditInput,
+  handleCancelEdit
+}) => (
+  <Box sx={{ display: 'flex', alignItems: 'center', gap: '10px' }}>
+    <Box sx={{ display: 'flex', alignItems: 'center', gap: '10px' }}>
+      {editInputValue ? (
```
timbot1789: This doesn't need to be a prop. It can be managed locally by this input component.

src/routes/Profile.jsx
```diff
@@ -0,0 +1,131 @@
...
+    setEditMemberOf(!editMemberOf);
+  };
+
+  const fetchProfileData = async () => {
+    const profileObject = await fetchProfileInfo(session);
+
+    if (profileObject.profileInfo.profileName !== null) {
+      setProfileName(profileObject.profileInfo.profileName);
+    } else {
+      setProfileName('No name set');
```
timbot1789: This seems error-prone. I feel like this will result in people submitting their name as "No name set". If a person doesn't have a name, leave it as `null` or `''`. Then if you want to display a user-friendly message in the component, display the message there. Something like:

`<NameDisplay>{profileName ? profileName : 'No name set'}</NameDisplay>`

or 

```
const NameDisplay = ({ name }) => (
 <span>{name ? name : 'No name set'}</span>
)

... In a different file ...

return (
  <NameDisplay name={profileName} />
)
```

timbot1789: In fact, I was able to do just that by hitting `edit` and `update` without changing the value. It set the name on my profile as 'No name set'

src/utils/network/session-core.js
leekahung: Switching this back to Organization.

src/routes/Profile.jsx
leekahung: Moving edit states into `ProfileInputField`. Also moved form into each child component to more easily handle form submission and edit state.

src/components/Profile/ProfileInputField.jsx
leekahung: Falling back to just using InputLabel and Input MUI components.

src/routes/Profile.jsx
leekahung: Will address this soon in another commit later today.

leekahung:
> A few changes requested.
> 
> Have you shown this page to Gabby or the rest of UX? It seems pretty similar in concept to a lot of the mockups they have. Would be nice to see one of those used in the implementation. I think basing this PR off one of the forms they have in the mockups would make it more useful.

Not yet, but yeah, I'm planning to. From what I can see from Figma, there are fields of interest we should consider including to this page from viewing profile information, but I don't see a mockup for editing profile information yet so we'll wait for that.

For now, I think we can have this page set up first.

src/components/NavBar/NavbarLinks.jsx
leekahung: Created an array of acceptable routes in Tabs and set any route outside of tabRoutes to false

src/routes/Profile.jsx
leekahung: Got this fixed with modifying setState for profileName and organization to accept potential null and set to "No value set" in state is null.

The updateProfileInfo has also been adjusted to handle cases with null values being submitted. Hitting `edit` and `update` without change the value if the state is null, "null", or "No value set" will trigger an early return. Hitting `edit` and `update` when the value is an empty string will trigger Thing to be removed from profile. Hitting `edit` and `update` when the value is a string will trigger Thing to be updated from profile by default.

src/utils/network/session-core.js
leekahung: Moving `fetchProfileInfo`, `updateProfileInfo`, and `updateActivity` functions into `profile-helper.js` inside `utils/network`

src/components/Profile/ProfileInputField.jsx
leekahung: JSDoc comment for `profileInputFieldProps` has been fixed along with `ProfileInputField`.

xscottxbrownx: > JSDoc comment for `profileInputFieldProps` has been fixed along with `ProfileInputField`.

Line 17:

Screenshot 2023-06-26 at 10 26 31 AM

src/components/NavBar/NavMenu.jsx
xscottxbrownx: Added as an issue for future

src/utils/network/profile-helper.js
```diff
@@ -0,0 +1,101 @@
...
+ * @param {Session} session - Solid's Session Object {@link Session}
+ * @returns {Promise<object>} Object - The object containing the information related
+ * to the person on their profile card, the profile dataset, and the profile Thing
+ */
+export const fetchProfileInfo = async (session) => {
+  const profileDataset = await getWebIdDataset(session.info.webId);
+  const profileThing = getThing(profileDataset, session.info.webId);
+
+  const profileName = getStringNoLocale(profileThing, RDF_PREDICATES.profileName);
+  const organization = getStringNoLocale(profileThing, RDF_PREDICATES.organization);
```
timbot1789: No, you had this relationship correct the first time. The person is a member of an organization. The issue was that the data you put for organization was incorrect.

I'd suggest you read the schema documentation more carefully. `Organization` is not text. It's a thing. These functions need to retrieve and restore things, not strings.

test/contexts/__snapshots__/SignedInUserContext.test.jsx.snap
```diff
@@ -2,8 +2,6 @@
 
 exports[`SignedInUserContext > fetches user data if user is logged in 1`] = `
 <div>
-  <div>
-    https://example.com/pod/
-  </div>
+  <div />
```
timbot1789: What's this update?

src/utils/network/profile-helper.js
```diff
@@ -0,0 +1,101 @@
+import {
+  buildThing,
+  getStringNoLocale,
+  getThing,
+  getWebIdDataset,
+  removeStringNoLocale,
+  saveSolidDatasetAt,
+  setThing
+} from '@inrupt/solid-client';
+import { RDF_PREDICATES } from '../../constants';
```
timbot1789: I feel like this would be better set up like the model-helpers file. That way the project doesn't have several different styles of retrieving and interacting with data models.

src/components/NavBar/NavbarLinks.jsx
```diff
@@ -35,6 +35,8 @@ const NavbarLinks = () => {
     { label: 'Messages', path: '/PASS/messages' }
   ];
 
+  const tabRoutes = ['clients', 'documents', 'messages'];
```
timbot1789: This duplicates the data found in routesArray immediately above. Why can't you just use the routes array?

timbot1789 (changes requested):
I'm still not sure what value this page provides. Can you write up a design doc of what exactly you intend for this page to do? Right now it's really only a name edit. We don't have a strong idea of what an Organization is in PASS, so that organization field doesn't provide any value. It will need to be rewritten.

leekahung:
> I'm still not sure what value this page provides. Can you write up a design doc of what exactly you intend for this page to do? Right now it's really only a name edit. We don't have a strong idea of what an Organization is in PASS, so that organization field doesn't provide any value. It will need to be rewritten.

I believe it's stated in the opening post? The page is intended to be a place where users can update their profile without needing to touch the TTL file for it. So in the future when a user needs to fill out say their "Phone number" or simple address:

Screen Shot 2023-06-26 at 11 39 25 (2)

What should be in that profile card could be discussed further in depth during meetings. With this page users can fill it all out in `Profiles`.

For Organization, I think we could leave it in there as a stand-in for something we could replace it with.

src/components/NavBar/NavbarLinks.jsx
leekahung: I think I could do that, I thought this was what was being described as the suggested solution originally. But this can be adjusted, I'm going to extract location names from inside routesArray via label.

src/utils/network/profile-helper.js
leekahung: Do you mean to move `profile-helper` into `model-helpers` so they're all in one location? Think I could do that.

test/contexts/__snapshots__/SignedInUserContext.test.jsx.snap
leekahung: I believe this has to do with the new updateActivity function for the profile card inside useEffect. When including it, says the container does not match the snapshot.

Screen Shot 2023-06-26 at 12 08 00

So I've updated the snapshot with `u`.

Screen Shot 2023-06-26 at 12 10 06

Without it, it's the same as before asking for `https://example.com/pod/` in that location.

Screen Shot 2023-06-26 at 12 00 32

Should I revert the snapshot back?

src/utils/network/profile-helper.js
leekahung: I think we can build another Thing for `Organization` inside the profile card in the mean time.

I'm not too sure about which functions could be use to set Things inside of other Things at this moment.

leekahung: Think I got this working as intended (as in making a new Thing as oppose to making a Thing inside a Thing).

Organization would be created as a new Thing and be adjust for accordingly.

https://github.com/codeforpdx/PASS/assets/14917816/bd30f241-e306-42c2-ad2c-bd4624d090b0

test/contexts/__snapshots__/SignedInUserContext.test.jsx.snap
leekahung: For now, I've reverted the snapshot back to the original form in Development.

timbot1789: If it's affecting the rendering of the snapshot, and you're not sure why, there's probably a bug.

Although I think there may be a pre-existing issue with that test. It seems like it's not reading the mocked ENV properly. I'd leave this snapshot unchanged from development for now, and we can fix that issue in another PR.

leekahung: Got it, I'll leave it alone. I'll start an issue for it stating it could be a bug.

leekahung (commented):
Think I've touched on comments that could be addressed here. Let me know if there's still anything else that needs to be changed or comment on.

@xscottxbrownx @timbot1789 @andycwilliams

src/components/Profile/ProfileInputField.jsx
xscottxbrownx: Still this

src/components/Profile/ProfileInputField.jsx
```diff
@@ -0,0 +1,98 @@
...
+            variant="outlined"
+            type="button"
+            color={edit ? 'error' : 'primary'}
+            endIcon={edit ? <ClearIcon /> : <EditIcon />}
+            onClick={edit ? handleCancelEdit : handleEditInput}
+          >
+            {edit ? 'Cancel' : 'Edit'}
+          </Button>
+        </Box>
+        {edit && (
```
xscottxbrownx: Feel the conditional rendering of `Update` button should also be inside the inner `<Box>` with the rest.

src/routes/Profile.jsx
```diff
@@ -0,0 +1,78 @@
...
+          User WebId:{' '}
+          <Link href={session.info.webId} target="_blank" rel="noreferrer">
+            {session.info.webId}
+          </Link>
+        </Typography>
+
+        {/* TODO: Refactor/optimize the form below once we have more input */}
+        {/* fields to update profile for */}
+        <Box style={{ display: 'flex', flexDirection: 'column', gap: '15px' }}>
+          <ProfileInputField
```
xscottxbrownx: Ultimately, I think from a UX perspective I'd prefer the whole thing be a single form instead of a form for each input and an edit button for each input.   I think 1 edit button that opens all inputs to make changes and a single "SAVE CHANGES" button to submit the form and run the functions.    If this is agreed upon direction, could be a future PR to address/optimize.

src/components/Profile/ProfileInputField.jsx
```diff
@@ -0,0 +1,98 @@
...
+          <InputLabel htmlFor={`input-${inputName}`} sx={{ color: 'black' }}>
+            {inputName}:{' '}
+          </InputLabel>
+          <Input
+            id={`input-${inputName}`}
+            value={inputValue === null ? 'No value set' : inputValue}
+            disabled={!edit}
+            onChange={(e) => setInputValue(e.target.value)}
+          />
+          <Button
```
xscottxbrownx: Nitpick thought...

Like this or make some conditional rendered buttons based on edit state/not edit state instead?

Example: 

```js
return (
    <form onSubmit={handleUpdateProfile}>
      <Box sx={{ display: 'flex', alignItems: 'center', gap: '10px' }}>
        <Box sx={{ display: 'flex', alignItems: 'center', gap: '10px' }}>
          <InputLabel htmlFor={`input-${inputName}`} sx={{ color: 'black' }}>
            {inputName}:{' '}
          </InputLabel>
          <Input
            id={`input-${inputName}`}
            value={inputValue || ''}
            placeholder={inputValue || 'No value set'}
            disabled={!edit}
            onChange={(e) => setInputValue(e.target.value)}
          />
          {!edit && (
            <Button
              variant="outlined"
              type="button"
              color="primary"
              endIcon={<EditIcon />}
              onClick={handleEditInput}
            >
              Edit
            </Button>
          )}
          {edit && (
            <>
              <Button
                variant="outlined"
                type="button"
                color="error"
                endIcon={<ClearIcon />}
                onClick={handleCancelEdit}
              >
                Cancel
              </Button>
              <Button variant="outlined" type="submit" endIcon={<CheckIcon />}>
                Update
              </Button>
            </>
          )}
        </Box>
      </Box>
    </form>
  );
  ```

File: src/components/Profile/ProfileInputField.jsx [hunk 21]
```diff
@@ -0,0 +1,98 @@
...
+  return (
+    <form onSubmit={handleUpdateProfile}>
+      <Box sx={{ display: 'flex', alignItems: 'center', gap: '10px' }}>
+        <Box sx={{ display: 'flex', alignItems: 'center', gap: '10px' }}>
+          <InputLabel htmlFor={`input-${inputName}`} sx={{ color: 'black' }}>
+            {inputName}:{' '}
+          </InputLabel>
+          <Input
+            id={`input-${inputName}`}
+            value={inputValue === null ? 'No value set' : inputValue}
```
xscottxbrownx: Piggybacking on Tim's earlier comment, I'm not positive this is what we want. 
BUT I will gladly admit my understanding can be completely wrong.

---

I don't think the value should ever be `No value set`.  I think that makes a great placeholder though.
I feel this should be `value={inputValue || ''}` and also a `placeholder={inputValue || 'No value set'}`

---

Also the current way it is, you can make the name blank - but there is an error if you try to make the organization blank.
1) gotta deal with error (the code above seems to be fine for that)
2) Name should be required
3) Organization - maybe if in edit mode it should be required (so can't be saved as blank), but still can be cancelled?

src/components/Profile/ProfileInputField.jsx
```diff
@@ -0,0 +1,98 @@
...
+    const inputField = inputName === 'Name' ? 'profileName' : inputName.toLowerCase();
+
+    await updateProfileInfo(session, profileData, inputField, inputValue);
+
+    fetchProfileData();
+    setEdit(false);
+  };
+
+  const handleCancelEdit = () => {
+    fetchProfileData();
```
xscottxbrownx: I might be missing why we need to `fetchProfileData()` here. Seems to me that sets the name and organization, but there would be no reason to set those states as there was no change correct? This is canceling the edit.

src/routes/Profile.jsx
```diff
@@ -0,0 +1,78 @@
...
+  const [organization, setOrganization] = useState(null);
+
+  const fetchProfileData = async () => {
+    const profileObject = await fetchProfileInfo(session);
+
+    setProfileName(profileObject.profileInfo.profileName);
+    setOrganization(profileObject.profileInfo.organization);
+  };
+
+  useEffect(() => {
```
xscottxbrownx, line 72: I'm trying to understand this `useEffect` better. 

If this is for first render, we could just have a `fetchProfileData()` without the `useEffect`.
If we are watching to see if the `profileName` is null, I don't think we should allow the user to be able to make it null or empty.

src/routes/Profile.jsx
```diff
@@ -0,0 +1,78 @@
...
+const Profile = () => {
+  const location = useLocation();
+  const { session } = useSession();
+
+  localStorage.setItem('restorePath', location.pathname);
+
+  const [profileName, setProfileName] = useState(null);
+  const [organization, setOrganization] = useState(null);
+
+  const fetchProfileData = async () => {
```
xscottxbrownx: naming?   The function does fetch the profile info, but ultimately sets the profile name and organization.

So, is the function doing too many things or just named a little funky?

xscottxbrownx (commented):
Few things to discuss in there...

src/components/Profile/ProfileInputField.jsx
leekahung: A bit confused at first since I thought I fixed that on the file, but I'll have this patched up.

src/routes/Profile.jsx
leekahung: Think we could have this opened as a discussion on an issue.

src/components/Profile/ProfileInputField.jsx
leekahung: Could certainly break it up into more distinct elements within the component. Think a ternary might work a bit better for this example.

I'll have this re-worked as a ternary.

src/components/Profile/ProfileInputField.jsx
leekahung: I think that'll be good for `value` and `placeholder`. I'll have this updated to use:

```
value={inputValue || ""}
placeholder={inputValue || 'No value set'}
```

As for Organization, I could be misunderstanding the issue, but I could save blanks for company name. If the name is blank, it's basically the same as no value set.

https://github.com/codeforpdx/PASS/assets/14917816/768c9719-6270-4f40-9d85-f5f3a287d148

src/components/Profile/ProfileInputField.jsx
leekahung: When editing the inputs, if you were to cancel after editing, it'll remain what the input was when you canceled it. This is because onChange uses a set function each time the state is updated.

For example, say input was "Apple" when you want to edit, but was "Appl" when you canceled it. It'll remain "Appl" rather than the original "Apple".

src/routes/Profile.jsx
leekahung: Think we can make the dependency array empty so that it'll only run once when rendered. It was originally intended for just that.

src/routes/Profile.jsx
leekahung: The function suppose to fetch profile info from the profile and set the local states based on what was on the profile card.

Perhaps I should rename this `getInitialProfileData` or just `getProfileData`? I want to avoid using `set` to prevent confusion with React's set functions.

I'm also thinking along the lines of `initializeLocalProfileData` or `loadProfileData`. For now, I think `loadProfileData` seems more appropriate?

src/components/Profile/ProfileInputField.jsx
leekahung: Got this moved inside when switching to use ternary.

src/routes/Profile.jsx
timbot1789: I think Scott's right, that would be a much simpler and more intuitive form. I'd rather not introduce code into the system that immediately creates issues. If there are follow-up tasks needed, then I think it's a good idea to state explicitly in advance what those follow-up tasks are and why they aren't being done here.

It's ok if this PR doesn't finish the feature. I think it's perfectly fine to say something like "this is the first of 3 planned PRs. The first PR does these things, the second PR will build off that with these other things, and the third PR will package everything up in a nice tidy bow" I just want to know what those PRs are, and what the ultimate goal of this system is. What do you expect this system to look like when everything is done?

timbot1789 (changes requested):
There are some code issues left, but it's still not clear to me what problem this is trying to solve. Why would a person use this page? Why would they want to set up a profile? It's a lot of work, so what value do they get out of it?

A few specific things that are unclear to me:

- Whose profile is this? You say in the opening that it's the signed-in user's profile, but who is that? The example figma that you cite shows a case manager viewing a client's profile, which is a different situation. Most of the information on that figma also relates to clients, not case managers. So are you expecting the case manager to have an identical profile page to a client?
- Why would they want to edit their name? Most people don't change their names very frequently. They set it once, and then need to go through a process to change it again. There are also downstream implications whenever a person changes their name. People need to be notified of the name change.
- What will the organization be, and why would the user want to edit it? If this is a profile of a client, it's not clear to me they'd be part of a single organization at all. If it's a case manager, it's not clear to me they should be able to change what organization they belong to.
- What are the intended privacy levels of this data? As is, everything is being stored directly on the solid profile card, which is public. Some of the data your example final UI has probably shouldn't be public.

src/routes/Profile.jsx
leekahung: Basically, what I'm expecting when this page is done, a user can view and edit their profile card on PASS. How it’ll be implemented I felt a singular input approach seemed appropriate, which was why the PR was structured like this.

But I could switch this up to have it be just one big form.

leekahung: If I'm having it switched up to one form, I think we could keep thinking in one PR. Any potential extentions to this will just be adding input fields to the form.

leekahung:
> There are some code issues left, but it's still not clear to me what problem this is trying to solve. Why would a person use this page? Why would they want to set up a profile? It's a lot of work, so what value do they get out of it?

The main value of this page is so that users could go into PASS, view and update their profile card when needed. For example in opencommons.net, the user's name does not exist compared to solidcommunity.net which asks users for their names on Pod creation. So for servers such as those, webIds associated with these servers will have names turn out as null. They'll still have the webId, but no name, see images of untouched profile card from opencommons.net and solidcommunity.net.

Screen Shot 2023-06-29 at 13 32 00 (2)

Screen Shot 2023-06-29 at 13 34 23 (2)

We could potentially funnel/recommend users of PASS into servers that mandates names or other information we might want for a profile card during account creation. But I think that'll be a separate topic on its own.

Outside of names, there a lot of information a user could update over time such as phone number, street address, date of birth. Those could be categorized further when we want to separate what information should be on the public profile card and a private profile card, which I'll call `private_card.ttl`.

The `private_card` will be stored inside `PASS` itself as oppose to the public profile card, which remains in `profile`. We don't have any private information included yet in this PR (it's just the name and organization), so a `private_card` has not been made for this.

> * Whose profile is this? You say in the opening that it's the signed-in user's profile, but who is that? The example figma that you cite shows a case manager viewing a client's profile, which is a different situation. Most of the information on that figma also relates to clients, not case managers. So are you expecting the case manager to have an identical profile page to a client?

This profile is the user's personal card. This page was intended to be accessed by just the user like how users manage information on their own account. The Figma example includes information related to the client and that is from the case worker/manager's perspective. And that perspective will be in the "Clients" page, not "Profile". If a case manage clicks on "Profile", it'll being up their own profile card. Case workers/managers are still PASS users. They could go in and update their own profile cards if they like.

The information that a case worker sees from the Figma comes from the clientList. Information from the client list could potentially use information from the public profile card of the client and, with permission, the `private_card` in `PASS` to fill out fields if information is missing after the initial appointment, but these two sources (clientList and profile card) are separate entities.

> * Why would they want to edit their name? Most people don't change their names very frequently. They set it once, and then need to go through a process to change it again. There are also downstream implications whenever a person changes their name. People need to be notified of the name change.

It's because names are not mandated by Solid, just webIds for denoting an agent. There could be servers that doesn't ask for names and for those servers, the names of the user will just be `null` since there's no name associated with their webId.

Here is an example of using name from an account that's unaltered:

Screen Shot 2023-06-29 at 13 19 57

A more elaborate process in the future for further name changes outside of the initial `null` to `something` could be considered.

For example, perhaps an organization could verify that the name is the profile card matches the information found in their documents and a marker could be used to identify it as valid.

If they change their names again after that, they'll lose that marker until it's verified again.

> * What will the organization be, and why would the user want to edit it? If this is a profile of a client, it's not clear to me they'd be part of a single organization at all. If it's a case manager, it's not clear to me they should be able to change what organization they belong to.

As for what would the organization be and why would a user want to edit it, it'll be up to the case worker to update their own profile card. If they're not belonging to an organization, it does not have to be filled out. If they know they're part of a group, they could fill out the organization's name. Whether it'll be verified the person is actually part of the organization will have to be resolved later similar verified files.

If it's an organization associated with using PASS (i.e. an organizational Pod) it could be verified through some kind of verification process from that organization. If it's an organization not associated with PASS there would be an alternative approach to verifying a member is part of an organization (i.e. a redacted paystub, member card, etc.)

How to indicate that it's a verified data point could be similar to the name. If the user change the name after it's verified, it'll lose it's verified status until it's reverified.

> * What are the intended privacy levels of this data? As is, everything is being stored directly on the solid profile card, which is public. Some of the data your example final UI has probably shouldn't be public.

The intended privacy level for this data at least that in the profile card is public. Things like name or name of organization could be placed in there. As for some of the data in my example, I don't see which information on there should remain private. At the moment, it's showing:

- User's webId
- User's name
- User's Solid Provider
- User's Last login date
- User's organization name

https://github.com/codeforpdx/PASS/assets/14917816/f25ca135-6c02-4f1a-b3cf-51fe4d9f72e3

When when it was bare at account creation, it showed:

- User's webId
- User's Solid Provider

Screen Shot 2023-06-29 at 13 32 00 (2)

Do you mean that we might want to hide the user's login date to a `private_card.ttl`?

As mentioned in earlier comments, everything will be stored in profile card will be public information. Items that are more sensitive/private like street address or date of birth will be stored in a separate location in `/PASS/private_card.ttl`. That card would require explicit permission to access. But we don't have any information related to that currently so it's not built.

src/components/Profile/ProfileInputField.jsx
leekahung: Believe I have this fixed. The comments are now:

Screen Shot 2023-06-29 at 13 25 30

timbot1789:
@leekahung Wow, that's a lot of writing. If you think adding a solid profile card editor to PASS is a valuable use of time, I'd recommend you base it off the minimum recommended profile information given in the solid spec: https://github.com/solid/solid-spec/blob/master/solid-webid-profiles.md#minimum-recommended-profile-information

Leave off organization for now. It's not clear what precisely that should be, who should have it on their profile, or if it's public information.

Also leave off the user activity. What actions a user takes on the web is most certainly NOT public information.

Adding a profile picture upload to this would be a good addition that could make PASS look a lot friendlier.

You'll need to be careful with the implementation. As [Jeff points out here](https://forum.solidproject.org/t/how-does-one-expand-the-profile-card/1264), creating an invalid card will lock the person out of Solid, so your code will need extensive unit tests.

> We could potentially funnel/recommend users of PASS into servers that mandates names or other information we might want for a profile card during account creation.

Or we could require that they give a name when they first register with PASS. There are always several alternative designs. Part of this process is figuring out what all reasonable designs are, and deciding what seems best for the use case.

leekahung:
> @leekahung Wow, that's a lot of writing. If you think adding a solid profile card editor to PASS is a valuable use of time, I'd recommend you base it off the minimum recommended profile information given in the solid spec: https://github.com/solid/solid-spec/blob/master/solid-webid-profiles.md#minimum-recommended-profile-information
> 
> Leave off organization for now. It's not clear what precisely that should be, who should have it on their profile, or if it's public information.
> 
> Also leave off the user activity. What actions a user takes on the web is most certainly NOT public information.
> 
> Adding a profile picture upload to this would be a good addition that could make PASS look a lot friendlier.
> 
> You'll need to be careful with the implementation. As [Jeff points out here](https://forum.solidproject.org/t/how-does-one-expand-the-profile-card/1264), creating an invalid card will lock the person out of Solid, so your code will need extensive unit tests.
> 
> > We could potentially funnel/recommend users of PASS into servers that mandates names or other information we might want for a profile card during account creation.
> 
> Or we could require that they give a name when they first register with PASS. There are always several alternative designs. Part of this process is figuring out what all reasonable designs are, and deciding what seems best for the use case.
> 

Ah, there's a spec for the profile! Thanks Tim, I'll have a look over it.

From a brief reading, I think we can have, at the very least, the foaf:name, foaf:img, foaf:nick, and pim:storage as stated in the minimal recommended profile information.

Img and nick should be straightforward in our implementation. Think we could have it follow what I've already implemented with name and have it placed within `#me`. For Img, think we could be replaced with a placeholder image when null and a fixed size img element when filled.

For pim:storage, I'll take a look at solidcommunity.net again to see what route they're using for the value, but it's likely doing to be excluded from being displayed since it'll be more of an under the hood value for Solid.

As for the cert:key, I'm not too familiar with how I'll go about generating it for the user. I think we can consider that in the near future.

timbot1789:
> For pim:storage, I'll take a look at solidcommunity.net again to see what route they're using for the value, but it's likely doing to be excluded from being displayed since it'll be more of an under the hood value for Solid.
> 
> As for the cert:key, I'm not too familiar with how I'll go about generating it for the user. I think we can consider that in the near future.

Yeah, don't worry about cert:key and pim:storage. Community solid server doesnt support `pim:storage` properly anyway, and setting public keys is too complex for the average user. It's useful for machines, but not humans. We can come back to that later.

src/model-helpers/Profile.js
```diff
@@ -0,0 +1,128 @@
...
+  createThing,
+  getStringNoLocale,
+  getThing,
+  getWebIdDataset,
+  removeStringNoLocale,
+  removeThing,
+  saveSolidDatasetAt,
+  setThing
+} from '@inrupt/solid-client';
+import { RDF_PREDICATES } from '../constants';
```
timbot1789, line 18: Since these functions are all related to the currently signed in user, can we put them in the `SignedInUserContext`? That way more components will be able to access the data. You can look at `UserList` and `DocumentList` for models of how to do that.

leekahung: I have placed the functions being used in `Profiles` as part of `SignedInUserContext`. I think that's what this comment refers to.

src/routes/Profile.jsx
```diff
@@ -0,0 +1,182 @@
...
+import EditIcon from '@mui/icons-material/Edit';
+import Link from '@mui/material/Link';
+import Typography from '@mui/material/Typography';
+// Utility Imports
+import { getBlobFromSolid } from '../utils';
+// Contexts Imports
+import { SignedInUserContext } from '../contexts';
+// Component Inputs
+import ProfileInputField from '../components/Profile/ProfileInputField';
+import ProfileImageField from '../components/Profile/ProfileImageField';
```
xscottxbrownx: don't need both

src/model-helpers/Profile.js
```diff
@@ -0,0 +1,141 @@
...
+  });
+
+  profileDataset = setThing(profileDataset, profileThing);
+
+  await saveSolidDatasetAt(session.info.webId, profileDataset, { fetch: session.fetch });
+};
+
+/**
+ * Function that upload profile image of user onto their profile container and
+ * reference it on their
```
xscottxbrownx: "on their" ...   ?

src/routes/Profile.jsx
leekahung: I believe you mean to combine them into one import line? I'll create a new index.js for `/components/Profile/`

src/model-helpers/Profile.js
leekahung: Good catch! I'll fix this up to `on their Pod`.

src/routes/Profile.jsx
xscottxbrownx: I mistakenly read this as importing the exact same thing twice!

xscottxbrownx:
All of my comments/concerns were addressed on this PR.  Leaving open for Tim to check his (or if anybody else has any.)

xscottxbrownx:
The `ProfileImageField.jsx`'s `<InputLabel>` isn't linked to the input.  Need the `id` on the `input` instead of the `button`. There's an issue in console currently.

leekahung:
> The `ProfileImageField.jsx`'s `<InputLabel>` isn't linked to the input. Need the `id` on the `input` instead of the `button`. There's an issue in console currently.

Getting this fixed up by changing `<InputLabel>` into `<Typography>`. Since this field is mostly just an image and a button which fetches or removes the image, think we can make this switch without an issue.

xscottxbrownx:
Ok, NOW all my concerns have been addressed. Looks like something with test is failing. 

I don't have the knowledge to help there.

src/constants/rdf_predicates.js
```diff
@@ -24,7 +24,10 @@ const RDF_PREDICATES = {
   sender: 'https://schema.org/sender',
   profileName: 'http://xmlns.com/foaf/0.1/name',
   sha256: 'https://schema.org/sha256',
-  accountablePerson: 'https://schema.org/accountablePerson'
+  accountablePerson: 'https://schema.org/accountablePerson',
+  profileImg: FOAF.img,
+  nickname: FOAF.nick,
+  type: RDF.type
```
timbot1789: Is this `type` parameter still being used? If not we can remove it.

leekahung: Yeah, think I can remove this.

src/components/Profile/ProfileImageField.jsx
```diff
@@ -0,0 +1,111 @@
...
+        maxWidth: '300px',
+        alignItems: 'flex-start',
+        gap: '10px'
+      }}
+    >
+      <Typography color="black">Profile Image: </Typography>
+      {profileImg ? (
+        <>
+          <Avatar
+            src={profileImg}
```
timbot1789: You don't need to manually handle loading this image. The HTML `img` element at the base of this avatar will handle it far better than we can. Just pass in the URL of the image and the browser will take care of it.

src/routes/Profile.jsx
```diff
@@ -0,0 +1,193 @@
...
+  // handleGetProfileImage
+  const [profileImg, setProfileImg] = useState(
+    JSON.parse(localStorage.getItem('profileImageBlob'))
+  );
+
+  const [edit, setEdit] = useState(false);
+
+  const loadProfileData = async () => {
+    const profileDataSolid = await fetchProfileInfo(session);
+    localStorage.setItem('restoreProfileData', JSON.stringify(profileDataSolid));
```
timbot1789: `localStorage` usually has a maximum size of 5mb. This will error on any image larger than that (and there are a lot of images larger than 5mb). You don't need to handle loading the image anyway. Just let the browser fetch it.

test/model-helpers/Profile.test.js
```diff
@@ -0,0 +1,30 @@
...
+let session = {};
+
+vi.mock('@inrupt/solid-client');
+
+describe('fetchProfileInfo', () => {
+  beforeEach(() => {
+    session = {
+      fetch: vi.fn(),
+      info: {
+        webId: 'https://newtestuser2.opencommons.net/profile/card#me'
```
timbot1789: Would be better not to use a real webId in tests. That way no one will get it confused with real data.

src/components/Profile/ProfileImageField.jsx
leekahung: Oh, I didn't know that. Going to switch to using `Avatar` as the default icon. Just going to set the condition for the buttons.

src/model-helpers/Profile.js
```diff
@@ -0,0 +1,141 @@
...
+ */
+export const updateProfileInfo = async (session, profileData, inputValues) => {
+  let { profileDataset, profileThing } = profileData;
+  const { profileInfo } = profileData;
+
+  Object.keys(inputValues).forEach((input) => {
+    switch (inputValues[input]) {
+      case null:
+      case 'null':
+      case 'No value set':
```
timbot1789: What are these cases? It seems to me that if a person sets their name as `'no value set'` or `'null'`, there's no reason to reject it.

For the case of `null`, it seems we can handle that the same as `''`

src/routes/Profile.jsx
leekahung: I think this is fine. What is being stored is just the string of the link to the blob like `blob:http://localhost:5173/5d4def0c-28eb-4d7d-913d-46a92f23b991` not the actual image file.

test/model-helpers/Profile.test.js
leekahung: Think I'll just mock the ttl file. I'll give it a go.

src/model-helpers/Profile.js
```diff
@@ -0,0 +1,141 @@
...
+ * @function uploadProfileImage
+ * @param {Session} session - Solid's Session Object {@link Session}
+ * @param {object} profileData - The object containing the information related
+ * to the person on their profile card, the profile dataset, and the profile Thing
+ * @param {Blob} inputImage - File blob being uploaded for profile image
+ */
+export const uploadProfileImage = async (session, profileData, inputImage) => {
+  let { profileDataset, profileThing } = profileData;
+
+  const profileContainer = `${session.info.webId.split('profile')[0]}profile/`;
```
timbot1789: `${session.info.webId.split('card')};`

src/routes/Profile.jsx
timbot1789: Then where are you storing the blob?

src/model-helpers/Profile.js
leekahung: Will remove cases for `"null"` and `"no value set"`. Because" no value set" is now just a placeholder, should be fine now compared to before.

For `null`, think I'm going to keep it since `''` doesn't fully cover the cases for when the property is null.

https://github.com/codeforpdx/PASS/assets/14917816/5fc843d3-3328-458c-b292-26ce4cffb589

If I were to were to remove it, it'll return the string `"null"` instead.

https://github.com/codeforpdx/PASS/assets/14917816/1a23a13b-de56-4737-a9fd-8d9002435361

src/model-helpers/Profile.js
leekahung: That was silly of me 😆. You're right, it'll be better to split from `card`. I'll have this switched up.

src/routes/Profile.jsx
leekahung: I'm using the `createObjectURL` from the `URL` API under the hood from `getBlobFromSolid` to reference the image file from the profile card. So the blob is consumed by `createObjectURL` to generate a reference URL to the image file stored in memory on the browser rather than in `localStorage`.

I think we could include an additional `URL.revokeObjectURL` function to explicitly release the blob from memory when a user no longer wish to have a profile image.

timbot1789: That describes how you create it, but not where the data is stored. It seems to end up in browser memory, which is normally cleared when the browser closes or the system restarts. There may be a bit more persistence, but it will vary between browsers, versions, and OSes. Either way I don't think this is a reliable solution. If you pass the URL to the `src` attribute of the `img` tag (or `Avatar` component), you don't need any of this load handling, and it will perform faster.

src/components/NavBar/NavBar.jsx
```diff
@@ -117,7 +141,15 @@ const NavBar = () => {
                   onClick={handleOpenMenu}
                   color="inherit"
                 >
-                  <AccountCircle />
+                  {profileImageBlob ? (
+                    <Avatar
+                      src={profileImageBlob}
```
timbot1789: You can just pass the URL in here as well.

test/model-helpers/Profile.test.js
```diff
@@ -0,0 +1,48 @@
+import * as solidClient from '@inrupt/solid-client';
+import { beforeEach, afterEach, describe, expect, vi, it } from 'vitest';
+import { fetchProfileInfo } from '../../src/model-helpers';
+
+let session = {};
+const mockWebId = 'https://example.com/pod/profile/card#me';
+
+vi.mock('@inrupt/solid-client');
+
+describe('fetchProfileInfo', () => {
```
timbot1789, line 16: Would also like some tests for `uploadProfileInfo`, `uploadProfileImage`, and `removeProfileImage`

src/routes/Profile.jsx
leekahung: I'm not too certain on an alternative at this time. I could take out the localStorage links to the reference URL, but the result would be it'll have to load the blob each time. See clip.

https://github.com/codeforpdx/PASS/assets/14917816/152ce169-60b6-40b3-8840-157f1aaeafe9

With regards to browsers (at least with the ones I've tested with: Chrome, Safari, Edge, and FireFox) it seems to be consistent using this approach.

https://github.com/codeforpdx/PASS/assets/14917816/91d16ad5-5fcb-4acd-9282-98e6f047dd61

For the URL, I could only pass in the reference URL after the blob object from Solid is referenced.

src/components/NavBar/NavBar.jsx
leekahung: Got this switched.

test/model-helpers/Profile.test.js
leekahung: Got the new unit tests for the three other functions in `Profile.test.js` included with this PR.

src/components/NavBar/NavMenu.jsx
```diff
@@ -55,12 +64,24 @@ const NavMenu = ({ menuId, openMenu, setOpenMenu, anchorEl, setAnchorEl, setShow
...
           <Button
             variant="text"
-            startIcon={<AccountCircle />}
-            href={session.info.webId}
-            target="_blank"
-            rel="noreferrer"
+            startIcon={
+              profileImageBlob ? (
```
leekahung: Moving this to just use `Avatar` as well.

src/model-helpers/Profile.js
```diff
@@ -0,0 +1,141 @@
...
+    fetch: session.fetch
+  });
+
+  profileThing = buildThing(profileThing)
+    .addUrl(RDF_PREDICATES.profileImg, `${profileContainer}${profileImgFilename}`)
+    .build();
+
+  profileDataset = setThing(profileDataset, profileThing);
+
+  await saveSolidDatasetAt(session.info.webId, profileDataset, { fetch: session.fetch });
```
timbot1789, line 120: Probably want to add an acl to this so the profile pic can be read by the public.

src/routes/Profile.jsx
timbot1789: No my point is you don't have to load the blob yourself. The avatar component can handle all loading for you. Just pass it the profile image url you got from `profileInfo`

That's what I did in this video and I was able to load the images without the flicker.
[Screencast from 07-04-2023 11:11:33 AM.webm](https://github.com/codeforpdx/PASS/assets/37914436/f3d86bca-0ca7-42ce-bf53-606373904a74)

src/model-helpers/Profile.js
leekahung: Got it included with `uploadProfileImage`. I've updated the unit test for it as well to take the new functions to build the ACL into account.

src/routes/Profile.jsx
leekahung: Oh! I get it. I'll test it out with just the URL link to the picture on Solid.

leekahung: I'm not really sure what I'm doing wrong, but I can't seem to get the flickering behavior to go away even when using the direct URL to the image file which I find surprising. The reference URL from the blob seems to work okay though.

https://github.com/codeforpdx/PASS/assets/14917816/1d8e0583-a230-4ab3-bd54-2bf5ab8e1246

I didn't expect the direct link to the image would force a re-render of the image, the `profileImg` prop just passes a URL to `src` in `Avatar` without any processing in Line 71.

Screen Shot 2023-07-04 at 14 22 53

src/model-helpers/Profile.js
```diff
@@ -0,0 +1,160 @@
...
+ * your pod at /profile
+ */
+export const uploadProfileImage = async (session, profileData, inputImage) => {
+  let { profileDataset, profileThing } = profileData;
+
+  const profileContainer = `${session.info.webId.split('card')[0]}`;
+  const profileImgFilename = inputImage.name.replace(' ', '%20');
+
+  await saveFileInContainer(profileContainer, inputImage, {
+    slug: profileImgFilename,
```
timbot1789, line 109: Image upload is failing on solid community. It seems that node solid server does not fully respect the slug suggestion here. If I upload a file name `tim.jpg`, node solid server stores it as `tim.jpg.jpeg`.

image


I believe the `saveFileInContainer` function should return the actual url of the created file. We should be able to store that on the profile instead.

leekahung: Found the bug. This happened because I've turned on `contentType` for `saveFileInContainer` from `solid-client`. If I kept `contentType: inputImage.type` it seems to lump `.jpg` and `.jpeg` together.

I'll have this removed and have the function imply what the file is. Seemed to have fixed this error.

https://github.com/codeforpdx/PASS/assets/14917816/a6356ad2-6563-4bde-b07c-67819f5ab453

src/routes/Profile.jsx
leekahung: Think I've got it to work for the most part. It's been chipped own to just use just the url to the image file on Solid rather than the blob.

Still using `localStorage` for persistence, but it's all moved to inside `ProfileImageField` and I've simplified to a single `useState` (link to image file on Solid) rather than two (one for the blob, one for link to image file on Solid).

`Navbar` and `NavMenu` simply pulls from `localStorage` to render the profile image.

leekahung: Think I got the flickering disappear as well. Think this solution would be good for our "Profile" page.

https://github.com/codeforpdx/PASS/assets/14917816/10509e94-644f-4480-860e-04b5e67fb9e5

src/model-helpers/Profile.js
leekahung: Nevermind, I still see the problem on solidcommunity.net. I'll see if I can patch this again.

leekahung: Seems to be a weird quirk with extensions that has 4- and 3-letter counterparts. For `tiff` files, it'll come out as `<filename>.tiff.tif`, similar to how `.jpg` is returning `<filename>.jpg.jpeg`.

Was able to fix the `.jpg` issue when uploading by converting all `.jpg` into `.jpeg`. The same cannot be said for `.tiff` to `.tif` unfortunately. 

Think we can drop `.tiff` though since we're already supporting `.png`, `.jpg`, `.jpeg`, `.gif`, `.webp`, not to mention `.tiff` seems to only be supported by Safari browsers (https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types). Gonna base most of what we accept as image files from the common image file types shown in that mdn link.

package.json
leekahung, line 1: Testing out GitHub's Release functionality to keep track of changes between versions instead of manually updating CHANGELOG.

Going to follow the semantic versioning convention as well so I'm changing the existing version on package.json to `v0.0.5.1` to `v0.5.1`.

src/model-helpers/Profile.js
timbot1789: Your current solution requires individual mappings for every individual file type to whatever abbreviation node solid server uses. That doesn't seem very maintainable to me. What if we find another server that saves files as `jpg` instead of `jpeg`? What if we get a server that ignores the slug suggestion altogether? Then this code breaks.

The `saveFileInContainer` function returns the URL the server assigned to the file. Just save that URL into the profile dataset. Then you don't have to worry about any of this filetype mapping. There are examples of how to do this on Inrupt's website: https://docs.inrupt.com/developer-tools/javascript/client-libraries/tutorial/read-write-files/#save-a-file-into-an-existing-container

src/components/NavBar/NavBar.jsx
```diff
@@ -60,6 +61,13 @@ const NavBar = () => {
     setShowConfirmationModal(false);
   };
 
+  const { profileData } = useContext(SignedInUserContext);
+  const [profileImg, setProfileImg] = useState(localStorage.getItem('profileImage'));
```
timbot1789, line 65: You already have this information in `profileData`. You don't need to store it in `localStorage` as well.

src/model-helpers/Profile.js
```diff
@@ -0,0 +1,164 @@
...
+  profileThing = buildThing(profileThing)
+    .addUrl(RDF_PREDICATES.profileImg, `${profileContainer}${profileImgFilename}`)
+    .build();
+
+  profileDataset = setThing(profileDataset, profileThing);
+
+  await saveSolidDatasetAt(session.info.webId, profileDataset, { fetch: session.fetch });
+
+  // Create ACL file for profile image
+  const imageResource = await getFile(
```
timbot1789: This file is returned by the `saveFileInContainer` function. You don't need to refetch it.

src/routes/Profile.jsx
```diff
@@ -0,0 +1,162 @@
...
+  localStorage.setItem('restorePath', location.pathname);
+
+  const [profileName, setProfileName] = useState(profileData?.profileInfo.profileName);
+  const [nickname, setNickname] = useState(profileData?.profileInfo.nickname);
+
+  const [edit, setEdit] = useState(false);
+
+  const loadProfileData = async () => {
+    const profileDataSolid = await fetchProfileInfo(session);
+    localStorage.setItem('restoreProfileData', JSON.stringify(profileDataSolid));
```
timbot1789: You don't ever read this item. You can remove it.

src/components/Profile/ProfileImageField.jsx
```diff
@@ -0,0 +1,109 @@
...
+  const { session } = useSession();
+  const { profileData, fetchProfileInfo, removeProfileImage, uploadProfileImage } =
+    useContext(SignedInUserContext);
+  const [profileImg, setProfileImg] = useState(localStorage.getItem('profileImage'));
+
+  const handleProfileImage = async (event) => {
+    await uploadProfileImage(session, profileData, event.target.files[0]);
+
+    const updatedProfileData = await fetchProfileInfo(session);
+    localStorage.setItem('profileImage', updatedProfileData.profileInfo.profileImage);
```
timbot1789, line 48: Is there a reason you're caching this in localStorage?

src/model-helpers/Profile.js
leekahung: I'm a bit confused by this. I believe I was following the example from Inrupt for `saveFileInContainer`. Do you mean the portion about the `getSourceUrl`? I don't think I've tried that before. I'll give it a go.

src/model-helpers/Profile.js
leekahung: Using the newly stored `savedFile` from `saveFileInContainer`

src/routes/Profile.jsx
leekahung: Unused item in localStorage is removed.

src/model-helpers/Profile.js
leekahung: Had planned on using `savedFile` since it seems to be the correct object to use for Inrupt's `createAcl`. Unfortunately, it seems to err out.

https://github.com/codeforpdx/PASS/assets/14917816/9459d955-2fca-4756-adec-a8873fe78433

Thinking might be due to `saveFileInContainer` and `getFile` returns different objects and `createAcl` expects the one from `getFile`, but I'm not too sure.

Screen Shot 2023-07-06 at 23 32 03 (2)

For now, I'm returning to using `imageResource`.

src/model-helpers/Profile.js
leekahung: Think I got this updated to using the source URL from `saveFileInContainer` using a new constant `savedFile`. Created a new function called `saveSourceUrlToThing` in session-helper to help with generating the associated URL for the Thing in the TTL file.

timbot1789: Yeah this looks good

src/model-helpers/Profile.js
timbot1789: Well that's very odd, and that error message is not all that helpful. Your current implementation is fine for new then. I guess we can dig into why inrupt's libraries are misbehaving later.

timbot1789 (approved):

xscottxbrownx (approved):

Jared-Krajewski (approved):

jeff-zucker:
Hey, I just happened to see this and haven't look at the full PR or really any of your code.  But  I did see this:

> > I'd recommend you base it off the minimum recommended profile information given in the solid spec: https://github.com/solid/solid-spec/blob/master/solid-webid-profiles.md#minimum-recommended-profile-information

The  github.com/solid/solid-spec/* tree  is **old**, **outdated** versions of the specs.  For current work on the profile, see instead https://github.com/solid/webid-profile/ or ping me, or post a message to https://app.gitter.im/#/room/#solid_webid-profile:gitter.im which will go straight to the editors of the spec who should be happy to answer questions (no matter what level) about current thinking on the profile.

leekahung:
> Hey, I just happened to see this and haven't look at the full PR or really any of your code. But I did see this:
> 
> > > I'd recommend you base it off the minimum recommended profile information given in the solid spec: https://github.com/solid/solid-spec/blob/master/solid-webid-profiles.md#minimum-recommended-profile-information
> 
> The github.com/solid/solid-spec/* tree is **old**, **outdated** versions of the specs. For current work on the profile, see instead https://github.com/solid/webid-profile/ or ping me, or post a message to https://app.gitter.im/#/room/#solid_webid-profile:gitter.im which will go straight to the editors of the spec who should be happy to answer questions (no matter what level) about current thinking on the profile.

Hello @jeff-zucker, thanks for letting us know about the updated specification for the profile. I'll have a look over the document. With regards to a basic set up for a profile, would what we could see currently from solidcommunity.net be a good example of the specification?

jeff-zucker:
Other than structural things like oidcIssuer, preferencesFile, publicTypeIndex, etc, it is unlikely that there will be any required predicates.  Users, for example, are not required to a give a name.  Without the oidcIssuer, the profile won't function. Without the other structural predicates it won't be interoperable with other apps.  The predicates shown in NSS are the common ones.  You can use any other predicates you want.  So NSS is mostly a good example.  Unfortunately ESS keeps the WebID-profile on a non-Solid server that can not be modified by Solid apps - the user must manually make changes to it through an ESS-specific, non-Solid interface.  So the ESS profile is mostly just the issuer and the storage with a pointer to extended documents that are on a Solid server and can be written to by Solid apps.

The profile editor app *must* prevent writing of invalid profiles.   By that I mean syntactically invalid RDF.  If the profile has invalid RDF it can't be opened and read.  So if you validate the RDF before writing and don't mess with the structural predicates, you should be fine and can use or omit other predicates as you wish.

timbot1789:
@jeff-zucker  Thanks for your input! It's definitely been a challenge to find up-to-date specs, so we appreciate you pointing us to the right place.

jeff-zucker:
@timbot1789 - ping me anytime, I want to be available for your great project.

[gitter or email is best, with the volume of github notifications, I may miss it]

PR #276, opened 2023-06-24
Fixing redundant code from issue 218

andycwilliams:
Quick merge to fix my error in removing redundant code. With issue #218, `AddClient.jsx` was refactored into `AddClientModal.jsx` and therefore no longer needed. It was to be deleted but I didn't before going ahead with the merge. This fixes the issue.


leekahung:
While I saw that the earlier PR has been approved, please refrain from merging unless you're sure things have been double-checked on your end.

xscottxbrownx (approved):
Looks good to me.  Just deleting the original `AddClient.jsx` and fixing it's index.js file to match

src/components/Clients/AddClientModal.jsx
leekahung, line 1: For the Modal, since we're already using MUI components instead of `<input>`, could we swap out the useField hooks for givenName and familyName with useStates?

Think we could also remove the disable eslint lines here without eslint complaining.

leekahung (changes requested):
See feedback for requested changes

src/components/Clients/AddClientModal.jsx
leekahung: Any progress in this front? Think it'll be a good update to have while we're having this patched up.

andycwilliams: Been extremely busy with work. Will try to get back in the PASS rhythm now.

leekahung: No rush.

leekahung (approved):
I think this should be fine for merging.

PR #277, opened 2023-06-25
Disable Upload button on certain conditions

### LINKED ISSUE #268: Upload Document - start with "upload file" button disabled
## Issue:

In `UploadDocumentForm.jsx`, the `Upload File` button should remain disabled until all necessary fields of form are filled out AND a file has been chosen.

Currently, a user can click `Upload File` button as soon as a file type is selected from the dropdown, and the error displayed doesn't make much sense in this case.

---

## Proposed solution:

some sort of state that controls the `disabled` prop of this button, and state change based on validation.

timbot1789:
Disable upload button in the upload documents form if:

- There is no document to upload
- A document is currently being uploaded

This also means we can remove a notification about no document existing.

### DIFF
--- a/src/components/Form/UploadDocumentForm.jsx
+++ b/src/components/Form/UploadDocumentForm.jsx
@@ -36,6 +36,7 @@ const UploadDocumentForm = () => {
   const [verifyFile, setVerifyFile] = useState(false);
   const [file, setFile] = useState(null);
   const [inputKey, setInputKey] = useState(false);
+  const [uploading, setUploading] = useState(false);
   const { addDocument, replaceDocument } = useContext(DocumentListContext);
 
   const handleDocType = (event) => {
@@ -56,20 +57,13 @@ const UploadDocumentForm = () => {
   const handleDocUpload = async (e) => {
     e.preventDefault();
 
-    if (!docType) {
-      runNotification('Upload failed. Reason: No document type selected.', 5, state, dispatch);
-      setTimeout(() => {
-        dispatch({ type: 'CLEAR_PROCESSING' });
-      }, 3000);
-      return;
-    }
-
     const fileDesc = {
       name: file.name,
       type: docType,
       date: expireDate,
       description: docDescription
     };
+    setUploading(true);
     runNotification(`Uploading "${file.name}" to Solid...`, 3, state, dispatch);
 
     try {
@@ -91,6 +85,7 @@ const UploadDocumentForm = () => {
       }
     } finally {
       clearInputFields();
+      setUploading(false);
     }
   };
 
@@ -171,7 +166,7 @@ const UploadDocumentForm = () => {
           </FormControl>
           <br />
           <FormControl fullWidth>
-            <Button variant="contained" disabled={state.processing} type="submit" color="primary">
+            <Button variant="contained" disabled={uploading || !file} type="submit" color="primary">
               Upload file
             </Button>
           </FormControl>
--- a/src/components/Form/UploadDocumentForm.jsx
+++ b/src/components/Form/UploadDocumentForm.jsx
@@ -36,7 +36,6 @@ const UploadDocumentForm = () => {
   const [verifyFile, setVerifyFile] = useState(false);
   const [file, setFile] = useState(null);
   const [inputKey, setInputKey] = useState(false);
-  const [uploading, setUploading] = useState(false);
   const { addDocument, replaceDocument } = useContext(DocumentListContext);
 
   const handleDocType = (event) => {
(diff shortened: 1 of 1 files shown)


src/components/Form/UploadDocumentForm.jsx
```diff
@@ -56,20 +57,13 @@ const UploadDocumentForm = () => {
   const handleDocUpload = async (e) => {
     e.preventDefault();
 
-    if (!docType) {
```
timbot1789, line 59: Other recent changes (not in this PR) block form submit if this  field isn't filled out, so we can't hit this code branch. Removing.

xscottxbrownx:
This looks like it should work, but cannot test. Login is not working.  Solid servers probably having issues again, will check in morning.   Need to run prettier formatting to pass test.

src/components/Form/UploadDocumentForm.jsx
```diff
@@ -171,7 +166,7 @@ const UploadDocumentForm = () => {
           </FormControl>
           <br />
           <FormControl fullWidth>
-            <Button variant="contained" disabled={state.processing} type="submit" color="primary">
+            <Button variant="contained" disabled={uploading || !file} type="submit" color="primary">
```
leekahung: I believe the `!file` condition is enough to trigger disabled?

xscottxbrownx: only when there's no file. While it's uploading that file it stays active until it's complete.  so the uploading state is necessary to trigger disabled while that is happening, it's replacing what used to be state.processing.

leekahung: I'm a but confused here.

Disabled is being triggered only when there's a file in the new state. Which makes sense. Condition doesn't clear until the file is cleared, which is good too since it's only triggered at the end of handler function or when clear forms are clicked.

So doesn't that make uploading redundant?

timbot1789: No, a file in the new state does not disable the upload button. I'm not actually sure of any value that corresponds to a `new file` state. Upload is disabled in two cases:

1. The user has not attached a file to the form (`file` is null or undefined)
2. The form is currently performing an upload (`uploading` is true)

As long as both those cases are false (i.e. there is a file attached to the form, and the form is not currently uploading), the upload button is enabled.

xscottxbrownx: ^ yep... The button should not stay enabled while the file is being uploaded, just allows the user to click it more times.  It should be disabled while uploading.    The issue is that the `file` state is still true while it is uploading, so it requires both conditions (`!file` and `uploading`) to solve.  That's what I was trying to describe in our DM as well.

leekahung: That should be fine then, I thought we're replacing state.processing completely.

I'm just curious now as to why we couldn't just use state.processing then with the dispatch functions for set and clear processing.

So instead of a new `setUploading(true)` in Line 66, we could use the existing `dispatch({ type: "SET_PROCESSING" })` and instead of using `setUploading(false)` in Line 88, we could use `dispatch({ type: "CLEAR_PROCESSING"})` and use `disabled={state.processing || !file}`

This should produce the same effect.

xscottxbrownx: have you tested?  It should, but I was having trouble with it. Like the state.processing stopped working or something. I was going to ask about it too.  I could never get state.processing to be true anymore.

leekahung: Yeah, I've tested it out using those, see clips on how `state.processing` gets toggled. The location of the dispatch functions are where the `setUploading(true)` and `setUploading(false)` use to be.

Clip to mimic long processing time:

https://github.com/codeforpdx/PASS/assets/14917816/0dd32ef9-0c44-4ea9-bb5d-dcc0a3f1dcc0

Clip without custom delay function:

https://github.com/codeforpdx/PASS/assets/14917816/f5e89c13-6305-4717-b7ef-4a0f88caa212

Also see clip with `state.processing || !file` which sets the condition for `disabled` below. As shown, it remains `true` for the most part. The only time it's `false` was when `!file` was `false` after choosing a file and before `upload file` was clicked.

https://github.com/codeforpdx/PASS/assets/14917816/6f0fba51-8e91-4168-9c4c-03fa179fd64e

timbot1789: I'll give this method a shot and update.

timbot1789:
> This looks like it should work, but cannot test. Login is not working. Solid servers probably having issues again, will check in morning. Need to run prettier formatting to pass test.

@xscottxbrownx  Yeah solidcommunity.net seems to have some pretty major performance issues. It may be easier to test on opencommons.net, or run the dev server on localhost.

src/components/Form/UploadDocumentForm.jsx
timbot1789: Updated this to use notification dispatch instead. As far as I can tell, though, the notification is global state, yes? So we're only able to set one `SET_PROCESSING` value for the whole system? That seems like an odd limitation.

leekahung (approved):
Have tested this out locally. Think it'll be fine for merging.

PR #281, opened 2023-06-27
Fix client refresh and url redirection

### LINKED ISSUE #279: PASS URLs should not contain # symbol
**Describe the bug**
Currently, PASS URLS that require a login to access include a `#` in their path. However, this makes them invalid targets of redirect in certain libraries. The `#` in a URI is used to indicate a [Fragment Identifier](https://www.w3.org/Addressing/URL/4_URI_Recommentations.html) , a certain place within a page. It should not be used to identify a page itself.

**Expected behavior**
We can update the URL paths to remove  to make them simpler. Ideally we should be able to just have things like:

- www.pass.com
- www.pass.com/login
- www.pass.com/clients
- www.pass.com/documents
- www.pass.com/messages

timbot1789:
The goal of this PR is to remove the `#` parameter in the PASS url so we are able to use server-side redirects. These redirects will make it easier to have multiple different login flows (e.g. existing vs new user). Along the way we rebuilt and simplified the login and routing flow of the app.

I have a production build of this running up at https://279-remove-hash-from-url--splendorous-nougat-3f0062.netlify.app 

That should let you test prod behavior without difficulty.

A few changes in this PR:

1. Removed `@inrupt/solid-ui-react` and added the components we use as native code. We really only use this library as a convenient way to manage the session provided by `solid-client-authn-browser`. However, that auth library is fairly buggy, and we need to implement a few workarounds to get it running properly. Having it hidden behind a react library makes that difficult.
2. Updates `solid-client-authn-browser` to get a fix for `restorePreviousSession`. This allows the solid client to handle session restoration for us, and makes the silent authentication that happens on page refresh faster. This also made some code redundant, so I removed that. Less code ≈ fewer bugs.
3. Added code in `vite.config.js` to force vite to build `solid-client` as a ES module instead of a CommonJS bundle. This is a workaround for this bug: https://github.com/inrupt/solid-client-authn-js/issues/3001
4. Renamed the `routes` folder to `pages` to keep it consistent with how the files are referred to internally.
5. Removes the hash router and replaces it with browser router. This is to enable server redirection and other features used by PASS login. The main use case for this right now is in user registration. To register a user, we need them to log into a pod. That requires sending a redirect URL to the pod server. If the user registration page is hosted on `www.pass.com/PASS/#/signup`, the redirect link could be `www.pass.com/PASS/#/signup`, and when we get back to that page with a valid login, we continue with any  user registration flow. However, URLs with a `#` in them are not valid redirect links. Sending such a link to the server will cause an error. If we instead put signup at `www.pass.com/signup`, we can redirect to it without issue after we've finished logging into the pod.
 
NOTE: This will break github code pages, but we can deploy the project to another site without too much difficulty, We will need to do that at some point anyway.

6. Updated routes so that they don't have `#` in them anymore. Routes in this PR are:
  - `www.pass.com` - login/landing page
  - `www.pass.com/clients`
  - `www.pass.com/documents`
  - `www.pass.com/messages`
  - `www.pass.com/profile`
7. Added mappings of a lot of the `src` modules in `vite.config.js` so they have absolute references. This makes it easier to refer to them throughout the project, since they will always have the same path no matter what folder you're in.
 ---
No new unit tests because this doesn't implement a new feature, just refactors. The fact that there's no regression in the unit tests should demonstrate that the refactor is working correctly.

We could add some unit tests for routes, but that can be in a future PR.


src/pages/Documents.jsx
```diff
@@ -0,0 +1,68 @@
+// React Imports
```
timbot1789, line 1: This document wasn't created. It was moved from /routes/Document.jsx. Only new feature is the layout wrapping it.

test/components/NavBar/__snapshots__/OidcLoginComponent.test.jsx.snap
```diff
@@ -33,6 +33,7 @@ exports[`renders correctly 1`] = `
     class="MuiBox-root css-120gsus"
   />
   <div>
+    http://localhost:3000/PASS/
```
timbot1789, line 36: We now pass the redirect URL in explicitly, causing it to render correctly in the test (it wasn't before, and I didn't notice).

test/pages/__snapshots__/Home.test.jsx.snap
```diff
@@ -0,0 +1,280 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
```
timbot1789, line 1: Moved, not created

src/components/NavBar/OidcLoginComponent.jsx
```diff
@@ -4,14 +4,12 @@ import React, { useState } from 'react';
...
 import { TextField, Box, Button } from '@mui/material';
-// Custom Hook Imports
-import { useRedirectUrl } from '../../hooks';
 // Constants Imports
 import { ENV } from '../../constants';
 
 const OidcLoginComponent = () => {
-  const [oidcIssuer, setOidcIssuer] = useState(ENV.VITE_SOLID_IDENTITY_PROVIDER);
-  const redirectUrl = useRedirectUrl();
+  const defaultOidc = ENV.VITE_SOLID_IDENTITY_PROVIDER || '';
```
timbot1789, line 12: We were getting a controlled component error in dev console if the ENV didn't set this value. This guard cleans up the error.

test/components/NavBar/__snapshots__/OidcLoginComponent.test.jsx.snap
```diff
@@ -82,6 +83,7 @@ exports[`sets OIDC provider on login 1`] = `
   />
   <div>
     oidc.provider.url
+    http://localhost:3000
```
timbot1789: We now pass the redirect URL in explicitly, causing it to render correctly in the test (it wasn't before, and I didn't notice).

xscottxbrownx:
I haven't looked at this at all yet, but if it does anything to help the redirects or logged in ? check - then I'm all for it. Constantly get frustrated with getting stuck at splash/intro page of PASS and nothing happens when clicking `login`.    I mean I never log out of solid, it should be able to see that I am logged into solid and bypass the intro screen altogether I would think.

timbot1789:
@xscottxbrownx correct, we are able to do that right now. If you leave the site without logging out and then later return (within a certain amount of time), you will automatically log back in. We've actually had that feature for a while, but this fixes some things and makes the refresh a bit faster.

The more significant change is the removal of the hash router.

I think I've updated the PR description since your comment to make things a bit clearer.

leekahung:
Hey @timbot1789, just want to check in on this PR. After this updating this, are there plans to migrate hosting services from GitHub Pages?

If we are, I think I know of a good alternative. I've been using Netlify before for some of my personal projects and it seems to work great with BrowserRouter and integrates well with GitHub repositories.

We just need to take into account Netlify specific files like `netlify.toml` for setting things up, but it should be relatively simple from my experience with that platform. Let me know what you think. Think we could speak with Hugh about getting Netlify integrated to our project if we decide to pursue this route.

src/App.jsx
```diff
@@ -24,18 +23,18 @@ import AppRoutes from './AppRoutes';
...
-    <SessionProvider>
-      <CssBaseline />
-      <ThemeProvider theme={theme}>
-        <Layout ariaLabel="Home Page">
-          <UserDataContextProvider>
+  <SessionProvider restorePreviousSession>
+    <CssBaseline />
+    <ThemeProvider theme={theme}>
+      <UserDataContextProvider>
+        <BrowserRouter>
```
leekahung, line 29: Just a bit curious, is there a reason for moving the Router from the uppermost later to just before Layout compared to before where the Router is wrapping around everything?

timbot1789: No strong reason, just thought it looked cleaner. Based the idea off of bulletproof react:

https://github.com/alan2207/bulletproof-react/tree/master

Which just contains a lot of good react programming patterns, and does the routes and provider structure that way.

timbot1789:
> Hey @timbot1789, just want to check in on this PR. After this updating this, are there plans to migrate hosting services from GitHub Pages?
> 
> If we are, I think I know of a good alternative. I've been using Netlify before for some of my personal projects and it seems to work great with BrowserRouter and integrates well with GitHub repositories.
> 
> We just need to take into account Netlify specific files like `netlify.toml` for setting things up, but it should be relatively simple from my experience with that platform. Let me know what you think. Think we could speak with Hugh about getting Netlify integrated to our project if we decide to pursue this route.

Yeah I've been looking at a few different hosting services. Whatever hosting service we go with (or if we need one at all) is something we can talk about in a dev meeting. Netlify is definitely a good option.

Haven't been pushing this because there's a bug in the PR now that I can't track down. The routing works when I run the dev server, but fails when I do the production build and preview it with `npm run preview`. It seems to be an issue with the inrupt auth library, but not 100% sure.

timbot1789:
@leekahung I did some more research, and to clarify, the delay in the PR is due to this bug in the `solid-client-authn-browser`: https://github.com/inrupt/solid-client-authn-js/issues/3001

Basically, the most recent versions of the authn library aren't compatible with vite, and break when passed through `vite build`. I'll investigate some workarounds, but really it would need to be fixed in the library itself.

public/_redirects
```diff
@@ -0,0 +1 @@
+/* /index.html 200
```
timbot1789: This is a routing config for netlify. Without this, reroutes don't work properly when the code is deployed there.

src/index.jsx
```diff
@@ -8,8 +8,4 @@ import App from './App';
 import './style.css';
 
 const root = ReactDOM.createRoot(document.getElementById('root'));
-root.render(
-  <React.StrictMode>
```
leekahung: I think we can keep `React.StrictMode` to the build. Think, it'll be useful to have it check for any unexpected side-effects during development. Shouldn't affect production.

timbot1789: Woops forgot I removed that. I'll add it back in.

timbot1789: Fixed

netlify.toml
```diff
@@ -0,0 +1,4 @@
+[[redirects]]
```
timbot1789: This is a netlify config file. The `[[redirects]]` settings tell the netlify server to let react handle the routing.

test/components/NavBar/__snapshots__/OidcLoginComponent.test.jsx.snap
```diff
@@ -1,108 +0,0 @@
-// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
```
timbot1789, line 1: As we are using material UI more extensively now, these snapshots don't seem to be as useful in testing as I expected them to be. The way material UI writes CSS (composing a series of simple classes) means very small changes to components results in extreme, indecipherable changes to the snapshot. Both Material UI and react testing library also advise against snapshot testing, so I'm removing this snapshot and just using react testing library to look at the elements we're interested in directly.

netlify.toml
leekahung, line 1: I'm not sure if it's needed, but should we have a [build] configuration for the build command with Vite like with `npm run vite:build`?

Also for environment variables, I think there's a way to establish what we have here over to Netlify.

vite.config.js
```diff
@@ -1,17 +1,40 @@
...
   plugins: [
     react(),
     visualizer({
-      filename: 'bundleStats.html',
-      open: true
+      filename: 'bundleStats.html'
     })
   ],
+  resolve:{
+    alias:[
```
leekahung, line 26: I do like we're going to swap over to using alias for some of our imports. Helps cuts down the lines of text we might get from typing the relative pathing to certain files/directories.

netlify.toml
timbot1789: We can, but you're right we don't need to. If we don't give it other config, Netlify automatically runs `npm run build` for us and hosts the output. That's all our build process is at the moment. If we ever need a more complex build process than that (which I don't expect we ever will), we can tell netlify to do that here. 

And yes we can add env variables in the netlify terminal, like we do now for code pages. That's the best place to add them

vite.config.js
timbot1789: Yeah, and if we ever need more, we can just add them to the lists here and in eslint.rc. It made editing all the places we used `useSession` a hell of a lot easier cuz I could just do a regex to replace `@inrupt/solid-react-ui` with `@hooks` :)

leekahung: Haha, yeah 😆 

I'm good with this.

leekahung (commented):
I believe I've gone over all the files. The underlying code seems to be working fine for redirects and refreshing.

One thing I've noticed is that the auto-populate for adding webId to clients is gone in the production build (things work as intended on the local build on my machine).

https://github.com/codeforpdx/PASS/assets/14917816/ab03c147-0271-496b-9901-c2692d8f01fc

Should be fixed as soon as we got the environment variables set up in Netlify. Outside of that, I think I'm ready to approve this. Think it's ready for our switch from GitHub Pages to another hosting service.

timbot1789:
@leekahung Sorry, I should clarify about the netlify. That's just a demo environment that I set up to show that the code will work in a certain environment. It's not a final PASS prod environment. I mean to compare it to github pages. If you push this branch to github pages and test it there, it will fail.

That netlify environment isn't PASS'S. It's my own on my own account. It's not even using code taken from the PASS repo. I don't have permission to have netlify deploy off of PASS itself; github blocks my connection request. Instead, I made an identical fork and pushed it up.

If we want to use netlify as a prod environment, we'll need to set it up and configure it outside this PR. I just want to show that we _can_ use it.

leekahung:
Ah okay. No worries. I believe we could get @htharker42 to help us with getting the service set up at least with regards to GitHub integration?

timbot1789:
> Ah okay. No worries. I believe we could get @htharker42 to help us with getting the service set up at least with regards to GitHub integration?

Yeah we can talk about it in a dev meeting

package.json
```diff
@@ -54,6 +53,7 @@
     "last 3 edge versions"
   ],
   "devDependencies": {
+    "@rollup/plugin-node-resolve": "^15.1.0",
```
timbot1789, line 56: These 3 dev dependencies are used to get the `@hooks`, `@components`, etc... mappings working. This one runs on the prod build, the eslint ones run in eslint.

leekahung (approved):
This is fine. I'm okay with merging this.

leekahung:
Ah right, we'll have to adjust to using Heroku after speaking with Hugh. I'm open to reviewing this again after Heroku related configs are set up.

timbot1789:
> Ah right, we'll have to adjust to using Heroku after speaking with Hugh. I'm open to reviewing this again after Heroku related configs are set up.

I'd like to get this merged as soon as possible. I think there are useful changes here that other developers would benefit from having access to. And the current codebase is already beginning to diverge from the code in this PR. We can put up the heroku config in a separate PR once Hugh gets done setting that up. It's not super complex. I wrote one when I was looking at Heroku for this PR and it's only a few lines at most. If Hugh doesn't get back to us before the next time we want to deploy to prod, I can just put a fork of the project up on my own heroku account and we can use that for a little while.

leekahung:
> > Ah right, we'll have to adjust to using Heroku after speaking with Hugh. I'm open to reviewing this again after Heroku related configs are set up.
> 
> I'd like to get this merged as soon as possible. I think there are useful changes here that other developers would benefit from having access to. And the current codebase is already beginning to diverge from the code in this PR. We can put up the heroku config in a separate PR once Hugh gets done setting that up. It's not super complex. I wrote one when I was looking at Heroku for this PR and it's only a few lines at most. If Hugh doesn't get back to us before the next time we want to deploy to prod, I can just put a fork of the project up on my own heroku account and we can use that for a little while.

I'm good to merge as is now if you think it's ready. If there's other devs you think should need to review the code, I'll wait for their approval first.

xscottxbrownx (approved):
Went over this PR in discord VC - I had already looked over most files, but was a little hung up on the SessionContext and useSession hook, but good now.

leekahung (approved):

PR #284, opened 2023-06-28
Issue 226/reply messages

### LINKED ISSUE #212: Expand the fields for Message TTL file
As detailed from Tim in https://github.com/codeforpdx/PASS/pull/191#discussion_r1217249834 there could be other fields that would be useful:

- `delivery-status`: a string representing if the message has been delivered. e.g. pending, delivered, failed to send, refused, etc.
- `message-id`: a string containing a universally unique identifier. Will allow us to tell if a message in one person's inbox is the same as a message in another person's outbox. Can also be used to check if a message was successfully delivered. If a message with that message-id isn't in the inbox, it wasn't delivered
- `in-reply-to`: a message-id of a different message. Allows us to implement message threading and conversations

Implement them inside buildMessageTTL in session-helper.

### LINKED ISSUE #226: Add conversations/"reply to" for messages
Now that we have a working message system with storage, we could think creating a working reply system as oppose to creating a new message.

Use existing message TTL file from recipient inbox as a starting point and append the TTL before sending message back to sender.

leekahung:
## PR Details

This PR is mainly about creating a reply system for PASS using the messages found inside inbox/outbox.

https://github.com/codeforpdx/PASS/assets/14917816/d007a177-eebd-49e9-9d4d-3612f49a6c98

A new reply button will appear after you preview the message. The `<NewMessage>` component, the `buildMessageTTL` function, and the `sendMessageTTL` function have been modified to handle sending replies. In a reply, the fields for title, recipient, and message will be pre-populated. A new message added in the message body like a typical email.

Fields for the message file's location in Pod (`podUrl`) and a unique id (`identifier`) has been incorporated into the metadata as a new Thing (`#messageid`) with `parseMessageTTL` and `buildMessageTTL` being updated to take account for the change.

The reply functionality utilizes the fields created here and creates a new Thing into the metadata `#replyTo` on top what what's normally send in a message. In a reply `#replyTo` references the original message ttl via their unique id.

One new change to `<NewMessage>` is that it utilizes pod URLs instead of username. This would be analogous to using a email.

The "Reply" button is also relegated to just the "Inbox" in newer commits.

There could be times when the user doesn't have a name associated with their Profile (for example opencommons.net  doesn't come with names initially) so the value would often come back as `null` for the sender's name. As a fallback, if the user doesn't have a name associated with their Profile, the pod URL will be used instead as the name. The same was also done for the recipient name, though this value remains under the hood.

## Future Considerations

- The original message portion of the new messages would need to be expanded upon since it could have a very long body. How it's set up now is prone to potential visual bugs.
- UI of new messages in general
- How to populate previous messages using newly created linked list


xscottxbrownx:
I feel like the reply button should only show if in `Inbox`, not `Outbox`.

leekahung:
> I feel like the reply button should only show if in `Inbox`, not `Outbox`.
> 

Think I could swap to just having show on Inbox

xscottxbrownx:
Did you change how the "to:" field works?  I could've sworn I used to be able to just type in the username/webID (like: `scottxbrown`.)

It's not really the focus of this PR, but want to add a "No messages found" message (like in `Clients` and `Documents`) that displays when there are none?

leekahung:
> Did you change how the "to:" field works? I could've sworn I used to be able to just type in the username/webID (like: `scottxbrown`.)

That is correct, the previous iteration utilizes usernames. In this PR, moved over to using pod URLs as stated in the opening post:

> One new change to <NewMessage> is that it utilizes pod URLs instead of username. This would be analogous to using a email.

So `scottxbrown` would now be `https://scottxbrown.solidcommunity.net/`. It also let's users send message to PASS users of different Solid Providers potentially.

xscottxbrownx (changes requested):
## This is what I found in components.
If I can get a little help with `typedefs` and `utils` I'd be grateful.

---

### Add an `<EmptyListNotification type="messages" />` if no messages (like in `DocumentTable` and `ClientList`.)  
- Might have to adjust `str3` in `EmptyListNotification` or maybe line 31 of it needs to be conditional to make sense with "messages".

---

### `Messages.jsx`: Make the icon in the button.  
- Use `startIcon={<CreateIcon />}`. 
- Don't have to worry about the margin that way.

---

### `Messages.jsx` line 41: `handleMessageRefresh` function... 
- I'm not seeing the loading messages anymore after clicking refresh. Did something change with `MessageContext` or the `loadMessages` from it?

xscottxbrownx:
Screenshot 2023-07-10 at 1 48 37 PM
I also received this error...

leekahung:
> Screenshot 2023-07-10 at 1 48 37 PM I also received this error...

That's strange. I don't think I've seen that error before when using solidcomunity.net. Would like to reproduce it on my end to see where the problem could be. Could you detail what has been done?

xscottxbrownx:
> > Screenshot 2023-07-10 at 1 48 37 PM I also received this error...
> 
> That's strange. I don't think I've seen that error before when using solidcomunity.net. Would like to reproduce it on my end to see where the problem could be. Could you detail what has been done?

Nothing, that was while on your branch...

Going to Development I get this:
Screenshot 2023-07-10 at 2 20 08 PM
But yet, it still sent - it shows in the inbox and the outbox
Screenshot 2023-07-10 at 2 20 15 PM
Screenshot 2023-07-10 at 2 20 20 PM
But if I do `https://scottxbrown.solidcommunity.net/` instead (the last character being `/`) - it gives same error message, but doesn't send.

Here's profile:
Screenshot 2023-07-10 at 2 19 32 PM

leekahung:
> ### Add an `<EmptyListNotification type="messages" />` if no messages (like in `DocumentTable` and `ClientList`.)
> * Might have to adjust `str3` in `EmptyListNotification` or maybe line 31 of it needs to be conditional to make sense with "messages".

Got this updated to using that notification. Also tweaked `str3`.

> ### `Messages.jsx`: Make the icon in the button.
> * Use `startIcon={<CreateIcon />}`.
> * Don't have to worry about the margin that way.

Fixed.

> ### `Messages.jsx` line 41: `handleMessageRefresh` function...
> * I'm not seeing the loading messages anymore after clicking refresh. Did something change with `MessageContext` or the `loadMessages` from it?

I'm not too sure happening here. Neither of those has been touched in this PR.

leekahung:
Still a bit unclear as to what's happening with regard to those errors, but it seems to be running okay from testing.

https://github.com/codeforpdx/PASS/assets/14917816/a90eacf9-66e3-496f-9374-48399471a245

leekahung:
> Going to Development I get this: Screenshot 2023-07-10 at 2 20 08 PM

Oh, this is because Development is still using usernames. Only this branch `issue-226/reply-messages` started to use straight up pod URLs as mentioned before.

xscottxbrownx (changes requested):
All requests have been updated/fixed.  Looks good!

---

I accidentally hit `request changes` - approving now

xscottxbrownx (approved):

timbot1789 (changes requested):
I like a lot of what's in here, but there's one main issue. It appears that the replyTo is working by appending the whole text of the conversation up to that point into the body of the new email. Here's a single message I was able to generate:
```
<#message> <https://schema.org/uploadDate> "2023-07-11T01:24:38.970Z"^^<http://www.w3.org/2001/XMLSchema#dateTime>;
    <https://schema.org/title> "RE:RE:RE:RE:RE:RE:RE:testing";
    <https://schema.org/Message> ">>>>>>>hi self \n>>>>>>\n>>>>>>Hi back to myself \n>>>>>let's start a chat chain\n>>>>>\n>>>>> \n>>>>\n>>>>Indeed, we should start a really long chat chain \n>>>\n>>>How long of a chat chain should we make? \n>>As long as we can! Who knows what could happen? I could dump a whole file of content in here.\n>>\n>> \n>Wow we can make a chain that long?\n> \n\nOf course! Message chains can last for hundreds, even thousands of messages".
<#messageid> <http://schema.org/identifier> "679b9658-3d4d-4bcb-8c33-3f2fc3e542d8";
    <http://schema.org/url> <http://localhost:3000/timbot/PASS/Inbox/localhost:3000-20230711-012438.ttl>.
<#sender> <https://schema.org/sender> "timstanden";
    <http://schema.org/url> <http://localhost:3000/timbot/profile/card#me>.
<#recipient> <https://schema.org/recipient> "timstanden";
    <http://schema.org/url> <http://localhost:3000/timbot/profile/card#me>.
<#replyTo> <http://schema.org/identifier> "105b655b-cf8b-4bbb-87c2-f963e761719e";
    <http://schema.org/url> <http://localhost:3000/timbot/PASS/Inbox/localhost:3000-20230711-012408.ttl>.
```

This won't scale beyond a couple dozen messages. What we should be doing here instead is have each message store just the content of that one message, then use the `inReplyTo` field as a pointer to the previous message. That will allow us to create a linked list from which we can rebuild the conversation.

You could think of the data structure as looking a little like this:
image

src/utils/network/session-helper.js
```diff
@@ -405,45 +412,56 @@ export const parseMessageTTL = (messageTTLThing) => {
...
+  [newMessageTTL, newMessageID, senderInfo, recipientInfo].forEach((thing) => {
     newSolidDataset = setThing(newSolidDataset, thing);
   });
 
+  let replyTo;
+
+  if (messageObject.inReplyTo) {
+    replyTo = buildThing(createThing({ name: 'replyTo' }))
+      .addStringNoLocale(RDF_PREDICATES.identifier, messageObject.inReplyTo)
+      .addUrl(RDF_PREDICATES.url, messageObject.messageUrl)
```
leekahung: Depending on sender or recipient, this is not correctly giving the right url to the previous message. Needs to be updated

leekahung (commented):
See things that needs changing

src/utils/network/session-helper.js
leekahung: Got this fixed.

leekahung: Got this fixed.

leekahung:
> What we should be doing here instead is have each message store just the content of that one message, then use the `inReplyTo` field as a pointer to the previous message. That will allow us to create a linked list from which we can rebuild the conversation.

Hey @timbot1789,  I think I got the changes made so that it'll only point to the previous message url and store content of just that message.

```
<#message> <https://schema.org/uploadDate> "2023-07-11T05:11:42.412Z"^^<http://www.w3.org/2001/XMLSchema#dateTime>;
    <https://schema.org/title> "RE:New message";
    <https://schema.org/Message> "Test 2".
<#messageid> <http://schema.org/identifier> "e0192267-f999-48d5-a288-f0aeef270540";
    <http://schema.org/url> <https://kahunglee.solidcommunity.net/PASS/Inbox/newtestuser-20230711-051142.ttl>.
<#sender> <https://schema.org/sender> "Alice";
    <http://schema.org/url> <https://newtestuser.solidcommunity.net/profile/card#me>.
<#recipient> <https://schema.org/recipient> "Ka Hung Lee";
    <http://schema.org/url> <https://kahunglee.solidcommunity.net/profile/card#me>.
<#replyTo> <http://schema.org/identifier> "cea85857-aac1-4a2b-ad90-96a01f7ec17f";
    <http://schema.org/url> <https://kahunglee.solidcommunity.net/PASS/Outbox/kahunglee-20230711-050528.ttl>.
```

On PASS, the message of the previous message would simply be shown as "original message" on the modal, but wouldn't be used to build the next node for a linked list, only what's typed within the messages field.

Screen Shot 2023-07-10 at 22 25 54 (2)

Also made a few changes to title, so that it'll just return a single "RE:" for all future replies. This is only a temporarily solution to prevent the message title from exploding, but we'll probably need something better? I'm not really sure what at the moment, but was thinking about linking the URL to the raw message TTL file and generating something from that link utilizing the data structure we're building to chain the messages and replies. Not sure.

xscottxbrownx:
I like the single "RE:" - still not sure if this is the exact UI for a reply message to display in inbox, but it is a drastic improvement than first iteration. I still approve and we can discuss/decide in future about what is exactly displayed in the accordion div.  I feel like most users will somehow want to see the linked original message too.

timbot1789:
> I like the single "RE:" - still not sure if this is the exact UI for a reply message to display in inbox, but it is a drastic improvement than first iteration. I still approve and we can discuss/decide in future about what is exactly displayed in the accordion div. I feel like most users will somehow want to see the linked original message too.

I agree with this take. I think it's good to dedicate this PR to just being able to reply to messages at all. We can work in another PR about how best to display these conversations in the UI. We're iterating quickly, so we'll probably need to change any UI we build anyway.

leekahung:
Sure, think we can have that resolved in a future PR.

With regards to recent changes to this PR, do you think there's still some things that needs to be changed/worked on @timbot1789?

timbot1789:
> Sure, think we can have that resolved in a future PR.
> 
> With regards to recent changes to this PR, do you think there's still some things that needs to be changed/worked on @timbot1789?

I'm testing it now. I'll get back to you on that.

timbot1789 (changes requested):
New messages are displaying `null` in the outbox.

image

timbot1789 (commented):
One change. Other than that, this looks all right

leekahung:
> New messages are displaying `null` in the outbox.
> 
> image

I think I can change that, the `null` comes from the fact that the profile of the user that's sending the message does not have a name included in their profile. Think I'll use their username as a fallback value.

src/utils/network/session-core.js
```diff
@@ -163,7 +163,8 @@ export const sendMessageTTL = async (session, messageObject, podUrl) => {
   let recipientName;
 
   try {
-    recipientName = await getUserProfileName(session, recipientWebId);
+    const recipientUsername = recipientPodUrl.split('/')[2].split('.')[0];
```
timbot1789: I actually didn't look at this part of the code too closely, but looking at it now, I find it kind of confusing. 

We take in a pod url where we will be sending the message to, which makes it kind of analogous to an email address. Then we attempt to find a username associated with that pod url. If we can't find a username, we error, saying we can't send the message.

Except we can send the message. We don't use the username to send, we use the pod url. Whether the username is correct or not is irrelevant to the pod url.

I feel like we can simplify this lines 163-170 to just:
```
const recipientName = (await getUserProfileName(session, recipientWebId)) 
```

Then in `MessagePreview.jsx` we can just have the message display an unknown sender message if `message.sender` is `null`. Line 41 of `MessagePreview.jsx` can just be `{message.sender || 'unknown sender'} - {message.title || 'no title'}`

leekahung: Got this updated to using pod URLs as fallback values if profile name does not exist.

As for `message.title`, we previously had this as a required field to fill out so there wouldn't be a message with no title at this time. Think we adjust the validation for the `NewMessage` field to allow for empty titles in another PR.

timbot1789: This looks good to me now

timbot1789 (approved):
Think we got everything cleaned up here. Nice work!

PR #285, opened 2023-07-01
Issue 269/upload doc modal

### LINKED ISSUE #269: Move Upload Document into button/modal combo
## Issue:

Currently, `UploadDocumentForm.jsx` is at top of `Forms` page/route. 

https://github.com/codeforpdx/PASS/assets/71395931/9cdaaafa-ea16-4ac2-8194-33e4bf8e9da7

---

## Proposed Solution:

Top of this page/route should be a button to upload a new document, which would open `UploadDocumentForm.jsx` in a modal.  This way the user's eyes are actually drawn to their documents that render in the table.

andycwilliams:
Refactoring `UploadDocumentForm.jsx` into a modal (`UploadDocumentModal.jsx`)

Simple enough task but I am making this a draft because before creating a full PR, I wanted to discuss styling - namely the buttons. 

Do we want this modal to more closely match `AddClientModal.jsx`? We could change the Clear Form variant to `outlined` rather than `contained` and arrange at least two of the buttons into a grid, for instance. This could be done by wrapping the buttons with `<DialogActions>`, as with `AddClientModal.jsx`.

I was also wondering whether we want to change Clear Form since it already looks similar to the Cancel button in `AddClientModal.jsx` (including the same icon). They serve different functions, which might be slightly counterintuitive for the user. 

Will we also want a Cancel button in this modal? That seems like it could be too many buttons for a single modal.

Additionally, the icons are on different sides of the text. Not sure if that is a concern at all.

`UploadDocumentModal.jsx`:             |  `AddClientModal.jsx`:
:-------------------------:|:-------------------------:
  |  

Closes #269


xscottxbrownx:
Good thoughts

---

### Personally, I think there should be consistency in styling.

---

### My first thought is the buttons. Maybe try:
- `Clear Form` -  first button underneath the form. Full width as it is, same color, contained...
- `Choose File` - next under it. Full width as it is, contained, same color (or actually maybe the tertiary color)...
- Below those two, have 2 side by side buttons:
    - The `cancel` on the left - outlined like usual. 
    - The `upload file` on the right - contained like usual.

---

### Second thought is the icon placement. 
I think this should also be consistent as well.  I actually do think the `choose file` and `upload document` buttons and icons look good on the left, but feel the `clear`, `cancel`, and `add` look off being on left - for some reason my eyes like them on the right.

So:
**1)** move all icons to right
**OR**
**2)** have icons in form buttons on left, and icons in form submit/cancel buttons on right. 

As long as we are consistent with one of these, I think it'll look good and be fine for UX.

---

### Third thought is the `clear form` icon.
Could try the `BackspaceOutlined` icon instead for the `clear form` if you'd like it to be different from `cancel`?
See what that looks like.

---

### Fourth thought is button and title of modal.
I think we should change to "Add Document" instead of upload. 
Again, consistency with `clients` page and workflow/design.

andycwilliams:
Alright, trying out some of Scott's ideas. Here are my thoughts:

- Should the modal be moved to the Documents folder instead of Form?

- Maybe change all form Cancel buttons to a simple X in the upper right? That could clean up the interface. Four buttons so close together could be a bit "busy".

- Is the helper text for the Expiration Date  (`<FormHelperText>MM/DD/YYYY</FormHelperText>`) necessary? That same text already appears when this input is selected, so it may be redundant.

- Trying out consolidating the `<FormControl>` components into one, rather than wrapping each input individually. In my testing, it hasn't interfered with functionality.

- Not sure if I like using the tertiary color for Choose File. It makes the whole form look a little too much like ketchup and mustard. Not sure what I prefer instead, though.

- Also not sure about icon placement. I also think sometimes they look better on the left side rather than the right, but when there are both styles in the same form, it looks a bit strange.

- Decided for now not to rename the file. Would like to confirm that decision before altering the file structure.

- Outside of this scope, but it made me think: Should we add a column for verified status in the document table?

xscottxbrownx:
## Scott's thoughts

> Alright, trying out some of Scott's ideas. Here are my thoughts:
> 
> * Should the modal be moved to the Documents folder instead of Form?

yeah, not sure... but some of out components are organized by what they do (`Notifications`, `Table`, etc) and some are organized by what page they deal with (`Clients`, `Documents`, `Messages`, etc) - so I feel like we should discuss a folder structure pattern to stick to.   
I would prefer `Form` didn't exist and it all lived in `Documents` as we changed it.   
I also think Ka Hung had an idea of making a `Modals` folder and putting all pop-up modals in there (similar to `Notifications`.  
Personally, I really like the folders by what page they deal with, then we can always make subfolders within those if needed. But I'm up for whatever as long as it's consistent.

---

> * Maybe change all form Cancel buttons to a simple X in the upper right? That could clean up the interface. Four buttons so close together could be a bit "busy".

This would also be fine with me. Pretty common UI.   
In the case of this particular modal, I was going to say that I'm not sure a `Clear Form` button is actually needed. User could always `cancel` then go back to `upload/add document`. 

---

> * Is the helper text for the Expiration Date  (`<FormHelperText>MM/DD/YYYY</FormHelperText>`) necessary? That same text already appears when this input is selected, so it may be redundant.

Good point, don't think it's needed. As you said, it shows when user clicks in the input. And if they clicked directly on the date picker, it loads in that format. So not needed.

---

> * Trying out consolidating the `<FormControl>` components into one, rather than wrapping each input individually. In my testing, it hasn't interfered with functionality.

Yeah, I've yet to find an explanation of this exactly in the docs. From what I could find, it seems `<FormControl>` may only be needed on inputs that are NOT `<TextField>`.  It seems to link the focus and other states of the label, input, helper text, and validation together - but components that have all that already (like `<TextField>` don't need it.)  
But this example (if you click to expand the code), shows it for each input/label combo: https://mui.com/material-ui/react-text-field/#input-adornments .  And also this part from: https://mui.com/material-ui/api/form-control/  makes me think one per input as well:
Screenshot 2023-07-02 at 7 06 21 AM

---

> * Not sure if I like using the tertiary color for Choose File. It makes the whole form look a little too much like ketchup and mustard. Not sure what I prefer instead, though.

yeah, looks like a stop light. My vote is that `clear form` isn't needed honestly. User can click `cancel` and go back into it. Which of course isn't the best experience, but 🤷🏻‍♂️
Maybe, make the `choose file` the green as usual (when active), BUT have it have a decent margin bottom to separate it from the `cancel` and `upload` more?

--- 

> * Also not sure about icon placement. I also think sometimes they look better on the left side rather than the right, but when there are both styles in the same form, it looks a bit strange.
> * Decided for now not to rename the file. Would like to confirm that decision before altering the file structure.
> * Outside of this scope, but it made me think: Should we add a column for verified status in the document table?

I think this would be a good idea, once we figure out how and if documents can be legitimately verified.

andycwilliams:
Experimenting with a couple of layout changes. This definitely wouldn't be final. 
- I do like converting Cancel into a simple X, but it would take reworking if we do want to go with it. Might require tinkering with `<FormSection>`. 
- Also replaced the checkbox with a switch. Just to see how it looks.

leekahung:
Looking at the new modal, I think I prefer the previous set up with cancel/upload at the bottom. For the "clear form", think we can do away with it if "cancel" clears the form as well.  For the icons, we could also have it lined up either at the start or end of the button rather than a mix of both.

As for the modal itself, I think we can fit everything without needing to scroll. Looks like there enough room for that, but we'll need to adjust the paddings, margins, and gaps.

leekahung:
Given we have quite the number of modals in our project (like `<LogoutModal>`, `<InactivityMessage>`, `<AddClientModal>`, etc.), was thinking we could start having them moved into a folder called `Modals` inside components. That way we would know where to get the different modals from.

andycwilliams:
More testing.

- Reverted to the previous layout with the rectangular Cancel button design, though I prefer the X
- Widening `FormSection` to make use of more screen space
- Putting document and expiration date selections on the same line
- Moving icons all to the left side of the button
- Other various tweaks to the layout
- Moved modals into a new separate `Modals` folder
- As a side effect, the other forms in the Documents tab are also widened. Those can be reformatted as well, depending on how we intend to implement them in the future. Will they remain separate forms (and perhaps placed into modals) or will they be refactored into switches (similar to how the Delete Document form was turned into an icon button)?

Since we may replace the `Upload status` portion of `FormSection` with a pop-up, that would clear up even more space on each form.

EDIT: Thinking more about it, I don't love the full-width buttons here, at least for the desktop version. I think we could create a breakpoint for small screens where the layout shifts to a more column-like layout. But the desktop version could look cleaner than this, in my opinion.

`UploadDocumentModal.jsx`:             |  `AddClientModal.jsx`:
:-------------------------:|:-------------------------:
 |

andycwilliams:
Even more experiments, just to visualize different designs.
- Added `ButtonGroup` to make the buttons a bit neater/responsive to one another
- Rearranged the order of the buttons to see if it could be more intuitive. Needs further discussion
- Added a breakpoint for mobile screens that rearranges document selection and expiration date. These get even more squashed on mobile screens than buttons. But they would still need to be styled so they fill up the full width of the form
- Once again, everything here is to try things out and can be reversed

`UploadDocumentModal.jsx` mobile:             | `UploadDocumentModal.jsx` desktop:
:-------------------------:|:-------------------------:
 |

leekahung:
Think we should just incorporate the `clear form` functionality into the `cancel` button. That way, we just need to design around three buttons. I think one of the earlier iteration with `choose file` on top (without the `clear form`) and with `cancel` and `upload file` on the bottom looks pretty good.

On a different topic, I'm a bit curious about the location of the icons we're using for the buttons.

Since there's a lot of buttons that uses icons at the beginning (like the buttons in the menu, the add client button on the clients page, the new message button, etc.) and there are buttons that uses icons at the end (like the buttons in the logout modal, the add client button on the add client modal, etc.). Should we try to keep everything consistent by having them all placed on one side or do you guys think it's fine the way they are at the moment.

xscottxbrownx:
I personally prefer this version, **but with your toggle switch for the verify (instead of the checkbox) and the date picker you have**:
Screenshot 2023-07-10 at 10 09 18 AM

I also don't think "Select" needs to be in "Select Document Type" or "Enter" for "Enter Description" - but all that is nitpicky.

---

### back to the icon discussion... 
I prefer the final action buttons (Cancel, Continue) to have icons on right and any icons used in the forms/modals above those action buttons to have icons on left.
It's also not a big deal to me either way - as long as we decide on something and run with it consistently.

andycwilliams:
Alright, since we're prepping for another merge to the Master branch, I'm hoping to try to move this along now.

- Consolidated the Clear button with Cancel.
- Made the form a bit more responsive.
- Did a bit of general tweaking with the layout via several files, such as `StatusNotification.jsx`. Should now match `AddClientModal.jsx` more.
- I prefer icons on the left but can easily change them if that's what we want.
- Also, I added an icon to the webid field in `AddClientModal.jsx`. It's just me testing out adding a button for the user to copy the webid (using MUI's `InputAdornment` component) if they need to. There is no functionality yet, it's there as a concept. Can easily be removed.

Definitely willing to revisit some of these questions but in the interest of time we may want to focus more on clearing PRs for the upcoming merge.

src/components/Modals/AddClientModal.jsx
```diff
@@ -185,9 +186,21 @@ const AddClientModal = ({ showModal, setShowModal }) => {
               setWebId(e.target.value);
             }}
             fullWidth
+            InputProps={{
+              endAdornment: (
+                <InputAdornment position="end">
+                  <IconButton
+                    aria-label="copy-webid"
+                    // onClick={}
```
leekahung: Think we can remove these commented out lines in 194 and 195

src/components/Modals/UploadDocumentModal.jsx
```diff
@@ -1,34 +1,38 @@
...
 import { LocalizationProvider } from '@mui/x-date-pickers/LocalizationProvider';
-import FormHelperText from '@mui/material/FormHelperText';
 // Utility Imports
 import { runNotification } from '../../utils';
 // Custom Hook Imports
 import { useStatusNotification } from '../../hooks';
 // Component Imports
-import DocumentSelection from './DocumentSelection';
-import FormSection from './FormSection';
+import DocumentSelection from '../Form/DocumentSelection';
```
leekahung: Think you could Import DocumentSelection and FormSection from "../Form". We just need to have them included inside `/Form/index.js` for exports.

src/components/Modals/UploadDocumentModal.jsx
```diff
@@ -136,18 +145,16 @@ const UploadDocumentForm = () => {
...
               placeholder="Add a description here"
             />
-          </FormControl>
-          <br />
-          <br />
-          <FormControl fullWidth>
+            <br />
             <Button
-              variant="contained"
+              variant="outlined"
```
leekahung: Very minor thing, but I we can keep this `contained` when `file` is missing, then if `file` is chosen, we can switch this back to `outline`.

It'll help draws the eye to the `Choose file` button at the beginning then to `Upload file` after a file is chosen.

xscottxbrownx: I tested this and REALLY like it...  basically shows the user that all green buttons are good and should be done. And almost creates a step by step guide.  1st step is choose file, 2nd step is upload.

https://github.com/codeforpdx/PASS/assets/71395931/5b08988f-adbc-4e34-8a45-583cf29a307e

I did this using a ternary on the "variant" of the `choose file` button, checking for the `file` state.

andycwilliams:
Alright, I made the recommended changes. Looks pretty good to me. I do like the contained/outlined variant. 

We could even add that to `AddClientModal` for when a client's information is all filled out. 

The icons in that modal are also on the right side, instead of the left like this one. My vote is to change these ones to the left as well.

leekahung:
I think this is mostly good. We just need to have the formatting corrected.

andycwilliams:
> I think this is mostly good. We just need to have the formatting corrected.

Could you please elaborate?

xscottxbrownx:
> > I think this is mostly good. We just need to have the formatting corrected.
> 
> Could you please elaborate?

My guess is Ka Hung meant that this code is failing the prettier test at the moment.

leekahung:
> My guess is Ka Hung meant that this code is failing the prettier test at the moment.

Yeah, we just need to have the lint check passing. Shouldn't be too much.

andycwilliams:
I'm still getting those errors when running the prettier scripts. Could one of you possibly do it and push that? Thanks.

leekahung (approved):
Alright, things look fine now.

Still a bit confused over why the Prettier script is not working for your local machine, but think we got this covered for you. Approving PR for merge.

Issue #227, opened 2023-06-07
Update notification for message sent status


leekahung:
**Is your feature request related to a problem? Please describe.**
Could be useful to include indicators that a message has been sent successfully or has failed from a sent message. We have something like that now, but UI can potentially be improved.

**Describe the solution you'd like**
Update to message sent status with MUI

andycwilliams:
I've been looking at some MUI features and Snackbar seems like it could be a good use for this:
https://mui.com/material-ui/react-snackbar/

I've experimented with it a bit already. It's a highly customizable pop-up component and looks pretty good in my opinion.

Edit: It might also be an option to replace the existing status notifications in the forms.

leekahung:
Can certainly look into it. The interface looks really nice having it pop up.

leekahung:
Closing issue in favor of issue #216 and related to moving status notifications to MUI snackbar or equivalent.

Issue #219, opened 2023-06-06
Get rid of styled components styling & switch to MUI in ClientListTable.jsx and ClientListTableRow.jsx


xscottxbrownx:
### Issue:

**currently have some leftover styling in `ClientListTable.jsx` and `ClientListTableRow.jsx` from first implementation**

---

### Direction:

change to MUI components and styling instead

<img width="517" alt="Screenshot 2023-06-06 at 12 59 06 PM" src="https://github.com/codeforpdx/PASS/assets/71395931/354fd684-56a0-4994-906a-e081efdd62fa">

<img width="527" alt="Screenshot 2023-06-06 at 12 59 27 PM" src="https://github.com/codeforpdx/PASS/assets/71395931/f9aaf5f3-3b13-4f27-95bc-2a897ba7a990">

Sky-De:
Hi 
Where are those file?
Would you please mention dir to these two .jsx file

xscottxbrownx:
> Hi Where are those file? Would you please mention dir to these two .jsx file

On `Development` branch...  src/components/Clients

masenmatthews:
Hey @xscottxbrownx @Sky-De , is this being worked on? If not, could I have it assigned to me? Doesn't look like I have permissions to self-assign.

xscottxbrownx:
^ Thanks andy, I missed this in email.

andycwilliams:
Hi @masenmatthews. Just checking in. Have you made progress with this?

masenmatthews:
Hey @andycwilliams, apologies for the delayed response! Yeah I've made some progress but have gotten a bit sidelined due to work and travel demands in the last couple of weeks. I'm out of town this week, but should be able to continue working on this next week. 

If this has increased in urgency, then it might be worth reassigning if there are resources available. I should have a lot more bandwidth to work on tasks and attend meetings consistently after roughly 7/25.

leekahung:
Hey @masenmatthews was a bit curious as to the progress of this issue since it's related to PR #376. Let us know if you're still up to working on this issue. Thanks!

xscottxbrownx:
Unassigned, been 2.5months with no seen progress and lack of responses

masenmatthews:
@leekahung @xscottxbrownx 
Apologies for missing your previous message. Confirming that this can be reassigned.

andycwilliams:
This is a good issue for a newer team member to take on but I can pick it up if nobody else wants to.

[...]

PR #287, opened 2023-07-02
Issue 223/delete client modal scott

### LINKED ISSUE #223: Make the alert popup on delete of client a MUI modal component (ClientListTableRow.jsx)
### Issue:

When clicking the `delete` icon in the table of clients, an alert message pops up to confirm.  Need to use an MUI modal and styling to match rest of app instead.

---

### Direction:

Take inspiration from our other MUI modals to have similar styling for this. 
`LogoutModal`, `AddClientModal` (which isn't 100% complete yet), and even `InactivityMessage` are all closely styled and fit within the app theme.

Screenshot 2023-06-06 at 2 58 12 PM

xscottxbrownx:
# This PR closes #223 

This PR:
1. Creates `DeleteClientModal` component.
2. Renders this component onClick of trash icon in `ClientListTableRow`.
3. Renames states of `AddClientModal` component to achieve more clarity.
4. Automatically closes the `AddClientModal` component once client has been added to the list.

---

https://github.com/codeforpdx/PASS/assets/71395931/071d91d4-f4a9-4745-82b2-294d106289d2

### DIFF
--- a/src/components/Clients/ClientListTableRow.jsx
+++ b/src/components/Clients/ClientListTableRow.jsx
@@ -13,6 +13,8 @@ import { runNotification } from '../../utils';
 // Context Imports
 import { SelectedUserContext, UserListContext } from '../../contexts';
 import { StyledTableCell, StyledTableRow } from '../Table/TableStyles';
+// Component Imports
+import DeleteClientModal from './DeleteClientModal';
 
 /**
  * ClientListTableRow Component - Component that generates the individual table
@@ -28,6 +30,7 @@ const ClientListTableRow = ({ labelId, client, state, dispatch }) => {
   const { selectedUser, setSelectedUser } = useContext(SelectedUserContext);
   const { removeUser } = useContext(UserListContext);
   const [pinned, setPinned] = useState(false);
+  const [showDeleteClientModal, setShowClientDeleteModal] = useState(false);
 
   // determine what icon gets rendered in the pinned column
   const pinnedIcon = pinned ? <PushPinIcon color="secondary" /> : <PushPinOutlinedIcon />;
@@ -46,13 +49,7 @@ const ClientListTableRow = ({ labelId, client, state, dispatch }) => {
 
   // Event handler for deleting client from client list
   const handleDeleteClient = async () => {
-    if (
-      !window.confirm(
-        `You're about to delete ${client.person} from your client list, do you wish to continue?`
-      )
-    ) {
-      return;
-    }
+    setShowClientDeleteModal(false);
     runNotification(`Deleting "${client.person}" from client list...`, 3, state, dispatch);
     await removeUser(client);
     runNotification(`"${client.person}" deleted from client list...`, 3, state, dispatch);
@@ -65,45 +62,55 @@ const ClientListTableRow = ({ labelId, client, state, dispatch }) => {
   };
 
   return (
-    <StyledTableRow>
-      <StyledTableCell align="center">
-        <Checkbox
-          id={labelId}
-          color="primary"
-          checked={selectedUser.webId === client.webId}
-          inputProps={{
-            'aria-labelledby': labelId
-          }}
-          onClick={() => {
-            handleSelectClient(client);
-          }}
-        />
-      </StyledTableCell>
-      <StyledTableCell align="center">{client.person}</StyledTableCell>
-      {/* ***** TODO: Switch this webId to being a small Notes section */}
-      {/* seems having a link or even displaying another user's pod is completely useless/irrelevant */}
-      {/* see no reason it would ever need to be used, but notes/comments will be of utmost importance to caseworkers */}
-      <StyledTableCell align="center">
-        <Link
-          to={client.webId}
-          target="_blank"
-          rel="noreferrer"
-          style={{ textDecoration: 'none', color: theme.palette.primary.dark }}
-        >
-          Link to Pod Profile
-        </Link>
-      </StyledTableCell>
-      <StyledTableCell align="center">
-        <IconButton size="large" edge="end" onClick={handlePinClick}>
-          {pinnedIcon}
-        </IconButton>
-      </StyledTableCell>
-      <StyledTableCell align="center">
-        <IconButton size="large" edge="end" onClick={() => handleDeleteClient()}>
-          <DeleteOutlineOutlinedIcon />
-        </IconButton>
-      </StyledTableCell>
-    </StyledTableRow>
+    <>
+      <StyledTableRow>
+        <StyledTableCell align="center">
+          <Checkbox
+            id={labelId}
+            color="primary"
+            checked={selectedUser.webId === client.webId}
+            inputProps={{
+              'aria-labelledby': labelId
+            }}
+            onClick={() => {
+              handleSelectClient(client);
+            }}
+          />
+        </StyledTableCell>
+        <StyledTableCell align="center">{client.person}</StyledTableCell>
+        {/* ***** TODO: Switch this webId to being a small Notes section */}
+        {/* seems having a link or even displaying another user's pod is completely useless/irrelevant */}
+        {/* see no reason it would ever need to be used, but notes/comments will be of utmost importance to caseworkers */}
+        <StyledTableCell align="center">
+          <Link
+            to={client.webId}
+            target="_blank"
+            rel="noreferrer"
+            style={{ textDecoration: 'none', color: theme.palette.primary.dark }}
+          >
+            Link to Pod Profile
+          </Link>
+        </StyledTableCell>
+        <StyledTableCell align="center">
+          <IconButton size="large" edge="end" onClick={handlePinClick}>
+            {pinnedIcon}
+          </IconButton>
+        </StyledTableCell>
+        <StyledTableCell align="center">
+          <IconButton size="large" edge="end" onClick={() => setShowClientDeleteModal(true)}>
+            <DeleteOutlineOutlinedIcon />
+          </IconButton>
+        </StyledTableCell>
+      </StyledTableRow>
+
+      {/* modal/popup renders when showDeleteClientModal state is true */}
+      <DeleteClientModal
+        showDeleteClientModal={showDeleteClientModal}
+        setShowClientDeleteModal={setShowClientDeleteModal}
+        client={client}
+        handleDeleteClient={handleDeleteClient}
+      />
+    </>
   );
 };
 
--- a/src/components/Clients/ClientListTableRow.jsx
+++ b/src/components/Clients/ClientListTableRow.jsx
@@ -11,10 +11,8 @@ import PushPinOutlinedIcon from '@mui/icons-material/PushPinOutlined';
 // Utility Imports
 import { runNotification } from '../../utils';
 // Context Imports
-import { SelectedUserContext, UserListContext } from '../../contexts';
+import { SelectedUserContext } from '../../contexts';
 import { StyledTableCell, StyledTableRow } from '../Table/TableStyles';
-// Component Imports
-import DeleteClientModal from './DeleteClientModal';
 
 /**
  * ClientListTableRow Component - Component that generates the individual table
@@ -25,12 +23,17 @@ import DeleteClientModal from './DeleteClientModal';
  */
 
 // determine what gets rendered in the table body
-const ClientListTableRow = ({ labelId, client, state, dispatch }) => {
+const ClientListTableRow = ({
+  labelId,
+  client,
+  state,
+  dispatch,
+  setShowDeleteClientModal,
+  setSelectedClientToDelete
+}) => {
   const theme = useTheme();
   const { selectedUser, setSelectedUser } = useContext(SelectedUserContext);
-  const { removeUser } = useContext(UserListContext);
   const [pinned, setPinned] = useState(false);
-  const [showDeleteClientModal, setShowClientDeleteModal] = useState(false);
 
   // determine what icon gets rendered in the pinned column
   const pinnedIcon = pinned ? <PushPinIcon color="secondary" /> : <PushPinOutlinedIcon />;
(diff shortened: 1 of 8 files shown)


src/components/Clients/ClientListTableRow.jsx
```diff
@@ -65,45 +62,55 @@ const ClientListTableRow = ({ labelId, client, state, dispatch }) => {
...
-          {pinnedIcon}
-        </IconButton>
-      </StyledTableCell>
-      <StyledTableCell align="center">
-        <IconButton size="large" edge="end" onClick={() => handleDeleteClient()}>
-          <DeleteOutlineOutlinedIcon />
-        </IconButton>
-      </StyledTableCell>
-    </StyledTableRow>
+    <>
```
andycwilliams: Not critical at all, but empty fragments strike me as unusual. However, the component needs restyling anyway so this wouldn't even need to be addressed now.

src/components/Clients/DeleteClientModal.jsx
```diff
@@ -0,0 +1,64 @@
...
+    open={showDeleteClientModal}
+    aria-labelledby="dialog-title"
+    aria-describedby="dialog-description"
+    onClose={() => setShowClientDeleteModal(false)}
+  >
+    <DialogTitle id="dialog-tile">Delete Client?</DialogTitle>
+
+    <DialogContent>
+      <DialogContentText id="dialog-description">
+        {`You're about to delete ${client.person} from your client list, do you wish to continue?`}
```
andycwilliams: Minor point, but I notice in the demonstration video that there's a bit of extra spacing on the right side of the modal. That is, "to continue" is wrapped around but the "DELETE CLIENT" button stretches the modal out further to the right. It might be better if the edge of the buttons is made to align with the edge of the rightmost text, wherever that is.

andycwilliams (approved):
Approving these changes since my comments are not essential. The code is clean and addresses the related issue.

src/components/Clients/DeleteClientModal.jsx
xscottxbrownx: Interesting, I didn't notice that previously...

Looks like this modal, the Logout Modal, and the Inactivity Message are all this way. Like the dialog buttons are always over on the right, even if the sx styling is removed.  Have to look at theme or how to override for this.

src/components/Clients/ClientListTableRow.jsx
xscottxbrownx: React components can only return one thing, so it was necessary because I was returning the `<StyledTableRow>` and the `<DeleteClientModal>`.  The fragments don't produce an actual dom element so it's not in the user facing code, just here in the dev side of things - which I feel is better than using a `<div>` to accomplish which would be in the DOM at the end.

But with all that said, could put the modal inside the table row so no need for react fragments, but even though it would work fine I felt like this is a separate component from the table row so I separated it. 

I'm up for switching it, if you all feel like it should be switched no problem.

src/components/Clients/ClientListTableRow.jsx
```diff
@@ -46,13 +49,7 @@ const ClientListTableRow = ({ labelId, client, state, dispatch }) => {
...
   // Event handler for deleting client from client list
   const handleDeleteClient = async () => {
-    if (
-      !window.confirm(
-        `You're about to delete ${client.person} from your client list, do you wish to continue?`
-      )
-    ) {
-      return;
-    }
+    setShowClientDeleteModal(false);
```
leekahung: Not really an issue per-se since I believe this is appropriate too. But the modal seems to close even without this line when I click on "Delete Client".

https://github.com/codeforpdx/PASS/assets/14917816/f6839acb-401f-4fc0-8526-9d2e0f84ab10

I wonder if it has something to do with the client being removed from the list so the modal disappears on its own. If it is because of that reason, I think keeping this line should be fine since an error for whatever reason would cause the modal stuck being opened.

src/components/Clients/ClientListTableRow.jsx
leekahung: We do have two separate components in here after all (one for the row, one for the modal). I think wrapping this in a fragment is appropriate.

src/components/Clients/ClientListTableRow.jsx
```diff
@@ -65,45 +62,55 @@ const ClientListTableRow = ({ labelId, client, state, dispatch }) => {
...
-        </IconButton>
-      </StyledTableCell>
-      <StyledTableCell align="center">
-        <IconButton size="large" edge="end" onClick={() => handleDeleteClient()}>
-          <DeleteOutlineOutlinedIcon />
-        </IconButton>
-      </StyledTableCell>
-    </StyledTableRow>
+    <>
+      <StyledTableRow>
```
leekahung: Oh, one thing I want to touch, do we want to start moving these styled-components over to Material UI now, or have it in a separate PR?

leekahung:
My bad, I messed up updating the wrong branch. I've reverted the changes back to your latest commit

src/components/Clients/ClientListTableRow.jsx
andycwilliams: My vote would be to do it now, but am also okay with leaving for another PR.

src/components/Clients/ClientListTableRow.jsx
```diff
@@ -65,45 +62,55 @@ const ClientListTableRow = ({ labelId, client, state, dispatch }) => {
...
+        </StyledTableCell>
+        <StyledTableCell align="center">
+          <IconButton size="large" edge="end" onClick={() => setShowClientDeleteModal(true)}>
+            <DeleteOutlineOutlinedIcon />
+          </IconButton>
+        </StyledTableCell>
+      </StyledTableRow>
+
+      {/* modal/popup renders when showDeleteClientModal state is true */}
+      <DeleteClientModal
```
timbot1789: It makes more sense to me to put this on the Table or Page level. That way we won't be rendering 100 different modals for 100 different rows. It would also resolve the `React.Fragment` issue that Andy mentioned higher up.

You can then have the `handleDeleteClient` function call some passed-in messaging function that triggers the modal, like 

`const handleDeleteClient = () => props.callDelete(client)`

If you don't want to use props, you could set up a reducer and have the delete fire off an action, but that will be hard to integrate with the context. I haven't really thought about how to integrate contexts with reducers, so I don't have much advice there.

timbot1789 (commented):
Could you add a snapshot test for the modal? There are some example snapshots in the `tests` folder, and I can go over writing them with you if you like.

src/components/Clients/ClientListTableRow.jsx
```diff
@@ -46,13 +49,7 @@ const ClientListTableRow = ({ labelId, client, state, dispatch }) => {
...
   const handleDeleteClient = async () => {
-    if (
-      !window.confirm(
-        `You're about to delete ${client.person} from your client list, do you wish to continue?`
-      )
-    ) {
-      return;
-    }
+    setShowClientDeleteModal(false);
     runNotification(`Deleting "${client.person}" from client list...`, 3, state, dispatch);
```
timbot1789: Non-Issue: Any thoughts on moving these notifications into the modal? Seems like if we're going to be displaying messages about deletion, we can display them all in one place. Doesn't have to be in this PR if we want to change it.

src/components/Clients/ClientListTableRow.jsx
xscottxbrownx: That's Issue #219 assigned to Masen. Pinged them in our Discord DMs.

src/components/Clients/ClientListTableRow.jsx
xscottxbrownx: It closes when the action is complete.  I was having it close early, so user could see the status notifications that are displayed on the table.

Could move the notifications to the modal, as Tim said below.  

Or have notifications some other way:
 - Just a spinner while it's running (https://mui.com/material-ui/react-progress/#circular) and then a snackbar for error or success (https://mui.com/material-ui/react-snackbar/#customization)
 - Interactive integration (https://mui.com/material-ui/react-progress/#interactive-integration)

xscottxbrownx: @timbot1789 since you mentioned it in another comment, take a look at these options as well and see what you think.

src/components/Clients/ClientListTableRow.jsx
xscottxbrownx: See comment above from Ka Hung and my quickly researched response.

src/components/Clients/ClientListTableRow.jsx
timbot1789: I like the snackbar because we'll probably want to use something similar for other messages throughout the system.

andycwilliams: This is related to issue #227 and it looks like Russell just assigned himself to it

xscottxbrownx:
OK,

summary:

1. Renamed states for the `AddClientModal` to that instead of `showModal` - due to having another modal in `Clients` now (`DeleteClientModal`) just to keep code more clear.
2. Lifted the `DeleteClientModal` component up to `ClientList` (now contains the `ClientListTable` and this `DeleteClientModal`.)
3. Moved the status notifications to the bottom of the `DeleteClientModal` instead of the `ClientListTable`. (This is still planned to be updated to snackbars from MUI in a separate PR/Issue.)
4. Added all the `Clients` components to its `index` file.
5. Added test file for this component.

xscottxbrownx:
Thanks to @timbot1789 for giving a crash course in testing!

### I'm still having issues trying to pass my last test that I'd like to implement for this PR. So, I will describe what it is and the approaches I have tried so far - maybe somebody can see the issue, or think of another approach to try.

---

### I want to test that the modal does close without taking any action, when the `Cancel` button is pressed.

---

#### 1) Thought of checking to see if specific elements (that are only in the modal) exist after button click (this one was specifically the delete button.  I tried it two different ways:
Screenshot 2023-07-10 at 9 11 59 AM
Screenshot 2023-07-10 at 9 10 51 AM

--- 

#### 2) Thought of checking to see if the setter function for the state (that controls whether the modal is displayed or not) was fired on button click:
Screenshot 2023-07-10 at 9 16 04 AM

I believe approach 2 to be decent, but it is most likely not written correctly. I don't know how it can correlate my made up `closeModal` function to the modal actually calling `setShowDeleteClientModal` state setter function.

---

If anybody sees any issues with the code, has thoughts on a different approach, etc - please ping me with response.

timbot1789:
@xscottxbrownx Another thing you could do is write a simple wrapper component to launch the modal in, and test that:

```
import { render, cleanup, waitForElementToBeRemoved } from '@testing-library/react';
import userEvent from '@testing-library/user-event';
import { expect, it, vi, afterEach } from 'vitest';
import { DeleteClientModal } from '../../../src/components/Clients';
import { UserListContext } from '../../../src/contexts';

...


it('closes without an action when cancel button in clicked', async () => {
  const user = userEvent.setup();
  const client = { person: 'tim' }
  const removeUser = vi.fn();
  const ModalWrapper = () => {
    const [showModal, setShowModal] = useState(true)

    return <UserListContext.Provider value={{ removeUser }}>
      <DeleteClientModal
        showDeleteClientModal={showModal}
        setShowDeleteClientModal={setShowModal}
        selectedClientToDelete={client}
      />
    </UserListContext.Provider>
  }

  const { queryByLabelText } = render(<ModalWrapper/>);

  const cancelButton = queryByLabelText("Cancel Delete Client Modal");
  await user.click(cancelButton);
  // wait for the dialog to drop
  await waitForElementToBeRemoved(() => queryByLabelText("Cancel Delete Client Modal"))
  const hiddenCancelButton = queryByLabelText("Cancel Delete Client Modal");
  expect(hiddenCancelButton).toBeNull();
  expect(removeUser).not.toBeCalled();
});
```
This test seems to work without issue. One thing I had to look up was the `waitForElementToBeRemoved` function, which does about what it says on the box.

---
However, the React Testing Library docs do include a set of examples, one of which is testing modals:
https://testing-library.com/docs/example-react-modal/

You can see that they do the same test as you wrote. So I'm fine with either approach, but you're right, the test would be more robust if it checks that the modal is actually dropped. Not that it just calls the `drop` function.

leekahung (approved):
The changes seems to work fine and the code is running well. I'll approve this.

src/components/Clients/DeleteClientModal.jsx
```diff
@@ -0,0 +1,109 @@
...
+  // Event handler for deleting client from client list
+  const handleDeleteClient = async (event) => {
+    event.preventDefault();
+    runNotification(
+      `Deleting "${selectedClientToDelete?.person}" from client list...`,
+      5,
+      state,
+      dispatch
+    );
+    try {
```
timbot1789, line 45: This block seems odd to me. It says "attempt to delete the client, then, regardless of if the attempt succeeds or fails, tell the client that it succeeded." I think what you mean is something more like this:

```
try {
  await removeUser();
  runNotification("User deleted!", 5, state, dispatch)
} catch (e) {
  runNotification(`User Deletion failed. Reason: ${e.message}`)
} finally {
  setTimeout(() => {setShowDeleteClientModal(false)}, 2000)
}
```

That says "try to delete the user. If you succeed, display a success message. If you fail, display an error message. No matter what, drop the modal after 2 seconds"

src/components/Clients/DeleteClientModal.jsx
```diff
@@ -0,0 +1,109 @@
...
+  };
+
+  return (
+    <Dialog
+      open={showDeleteClientModal}
+      aria-labelledby="dialog-title"
+      aria-describedby="dialog-description"
+      onClose={() => setShowDeleteClientModal(false)}
+    >
+      <FormSection
```
timbot1789, line 69: Followup: This shouldn't really be a form. I get it's because you want to use the notifications system, but that seems to me like a sign that we need a better notifications system. We should be able to send notifications from any kind of component, not just forms. After we rework notifications to be robust, we should consider reworking this form too. That's for the future though.

src/components/Clients/DeleteClientModal.jsx
xscottxbrownx: oh yes, agree it needs a `catch` block.

But leaving this before the `try` block:
`runNotification(
      'Deleting "${selectedClientToDelete?.person}" from client list...',
      5,
      state,
      dispatch
    );`
    so the user has some sort of indication that something is going on.

src/components/Clients/DeleteClientModal.jsx
xscottxbrownx: Fully agreed. I thought we already had an issue for that, but I guess not. I will create one.

Revamping the UI/UX of `<StatusNotification>` is definitely out of the scope of this PR.  I just made the code here the same as it was when the notifications were showing in the client table (and it's how they are coded in the `AddClientModal`, and all the other forms we still have - so it's easier for whichever dev wants to tackle that issue.

timbot1789: Yup I totally agree that what you did here was the right thing given the current state of the code base. Mostly just wanted to call it out, since we don't want this design pattern to proliferate too much. Nice work either way

src/components/Clients/DeleteClientModal.jsx
timbot1789: Works for me!

timbot1789 (approved):

andycwilliams:
I know this just closed, but I'm running from the latest Dev branch and am seeing this:



I haven't had time to check what's causing this yet.

xscottxbrownx:
> I know this just closed, but I'm running from the latest Dev branch and am seeing this:
> 
> 
> 
> I haven't had time to check what's causing this yet.

Sorry, causing what?  I'm not seeing what you are pointing out.

andycwilliams:
I meant the uncentered buttons. Actually I just realized we have a couple of different philosophies with button placement. On some modals they're centered and on others they're not. Not something we need to address now, if at all.

timbot1789:
> I meant the uncentered buttons. Actually I just realized we have a couple of different philosophies with button placement. On some modals they're centered and on others they're not. Not something we need to address now, if at all.

Sounds like a great first issue for someone new to the project.

xscottxbrownx:
Yep, see our convo above... it's basically the modals that are just using `<Dialog>` and `<DialogActions>` compared to those that have a `<form>` or `<FormSection>` inside of them.

PR #288, opened 2023-07-03
Update snapshot

### LINKED ISSUE #280: Snapshot bug for OidcLoginComponent.test.jsx
**Describe the bug**
While the test passes when checked on GitHub, the snapshot seems to fail when it checks for it locally using `npm run test`. It could be related to the environment variables being used.

**Expected behavior**
To pass in both GitHub and locally.

timbot1789:
Simple snapshot update to show that the OidcLoginComponent is reading the mocked ENV properly.

### DIFF
--- a/test/components/NavBar/OidcLoginComponent.test.jsx
+++ b/test/components/NavBar/OidcLoginComponent.test.jsx
@@ -22,6 +22,14 @@ afterEach(() => {
   cleanup();
 });
 
+vi.mock('../../../src/constants/', () => {
+  const actual = vi.importActual('../../../src/constants/');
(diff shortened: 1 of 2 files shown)


leekahung:
Think we can have this merged into Development. It's probably the fix for the failed tests in PR #274

timbot1789:
@leekahung  I changed the implementation here. Can you take another quick look and hit approve if you're ok with it? Shouldn't be anything significant, but still good to get eyes on it.

leekahung (approved):
Ah okay. I see where the changes are. Should be fine. Thanks for the updates!

PR #295, opened 2023-07-08
Fixed bug for profile image when profileImage is null

leekahung:
This PR fixes a bug related to storing initial profileImage to localStorage. It should really occur if profileImage is not null, but exist within profile card.


Jared-Krajewski (approved):
reviewed change in discord for fix.

PR #297, opened 2023-07-09
Issue 296/initialize-document-acl

### LINKED ISSUE #296: Re-establishing initial ACL file for Documents
We had for a time removed the ability the cross-pod features while we set up the new documents scheme for uploaded files. However, in the process, swapped to using the default ACL settings for `/PASS/Documents` as a result.

I think we could simply initialize a ACL file like what is being done with other PASS-specific containers.

Create new ACL file if Documents container exist, but does not have one. Or create a new Documents container with a new ACL file if Documents does not exist. ACL setting will be set to the user initially, with setting to Agents being allowed to change from what we have previously with the set permission functions.

leekahung:
## PR Details

This PR will re-establish cross-pod functionality for file viewing and permission setting to `/PASS/Documents` by initializing ACL file for `/PASS/Documents` if ACL does not exist in `/PASS/Documents`.

See clip:

https://github.com/codeforpdx/PASS/assets/14917816/68105ebb-ca54-4a39-9309-258dd3ea19b2

`setDocContainerAclPermission` has been updated to only update permissions to `/PASS/Documents` with the Access Object for users who's permissions were given to read and append only, compare to read and write previously. This would allow case workers to able to upload new documents and read existing documents of clients, but not delete existing documents on the clients Pod.

`setDocContainerAclPermission` are also updated to utilize `podUrl` to simplify the generation of the containerUrl to `/PASS/Documents`.

`createPublicContainer` is also re-introduced to PASS, however, this would be a limited version meant to create a skeleton container for generating the `/PASS/Documents` ACL for the user. If the container already exist, but missing an ACL, it'll generate a new ACL. If the container is missing, it'll generate a new `/PASS/Documents` container with the new ACL. No additional containers would be built.

Additional changes has been made for `SignedInUserContext`, with `podUrl` and `profileData` now being stored within their own separate useStates, but keeping the same name for storing their relative states.

### DIFF
--- a/src/utils/network/session-core.js
+++ b/src/utils/network/session-core.js
@@ -3,7 +3,8 @@ import {
   getSolidDataset,
   getThingAll,
   getFile,
-  hasAcl
+  getSolidDatasetWithAcl,
+  hasResourceAcl
 } from '@inrupt/solid-client';
 import { INTERACTION_TYPES } from '../../constants';
 import {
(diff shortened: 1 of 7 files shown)


xscottxbrownx (commented):
The write-up here, and the component file make sense/look good to me.  But, I'd like a second set of eyes on the context, util, and test files that were changed.  I'm not as confident with them.  They seem to look good/make sense to me though.

src/utils/network/session-core.js
```diff
@@ -112,6 +117,39 @@ export const createPublicContainer = async (session, podUrl) => {
...
+ * access to the user
+ *
+ * @memberof utils
+ * @function createDocumentsContainer
+ * @param {Session} session - Solid's Session Object {@link Session}
+ * @param {URL} podUrl - The user's Pod URL
+ * @returns {Promise} Promise - Generates a Documents container for Pod upon log
+ * in if user's Pod does not have the an outbox to begin with
+ */
+export const createDocumentsContainer = async (session, podUrl) => {
```
timbot1789: This is fine for now, but I think it's a good idea to break up `session-core` into more specific files with more targeted responsibilities. The name for this is already a bit confusing. `session` and `networking` both imply that these functions are responsible for establishing a connection with the server. But `createDocumentContainer` is clearly devoted to container and file management, which isn't the same as session management.

timbot1789 (approved):

src/utils/network/session-core.js
leekahung: I think I could break this up from session-core, perhaps placing them in a folder named pod-management or something like that.

I could move the other create container functions into there too. I'll give that a go first.

timbot1789 (approved):

PR #299, opened 2023-07-09
Issue 298/update profile to spec

### LINKED ISSUE #298: Update Profile close to recent specifications
It would seem there's a newer specification with regards to a Profile on Solid. (see https://solid.github.io/webid-profile/)

Think we could incorporate into the existing profile card if the server does not implement them by default.

leekahung:
## Update: [Draft]

This PR will update the profile card based on more recent specifications as pointed by Jeff: https://solid.github.io/webid-profile/

On the frontend, the "Profile" page will remain unaffected. The changes are largely under the hood, where it updates a Solid WebId Profile from PASS to fill out certain fields that are consider a must or recommended in the spec for a full Profile. This includes:

- `pim:preferencesFile`
- `solid:publicTypeIndex`
- `solid:privateTypeIndex`
- `prim:storage`
- `ldp:inbox`

and are only triggered if any of the above is missing.

A new `settings` container has been created to store the relevant preference files and type index files.

Extended Profile Documents are excluded for now.

### DIFF
--- a/src/constants/rdf_predicates.js
+++ b/src/constants/rdf_predicates.js
@@ -1,4 +1,4 @@
-import { SCHEMA_INRUPT, FOAF } from '@inrupt/vocab-common-rdf';
+import { SCHEMA_INRUPT, FOAF, LDP, RDF } from '@inrupt/vocab-common-rdf';
 
 const RDF_PREDICATES = {
   ...SCHEMA_INRUPT,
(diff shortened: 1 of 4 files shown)


timbot1789 (commented):
I think this is getting too down in the weeds of solid, and this isn't really what PASS is good for. We're not trying to build a full solid pod manager. We're trying to use solid pods to allow homeless people to more easily share their documents with social services. I don't see how the fine-tuned pod preference control systems added in this PR help with that right now. Maybe in the future there could be a use for a PASS settings file, but we're not at that stage yet.

Every application needs to limit itself to features that help it achieve its goals, and that it can realistically implement with its resources. Managing the private type indices of a user's pod does not help PASS achieve its goals. And we don't have the resources to properly implement that even if we wanted to.

leekahung:
Think we could postpone this to if we would need this in the future. Wouldn't mind closing this.

PR #302, opened 2023-07-13
Updated Inactivity Message

brancwill:
Set inactivity popup timer to 25 minutes, and added another 5 minute timer that forcibly logs out the user.
inactivityWindowUpdate


leekahung:
It looks like the branch started out from an older version of Development, which is why the merge conflict exists when the updates are pushed (see how one still uses styled-components, the updated version od Development fully utilizes MUI components)

leekahung:
Oh, also something to note (though I don't think it's not too urgent at the moment), but could we have this branch renamed into something shorter? Looks like the branch just took the issue title and used it word for word instead of paraphrasing it into something more manageable.

src/components/Notification/InactivityMessage.jsx
```diff
@@ -58,25 +61,54 @@ const InactivityMessage = () => {
...
+          setSeconds(59);
+          setMinutes((minutes) => minutes - 1);
+        } 
+          else if (seconds > 0) {
+          setSeconds((seconds) => seconds - 1);
+        }
+      }, 1000)
+    }
+    return () => clearInterval(logoutTimer.current);
+  }, [showPopup, seconds, minutes])
```
timbot1789: Thanks for the submission! Overall, this looks like a really useful improvement. I think we can improve the performance here though. Currently, this function will run every second. Can we make it more efficient? We really only need the function to run once, when the timer has expired.

We could use something like `setTimeout` to fire off the logout command after the time has expired. Then, if we exit early, use `clearTimeout` to suppress the logout. There may be a more 'react' way to do that as well.

brancwill: I could move the logout command into a setTimeout, but I'd still need setInterval to maintain the timer that's displayed to the user

leekahung: About the `setInterval`, I don't necessarily think we need the full timer to be running since this PR is meant to force logout for inactive users which might not care about the specific amount of time they have left on the application.

If we wish to keep a timer, I believe a minute count should suffice, for instance `5 minutes remaining`, `4 minutes remaining`, etc.

src/components/Notification/InactivityMessage.jsx
```diff
@@ -58,25 +61,54 @@ const InactivityMessage = () => {
...
   // Returns user to home page upon successful logout
   const handleLogout = () => {
     localStorage.clear();
   };
 
+  // Event handler for popup's continue button
+  // Closes the window, and resets the timer
+  const handleClosePopup = () => {
+    setShowPopup(false);
+    setSeconds(0);
```
leekahung: For the set functions for the time, I think we could have them placed inside the clean-up function for the useEffect with the logoutTimer to reset the time rather than inside a different handler function.

src/components/Notification/InactivityMessage.jsx
```diff
@@ -58,25 +61,54 @@ const InactivityMessage = () => {
...
+     Logs the user out and refreshes the page if time elapses. */
+  useEffect(() => {
+    if (showPopup) {
+      logoutTimer.current = setInterval(() => {
+        if (seconds === 0 && minutes === 0) {
+          handleLogout();
+          window.location.reload(false);
+        } else if (seconds === 0) {
+          setSeconds(59);
+          setMinutes((minutes) => minutes - 1);
```
leekahung: For these set functions it looks like you're trying to use the previous state to set a new state.

If that's the case, I think you could use something like `setMinutes((prevMinute) => prevMinute - 1)` or `setMinutes((prev) => prev - 1)` instead since it'll already imply what `minutes` would be. It'll also avoid clashing with the original `minutes` state.

Else, I believe something like `setMinutes(minute - 1)` should suffice without using the previous state and just manipulate the state to be set directly.

src/components/Notification/InactivityMessage.jsx
```diff
@@ -58,25 +61,54 @@ const InactivityMessage = () => {
     };
   }, []);
 
+  /* Starts a five minute timer when the inactivity window pops up. 
+     Logs the user out and refreshes the page if time elapses. */
+  useEffect(() => {
+    if (showPopup) {
+      logoutTimer.current = setInterval(() => {
+        if (seconds === 0 && minutes === 0) {
+          handleLogout();
```
leekahung: I believe the linter would pick this up, it looks like `handleLogout` is used before it's defined. Should be able to fix this by moving `handleLogout` before this `useEffect`.

src/components/Notification/InactivityMessage.jsx
timbot1789: @brancwill  That's a fair point. Looking at it again, how you're using `setInterval` makes sense. I guess my question then is why is the `useEffect` watching the `seconds` and `minutes`? It seems to me it can just fire off the `setInterval` on mount and then clean up after it.

I also don't think you need to keep seconds and minutes separate. You can just have one value like `const [secondsToLogout, setSecondsToLogout] = useState(300)` and then compute the minutes and seconds in the JSX.

```
return <div>{Math.floor(secondsToLogout/60)}:{secondsToLogout % 60}</div>
```

brancwill: If the seconds state wasn't in the dependency array, then the timer wouldn't update properly and subtract minutes like they were seconds. With the implementations you suggested however, that is no longer necessary.

src/components/Notification/InactivityMessage.jsx
brancwill: Ok, I moved them there and removed the "handleClosePopup" function.

src/components/Notification/InactivityMessage.jsx
brancwill: I tried the last thing you suggested, but the timer would stop counting down after one second. I ended up going with the ((prev) => prev - 1) you suggested.

src/components/Notification/InactivityMessage.jsx
brancwill: I moved the handleLogout function up as you suggested. You mentioned before about the branch starting from an older version of Development. Does that mean I should start a new branch to put these changes into?

src/components/Notification/InactivityMessage.jsx
leekahung: All good. Think it makes sense to use a previous state to update the next state for a timer.

src/components/Notification/InactivityMessage.jsx
leekahung: You could do that, but you don't have to. You simply need to update this branch by merging the changes made in Development into this one. But because they're from diverging branches, it would have to deal with merge conflicts.

leekahung:
The force logout seems to be working as intended, I've just noticed a UI thing with regards to the timer, see clip below.

https://github.com/codeforpdx/PASS/assets/14917816/74970a5f-77ed-44ea-bfb4-d974c3ec7d9b

To testing things out, I've reduced the time to trigger the inactivity modal at 1 second and trigger logout 5 seconds after that.

The `handleLogout` function still triggers when the time is 0, but the counter still counts down until `handleLogout` is completed. Was thinking we could have the timer stay at 0 when it reaches 0 or have `handleLogout` trigger a little early. I'm fine with either solutions.

src/components/Notification/InactivityMessage.jsx
Jared-Krajewski: If you were to merge development into this branch you would have to re base off of Development. Re basing should resolve the merge conflicts. I Updated instructions in contributing.md because previous instructions did not specify what to base from when creating a branch on GitHub issues. This looks like it was created through that and by default based off of Master.

brancwill:
I made these changes: 

1. `Rebased` the branch.
2. Removed `seconds` and `minutes` in favor of `secondsToLogout`.
3. Added a conditional statement to the time increment to prevent the timer from dropping below 0.
4. Removed the `handleClosePopup` function I created, and moved the functions within it elsewhere.

I was forced to put `secondsToLogout` as a `useEffect` dependency, else the function wouldn't check the condition of `secondsToLogout > 0`, and cause the problem with the timer going into negatives like @leekahung pointed out above. This came into question with `seconds` and `minutes`, so I don't know if it's a problem here.

I was also unable to put the `clearTimeout` inside of the `useEffect` cleanup function, as that caused it to remove the timeout at incorrect times. I instead placed it inside of the "Continue" button's `onclick`.

Please let me know if this is fine now, or if more refactoring is needed.

leekahung:
Hey @brancwill, thanks for the updates! Unfortunately, it looks like there could be a mishap with the branch updating. I'm seeing earlier commits from Development mixed in this PR along with the Files changed even though it suppose to only be InactivityMessage.jsx. Let us know if you're having difficulties handling/resolving the merge conflicts.

leekahung:
Other than that, think there's still one more thing (see clip below). The continue button seems to work as intended, which is nice. I also like the updated design to make the modal match closer to what other modals look on PASS. However, if a user accidentally clicks outside the modal, the modal closes and the user no longer have the option to stop the forced logout.

https://github.com/codeforpdx/PASS/assets/14917816/e54f07dd-8f1c-4f92-8aa8-84110f6d290d

I was thinking we could perhaps have the backdrop as part of the modal. That way, a user wouldn't be forced out of the modal.

brancwill:
That's weird, I'm not sure why. Should I just try to `rebase` again?

And that is a problem. I could write it like I did initially with a second condition inside the timer's `setInterval` that checks for when `secondsToLogout` reaches 0, and log the user out that way. The interval clears no matter how the modal is closed.

leekahung:
> That's weird, I'm not sure why. Should I just try to `rebase` again?

You could perhaps save the changes you're done on InactivityMessage.jsx outside the repo directory and revert the changes to the commit before you rebased and try again. After everything got resolved, you could move the changes back into the repo. That's one way to do it. 

If that couldn't seem to work for you or you're still having trouble with the rebasing, you could always directly checkout a new branch from Development and save your changes there if you wish to just get the changes for InactivityMessage.jsx made.

> And that is a problem. I could write it like I did initially with a second condition inside the timer's `setInterval` that checks for when `secondsToLogout` reaches 0, and log the user out that way. The interval clears no matter how the modal is closed.

Think it'll be better to have the user actively click on the button to stay logged in rather than clicking anywhere else for that.

brancwill:
@leekahung I checked out a new branch. In this one though, the inactivity message doesn't appear when `showPopup = true`. I switched to the Development branch to test, and the same thing happened there. Do you know if this happens for everyone, or if this is just a problem on my end?

leekahung:
> @leekahung I checked out a new branch. In this one though, the inactivity message doesn't appear when `showPopup = true`. I switched to the Development branch to test, and the same thing happened there. Do you know if this happens for everyone, or if this is just a problem on my end?

No, you're correct. There seems to be a bug for the notification in Development.

I've done a little digging on my end, but this is likely caused by other state, `activeUser`, not changing to `true`. The `loggedIn` state in localStorage seems to no longer be in use, hence the reason by `setActiveUser` never gets set to `true`.

If you want to, we could use the `session.info.isLoggedIn` value instead. You could call session via useSession() (you could check out AppRoutes.jsx to see how to extract that value).

Issue #247, opened 2023-06-11
Replace Javascript's Date function with Day.js Date Objects throughout codebase


timbot1789:
# Update July 7, 2023 [Ka Hung]

Day.js will be the ISO-8601 compliant library we'll be using moving forward. We have the library installed in the project, we just need to replace the old JS Date Objects with the Day.js Day Objects.

---

## Original Post:

**Is your feature request related to a problem? Please describe.**
Currently for dates, we are using javascript's built-in `Date` library. However, this implementation is notoriously buggy, and does not comply with the ISO 8601 date format (the most widely-used date interchange format). Specifically, it does not correctly render timezone information when the timezone is missing. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#date_time_string_format

You can read a clearer discussion of this problem at the [linked blog post](https://maggiepint.com/2017/04/11/fixing-javascript-date-web-compatibility-and-reality/)

**Describe the solution you'd like**
Remove `Date` from our code and replace it with a javascript library that correctly implements the ISO 8601 spec. `moment.js` is the most popular library of this type.

**Describe alternatives you've considered**
There is work being done on a built-in library called [Temporal](https://tc39.es/proposal-temporal/docs/index.html), however it's not ready for production yet.

**Additional context**
The issue here is mostly around cross-platform and cross-timezone compatibility. As long as we only have javascript code using `Date` acting on PASS files, we should be fine. However, if we ever begin working with non-javascript code, and have users in multiple different timezones, the buggy `Date` implementation will start causing problems.

In fact, this may already be the case with the driver's license barcode scanner. I haven't tested it, but the dates on the barcode are ISO-8601.

timbot1789:
Not something that's needed for MVP, but the sooner we deal with it, the easier it will be to fix. If PASS takes off, this issue will rear its ugly head in unexpected places.

ajay2827:
@timbot1789 I am interested in this, can you assign this to me?

xscottxbrownx:
> **Is your feature request related to a problem? Please describe.** Currently for dates, we are using javascript's built-in `Date` library. However, this implementation is notoriously buggy, and does not comply with the ISO 8601 date format (the most widely-used date interchange format). Specifically, it does not correctly render timezone information when the timezone is missing. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#date_time_string_format
> 
[...]

PR #304, opened 2023-07-14
Issue 247/update to dayjs

leekahung:
PR's related to issue #247 where we want to replace existing JS Date objects with a library that's ISO-8601 compliant. As such, I've utilized the existing DayJS library in our project to do just that.

All JS Date objects has been converted to their DayJS equivalent in the codebase and some have been simplified for formatting.

One minor detail I've noticed was that the ordering of the `Choose File` variant was reversed. Since it's only one line, I've decided to I've patched it up in this PR as well.


timbot1789 (approved):
This looks good to me!

PR #305, opened 2023-07-14
Fix spacing in mock template

timbot1789:
eslint is configured to alert us if a file does not have a new line at the end. Not sure why it isn't doing that here. We can look into that later.


leekahung (approved):
Yeah, that is pretty weird. Think we can have it like this for now.

PR #306, opened 2023-07-15
Minor patch for DocumentSelection

leekahung:
Found a small bug where DocumentSelection MenuItem was using `key` instead of `value` for the `value` attribute. So the type value would be correct with "Bank Statement" instead of "BankStatement", "Driver's License" instead of "DriversLicense", etc. These changes should fix things without breaking anything.

https://github.com/codeforpdx/PASS/assets/14917816/de1cd54c-f993-45e8-b5ef-d5f143f9de9e

### DIFF
--- a/src/components/Form/DocumentSelection.jsx
+++ b/src/components/Form/DocumentSelection.jsx
@@ -30,7 +30,7 @@ const DocumentSelection = ({ htmlForAndIdProp, handleDocType, docType }) => (
         name="document"
       >
         {Object.entries(DOC_TYPES).map(([key, value]) => (
-          <MenuItem key={key} value={key}>
+          <MenuItem key={key} value={value}>
             {value}
           </MenuItem>
         ))}
--- a/src/model-helpers/Document.js
+++ b/src/model-helpers/Document.js
@@ -117,7 +117,7 @@ const createFileChecksum = async (file) => {
  * @returns {Promise<ThingLocal>} a thing
  */
 const addAdditionalInfo = async (docDesc, thing, file) => {
-  if (docDesc.type === 'DriversLicense') {
+  if (docDesc.type === "Driver's License") {
     const retThing = await addDriversLicenseInfo(thing, file);
     return retThing;
   }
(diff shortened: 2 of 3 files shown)


timbot1789 (commented):
I'm not sure I understand this patch. What exactly is the bug you found? The goal of the `DOC_TYPES` constant is to separate the displayed form of a document type from the actual data representation. Although something like 'Driver's License' makes sense to a human, the `'` and the space make it harder for a machine to parse. This structure lets us set a data value stored in the database (a "key") from the form displayed to users (a "value").

Is there an issue with this idea?

src/components/Form/DocumentSelection.jsx
```diff
@@ -30,7 +30,7 @@ const DocumentSelection = ({ htmlForAndIdProp, handleDocType, docType }) => (
         name="document"
       >
         {Object.entries(DOC_TYPES).map(([key, value]) => (
-          <MenuItem key={key} value={key}>
+          <MenuItem key={key} value={value}>
```
timbot1789, line 33: I admit the naming here is confusing, but the code itself is correct. The "key" here is the data value, while the "value" is the user-friendly form of that value. The value was not intended to be saved into data, or used in code, which is why it looks this way.

leekahung: The data was being stored into the TTL file as "BankStatement" and "DriversLicense" for the TTL file identifier and additionalType instead of "Bank Statement" and "Driver's License", so I thought it was a bug.

src/model-helpers/Document.js
```diff
@@ -117,7 +117,7 @@ const createFileChecksum = async (file) => {
  * @returns {Promise<ThingLocal>} a thing
  */
 const addAdditionalInfo = async (docDesc, thing, file) => {
-  if (docDesc.type === 'DriversLicense') {
+  if (docDesc.type === "Driver's License") {
```
timbot1789, line 120: A better way to do this would be `DOC_TYPES.DriversLicense`

src/components/Form/DocumentSelection.jsx
timbot1789: That's actually the intent of this code, not a bug. Store the data in a way that's easy for machines to interpret, display it in a way that's easy for humans to read. Spaces like we have in "Bank Statement" or "Driver's License" make the strings harder to parse in code. The goal is to store these types in a format that makes the data easier to query and reference in the future.

leekahung: Alright, that should be fine. I'll close this PR.

leekahung:
Closing PR. Could potentially use type for querying purposes.

timbot1789:
A good enhancement to make here would be to find RDF types to save instead of these strings. We could research what would be the best types for that.

PR #309, opened 2023-07-17
Issue 308/update status notification

### LINKED ISSUE #308: Switch to use "file.name" instead of "state.file.name" for StatusNotification
**Describe the bug**
At the moment, we're using a local useState for `file` in `UploadDocumentFile`. This makes `state.file` unused. Had thought about removing it completely, but the `state.file` is used for the `StatusMessage` component, where it is used to limit the size of the filename if it's too long to prevent the modal from stretching.

**Proposed Solution**
We can repurpose `StatusNotification` to take in an additional prop for file name. Instead of `state.file.name`, we could pass in `file.name` instead from the local useState, which can be then be passed into `StatusMessages`. The rest should work the same as before.

We do need to, however, remove unused parts of `state.file` from the reducer, reducer state, and typedefs since they're not longer in use.

leekahung:
This PR updates `FormSection` and subsequent children components `StatusNotification` and `StatusMessage` to utilize `file` instead of `state.file` since it's no longer in use.

The original intent of having the file name there in the first place was to decouple the file name from the status message in case there's a file with a long name (see Issue #244 and PR #245). This PR should restore this UI feature while removing the deprecated state for file from the reducer and relevant lines.

https://github.com/codeforpdx/PASS/assets/14917816/c056a9b7-20f3-4e94-8df2-f34280820ddf

Truncation like before utilizes CSS properties to truncate rather than explicitly truncating the file name to avoid data manipulation. That remains unchanged.

### DIFF
--- a/src/components/Form/FormSection.jsx
+++ b/src/components/Form/FormSection.jsx
@@ -20,7 +20,7 @@ import { StatusNotification } from '../Notification';
  * {@link formSectionProps})
  */
 
-const FormSection = ({ title, state, statusType, defaultMessage, children }) => (
+const FormSection = ({ title, state, statusType, defaultMessage, file = null, children }) => (
   <Box
     sx={{
       marginTop: 1,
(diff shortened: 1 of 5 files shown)


andycwilliams:
Looks good. Will definitely help whenever we begin implementing pop-up notifications.

timbot1789 (approved):

Issue #307, opened 2023-07-15
Change permission forms to use pod URLs explicitly instead of usernames


leekahung:
**Is your feature request related to a problem? Please describe.**
Similar to WebIds, since it'll be difficult to assume what form a pod URL would take shape on different servers, it's best if we treat pod URLs like email addresses, where we'll need to use for setting permissions. We're currently using usernames for the permission forms, that could be changed.

**Describe the solution you'd like**
Replacing username with podUrls for both set permission forms.

leekahung:
Issue resolved in PR #349. Closing issue.

PR #321, opened 2023-07-20
Issue 320/create client profile

### LINKED ISSUE #318: Refactoring Documents to Profile
Given that the profile page would be the location where the user's information would be stored, we could move what we have currently in Documents into Profile in the order shown in the UX design on Figma: https://www.figma.com/file/BfsGjGbFUVvtqNq18sykoY/Pass-Project?type=design&node-id=1216-2986&mode=design&t=rZZ41l9198JVqLUf-0

Move components from Documents into Profile. Additional changes to UI towards UX design can be done after the migration of components.

### LINKED ISSUE #320: Create separate Profile page for case workers to click into from Clients
**Feature Description**:
While there's a profile page in PASS, that's a profile page specifically for the user. We currently do not have one for case workers to click into from the Clients page.

**Proposed Solution**:
Create a separate Profile page that reads and display information about a client after clicking into it. Should display information similar to the user's profile page, except it is not editable. Should follow the order similar to what is shown in the Figma board: https://www.figma.com/file/BfsGjGbFUVvtqNq18sykoY/Pass-Project?node-id=537-28&t=e7ix0PRe35z9R5QF-0

UI/UX update could come after this feature is finished.

leekahung:
# PR Details

This PR aims to create a new `Client Profile` page that's separate from the `User Profile` page. It'll utilize the existing `Profile` page as a template and is distinguished with the router state for `client`, (i.e. if `client` state is null, will return the client's documents list, profile page, etc., else it'll return the user's own profile, documents list, etc.)

Note that a `Client` refers to "a person case workers work with" and that a `User` refers to "the person that's currently using PASS", the "PASS User".

https://github.com/codeforpdx/PASS/assets/14917816/d612c640-84e0-4b68-91b8-a145e00e7305

The new Client Profile page will have the route `/PASS/profile/:webId` as oppose to `/PASS/profile` which is the User Profile page. In this new page, profile information cannot be edited, just shown as it displays information found from the client's profile. The webId is converted into an encoded URI that could be read by the browser.

As shown in the clip, the `Link to Profile` has been re-purposed to link to that new page and been moved to the Client's name itself. The `Select` column has been removed in favor of selecting user this link. If clicked, the client will be selected based on that link. As such the column for the link is also removed.

The `Documents` page has been phased out completely.

As for the set permission components, they've been updated to recognize the different profile pages. For the client profile page, they'll be "Request Permission" forms. For the user profile page, they'll be "Set Permission" forms. For the time being, we have yet to create a request permission function, so TODOs are created instead.

The forms for set permissions would need to be converted similar to the add Documents modal, but that would be for a separate PR.

Other changes include changes to the `fetchProfileInfo` function, which simply uses `webId` now instead of both `session` and `webId`. All relevant code that previously uses `fetchProfileInfo` has been adjusted accordingly. Changes to package.json with new devDependency to help with unit testing of components `@testing-library/jest-dom/extend-expect`.

Three new unit tests has been created for the profile components in `test/components/Profile` which tests for component renderings, `ProfileComponent.test.jsx`, `ProfileImageField.test.jsx`, and `ProfileInputField.test.jsx`.

## Future Considerations

- The UX/UI would still need to be updated for the profile pages.
- Set/Request permission forms would need to be converted into buttons/modals
- Create request permission function for Documents and document files


timbot1789 (commented):
Why does the client profile need to be different than the current profile page? That doubles the amount of code we need to maintain for 2 features that seem quite similar. The only thing that really changes is editing permission, not the data itself.

There are 2 other issues I see:

1. Why do you need to select a client's box, then click the pod url link to open a client's profile? That seems like unnecessary clicking
2. The client profile itself looks identical to the current documents page, save one card. Why not just put that card on the documents page, or replace the documents page?

xscottxbrownx:
> Why does the client profile need to be different than the current profile page? That doubles the amount of code we need to maintain for 2 features that seem quite similar. The only thing that really changes is editing permission, not the data itself.
> 
> There are 2 other issues I see:
> 
> 1. Why do you need to select a client's box, then click the pod url link to open a client's profile? That seems like unnecessary clicking
> 2. The client profile itself looks identical to the current documents page, save one card. Why not just put that card on the documents page, or replace the documents page?

These were all the same thoughts I had when looking over this.   

I thought last night we discussed the "profile" component being a reusable component that just had some conditional rendering of editing the user info, and setting permissions on the documents.

---

My thoughts on an approach would be to have a "Info component" (that houses all the profile stuff as it is now) and a "Documents component" (that houses all the document stuff as it is now) in one "Profile page". So easier to update and maintain later as well.

I think we could:
1) move our `Documents.jsx` from `routes` to `components`
2) move what we have in our `Profile.jsx` into a new component: `ProfileCard` or `ProfileInfo`
3) make `Profile.jsx` display both of the above components (1 + 2)
4) then keep updating/iterating `Documents` component to what it needs to be

leekahung:
Was thinking of keeping the PRs for this and #322 separate. Think I could address the issues raised here after merging branch for PR #322 into this branch and updating from there.

leekahung:
@timbot1789 @xscottxbrownx I've just updated the branch to include changes made originally for PR #322 and have it included here. I've also updated the PR details and the clip for showing the new feature.

I believe that addresses some of the points brought up earlier.

src/AppRoutes.jsx
```diff
@@ -52,9 +51,9 @@ const AppRoutes = () => {
       />
       <Route element={<ProtectedRoute isLoggedIn={session.info.isLoggedIn} />}>
         <Route path="/PASS/clients" element={<Clients />} />
-        <Route path="/PASS/documents" element={<Documents />} />
+        <Route path="/PASS/clients/profile" element={<Profile />} />
```
xscottxbrownx: per your video this seems to work fine - but may want to double check with react-router docs...  I could be completely wrong, but may want to do a nested route here for the `clients/profile`.

When I can a little later, I will research it as well.

src/components/Clients/ClientListTableRow.jsx
```diff
@@ -63,32 +61,24 @@ const ClientListTableRow = ({
...
-            'aria-labelledby': labelId
-          }}
-          onClick={() => {
-            handleSelectClient(client);
-          }}
-        />
-      </StyledTableCell>
+    <StyledTableRow id={labelId}>
       <StyledTableCell align="center">{client.person}</StyledTableCell>
       {/* ***** TODO: Switch this webId to being a small Notes section */}
```
xscottxbrownx: Maybe get rid of this comment now?

src/components/Documents/DocumentTable.jsx
```diff
@@ -50,7 +52,7 @@ const DocumentTable = () => {
           </TableHead>
           <TableBody>
             {documentListObject?.docList.map((document) => (
-              <DocumentTableRow key={document.name} document={document} />
+              <DocumentTableRow key={document.name} document={document} user={user} />
```
xscottxbrownx: I see we were already using `document.name` as key here.

Do we have some sort of validation that a file can't be uploaded with same name as one that already exists OR it can, but it overwrites the original - so this won't be an issue?

xscottxbrownx: Looking into it, I believe it overwrites the original.  If that is the case, then this is fine.

xscottxbrownx (approved):
I like this approach better.  More reusability coupled with a few conditional renderings...

I think this is a solid update and looks good to me.

---

### If anybody is free to give a good check over the context files specifically. They look okay to me, but I'm a little new to this concept and want somebody with better understanding to look at them thoroughly please

src/components/Clients/ClientListTableRow.jsx
leekahung: Got it. I'll remove the comment for the webId link.

src/components/Clients/ClientListTable.jsx
```diff
@@ -7,26 +7,28 @@ import TableContainer from '@mui/material/TableContainer';
...
-import { useStatusNotification } from '../../hooks';
 // Context Imports
 import { UserListContext } from '../../contexts';
 // Component Imports
 import ClientListTableRow from './ClientListTableRow';
 import { StyledTableCell } from '../Table/TableStyles';
 
 // ===== MAKE CHANGES HERE FOR TABLE HEADER / COLUMN TITLES =====
-const columnTitlesArray = ['Select', 'Client', 'WebID', 'Pin', 'Delete'];
+const columnTitlesArray = ['Client', 'Client Profile', 'Pin', 'Delete'];
```
timbot1789: This Client Profile link is a great change. It removes the Web ID from the view while still keeping its utility

image

src/routes/Profile.jsx
```diff
@@ -74,85 +86,143 @@ const Profile = () => {
...
           </Link>
         </Typography>
 
         {/* TODO: Refactor/optimize the form below once we have more input */}
         {/* fields to update profile for */}
-        <Box style={{ display: 'flex', flexDirection: 'column', gap: '15px' }}>
-          <form onSubmit={handleUpdateProfile}>
-            <Box
-              sx={{
+        {user === 'personal' ? (
```
timbot1789: This "personal" vs "public" is section is pretty large. Could we factor them out into separate components?

src/routes/Profile.jsx
```diff
@@ -74,85 +86,143 @@ const Profile = () => {
...
+          <Link
+            href={user === 'personal' ? session.info.webId : selectedUser.webId}
+            target="_blank"
+            rel="noreferrer"
+          >
+            {user === 'personal' ? session.info.webId : selectedUser.webId}
           </Link>
         </Typography>
 
         {/* TODO: Refactor/optimize the form below once we have more input */}
```
timbot1789: We can probably remove this TODO. I'm not sure is relevant anymore.

src/AppRoutes.jsx
timbot1789: We could also use query params for these routes. The current system has several client profiles mapping to the same route. That can potentially confuse browser history and other browser caching systems.

We could have the url be something like:
`www.pass.com/profile?webid="some/web/id"` or similar. Then the profile page displays whatever data is fetched from that web ID (we may need to escape the webId).

We could even use this route for the signed in user's profile (github works like this). Then instead of passing a prop to the `Profile` component, we could just check if the webId from the query param is the same as the signed in user's web ID. Something like

```
import React, { useContext } from 'react';
import { useSearchParams } from 'react-router-dom';
import { signedInUserContext} from '@contexts';
import { PersonalProfileComponent, PublicProfileComponent} from '.'

const Profile = () => {
  const { curUserId } = useContext(signedInUserContext);
  const { searchParams } = useSearchParams();
  
   return (curUserId === searchParams.get('webId')) ? <PersonalProfileComponent/> : <PublicProfileComponent />
}
```

This method may make the `SelectedUserContext` unnecessary. If so you can delete it. Less code ≈ fewer bugs

src/routes/Profile.jsx
leekahung: Yeah, I could do that. I'll see what I could do to get these elements separated.

src/contexts/UserDataContext.jsx
```diff
@@ -22,9 +23,11 @@ const UserDataContextProvider = ({ children }) => (
   <SignedInUserContextProvider>
     <SelectedUserContextProvider>
       <UserListContextProvider>
-        <DocumentListContextProvider>
-          <MessageContextProvider>{children}</MessageContextProvider>
-        </DocumentListContextProvider>
+        <UserDocumentListContextProvider>
```
timbot1789: What's the difference between UserDocumentContext and ClientDocumentContext? They seem to be copies of each other.

leekahung: Yeah, one is based on `selectedUser` and the other is based on `podUrl`.

Making separate contexts for the user's document list and the client's document list.

src/AppRoutes.jsx
leekahung: I'll take a look at dynamic ids for React Router

src/contexts/UserDataContext.jsx
timbot1789: I'm not sure how useful this distinction is. There's no situation where the user will see more than one set of documents at a time. If the user wants to change to viewing documents from a different pod, the documents context can clear out and load a different document list.

This code was working before and covered these two cases without this distinction, so I don't see the value in adding it.

src/AppRoutes.jsx
timbot1789: Well, you want to look at url query parameters. That's what this concept is called:
https://www.branch.io/glossary/query-parameters/#:~:text=Query%20parameters%20are%20a%20defined,immediately%20by%20a%20query%20parameter.

Here's how you use query parameters in react router:
https://reactrouter.com/en/6.14.2/hooks/use-search-params

leekahung: Think I got something working using the props being passed into the `ClientListTableRow` (the `labelId`) and using that as the path.

src/routes/Profile.jsx
leekahung: Got this separated.

test/utils/session-helper.test.js
```diff
@@ -31,73 +22,3 @@ describe('createResourceTtlFile', () => {
     expect(Object.keys(result.predicates)).toHaveLength(7);
   });
 });
-
-describe('getContainerUrl', () => {
```
leekahung, line 35: Will soon replace functions that still uses usernames with podUrl. Phasing out unit test for getContainerUrl and getPodUrl. Will phase these functions completely in future PR when we replace them for set permission forms.

src/utils/network/session-helper.js
```diff
@@ -91,7 +91,7 @@ export const placeFileInContainer = async (session, fileObject, containerUrl) =>
  */
 
 export const getPodUrl = (username) => {
-  const podOidcIssuer = localStorage.getItem('oidcIssuer');
+  const podOidcIssuer = ENV.VITE_SOLID_IDENTITY_PROVIDER;
```
leekahung: Temp fix so set permission forms would not break for now.

src/components/Profile/ClientProfile.jsx
```diff
@@ -0,0 +1,59 @@
...
+ */
+const ClientProfile = ({ clientProfile }) => (
+  <>
+    <Link to="/clients" style={{ textDecoration: 'none', color: 'white' }}>
+      <Button
+        variant="contained"
+        color="secondary"
+        aria-label="Back Button"
+        startIcon={<BackspaceIcon />}
+        /* Temporary solution to clear Documents List after removing permissions */
```
leekahung: Still very confused as to why I'm getting 200 OK responses from getSolidDataset even after permissions revoked.

https://github.com/codeforpdx/PASS/assets/14917816/910168c4-e579-4a0f-8568-cbce80962716

Only seem to get 403 Forbidden back for the documents list if browser refreshes, which is not an ideal solution.

src/utils/network/session-helper.js
timbot1789: I don't think the permission forms are relevant to this PR. If there's an issue with them, you can fix them in a separate PR. Either way, this isn't a good fix. It ties the app to the ID provider set in the ENV, not the one used on login. That will create more bugs in the system than it fixes.

I wouldn't get too attached to those forms anyway. They're really just demo forms, and they need to be rewritten.

src/AppRoutes.jsx
```diff
@@ -29,10 +29,12 @@ const AppRoutes = () => {
         element={session.info.isLoggedIn ? <Navigate to={path} replace /> : <Home />}
       />
       <Route element={<ProtectedRoute isLoggedIn={session.info.isLoggedIn} />}>
-        <Route path="/clients" element={<Clients />} />
-        <Route path="/documents" element={<Documents />} />
+        <Route path="/clients">
+          <Route index element={<Clients />} />
+          <Route path=":id" element={<Profile />} />
```
timbot1789: This won't work. The ID you're using is the id of the row in the clients list table. If we sort the row differently, or we add a new client, or we remove a client, those IDs will change. And a route that once referred to one client will now refer to a different client. You'll need to use the id _of the client themselves_. That would be their web ID. You can pass the web ID as a query param to the route. See my previous comment about what these can look like and how to use query params.

If we use web IDs, we won't need separate routes for the logged in user profile and other people's profiles. They can all fall under a `/profiles?webId="webId"` route.

src/components/Form/SetAclPermsDocContainerForm.jsx
```diff
@@ -84,22 +89,31 @@ const SetAclPermsDocContainerForm = () => {
       return;
     }
 
-    try {
-      await setDocContainerAclPermission(session, permissions, podUrl, otherPodUsername);
+    // Routine for setting permissions to Documents
+    if (user === 'personal') {
```
timbot1789: What are these changes?

src/components/Form/SetAclPermsDocContainerForm.jsx
```diff
@@ -84,22 +89,31 @@ const SetAclPermsDocContainerForm = () => {
...
+          clearInputFields();
+        }, 3000);
+      } catch (error) {
+        runNotification('Set permissions failed. Reason: File not found.', 5, state, dispatch);
+        setTimeout(() => {
+          clearInputFields();
+        }, 3000);
+      }
+    } else {
+      // Routine for requesting permissions to Documents
```
timbot1789: I don't think we've finalized with UI/UX what we want permission requests to look like. It will probably look different than this. I think these TODO comments confuse more than they help.

src/components/Form/SetAclPermsDocContainerForm.jsx
```diff
@@ -108,21 +122,25 @@ const SetAclPermsDocContainerForm = () => {
 
   return (
     <FormSection
-      title="Permission to Documents Container"
+      title={
+        user === 'personal' ? 'Set Permission to Documents' : 'Request Permissions to Documents'
```
timbot1789: We haven't decided how to do request permissions yet. This will confuse people new to the project, and make them think we've made a decision already.

src/components/Modals/UploadDocumentModal.jsx
```diff
@@ -108,15 +113,17 @@ const UploadDocumentModal = ({ showModal, setShowModal }) => {
...
-          <FormControlLabel
-            control={<Switch />}
-            label="Verify file on upload"
-            id="verify-checkbox"
-            value={verifyFile}
-            checked={verifyFile}
-            onChange={() => setVerifyFile(!verifyFile)}
-            sx={{ mb: 1 }}
-          />
+          {user !== 'personal' && (
```
timbot1789: I don't think this is needed. We haven't worked out who will have authority to verify documents yet. There is a workflow that may allow for self-signing. Let's remove this for now.

src/components/NavBar/NavMenu.jsx
```diff
@@ -30,6 +32,7 @@ const NavMenu = ({
   profileImg
 }) => {
   const theme = useTheme();
+  const { setSelectedUser } = useContext(SelectedUserContext);
```
timbot1789: With how these things are routing now, the `selectedUserContext` isn't actually needed anymore. The 'selected user' information is contained in the route. You select a user by navigating to their profile page. I think we can remove this, and probably the entire context.

src/components/Notification/LoadingAnimation.jsx
```diff
@@ -20,7 +21,7 @@ import Typography from '@mui/material/Typography';
  * @param {loadingAnimationProps} Props - Component props for LoadingAnimation
  * @returns {React.ReactElement} a div of what is currently loading
  */
-const LoadingAnimation = ({ loadingItem }) => (
+const LoadingAnimation = ({ loadingItem, animationType = 'linear' }) => (
```
timbot1789: How about instead of passing in an `animationType`, we just let consumers pass in an animation component?

```
const LoadingAnimation = ({loadingItem, children}) => {
  <Box>
    <Paper>
      <Typography> Loading {loadingItem}... </Typography>
      {children ? children: <LinearProgress />}
    </Paper>
  </Box>
}

...

const LoadingAnimationConsumer = () => (
  <LoadingAnimation>
    <CircularProgress />
  </LoadingAnimation>
)
```

src/model-helpers/DocumentList.js
```diff
@@ -182,5 +182,6 @@ export const loadDocumentList = async (session, podUrl) => {
     });
     docList = [];
   }
+
```
timbot1789: Let's avoid random whitespace changes like this. They clutter up code reviews.

src/components/Clients/ClientListTableRow.jsx
```diff
@@ -1,55 +1,43 @@
...
  * rows of clients from data within ClientList
  *
  * @memberof Clients
  * @name ClientListTableRow
+ * @param {clientListTableRowProps} Props - Props for ClientListTableRow
+ * @returns {React.JSX.Element} The ClientListTableRow Component
  */
-
-// determine what gets rendered in the table body
 const ClientListTableRow = ({
```
timbot1789: Now that I think about it, we can actually reduce this client table by one more column if we just make the client name cell a clickable hyperlink to their profile. Then we can remove the `client profile` column

src/components/Clients/ClientListTable.jsx
```diff
@@ -56,20 +58,15 @@ const ClientListTable = ({ setSelectedClientToDelete, setShowDeleteClientModal }
...
-                dispatch={dispatch}
-                setShowDeleteClientModal={setShowDeleteClientModal}
-                setSelectedClientToDelete={setSelectedClientToDelete}
-              />
-            );
-          })}
+          {sortedUserList?.map((client, index) => (
+            <ClientListTableRow
+              key={client.webId}
+              labelId={index + 1}
```
timbot1789: We can remove this property. The cell it was labeling has been deleted.

src/components/Form/SetAclPermsDocContainerForm.jsx
```diff
@@ -108,21 +122,25 @@ const SetAclPermsDocContainerForm = () => {
...
       state={state}
       statusType="Permission status"
       defaultMessage="To be set..."
     >
       <Box display="flex" justifyContent="center">
         <form onSubmit={handleAclPermission} autoComplete="off">
           <FormControl>
-            <Typography htmlFor="set-acl-to">Set permissions to username:</Typography>
+            <Typography htmlFor="set-acl-to">
+              {user === 'personal' ? 'Set' : 'Request'} permissions to username:
```
timbot1789: Let's remove this for now. We should talk to UX/UI about the best way to send permission requests. Either way, it's not the subject of this PR.

src/components/Profile/ClientProfile.jsx
timbot1789: Permissions isn't the topic of this PR. We can handle this somewhere else.

src/components/Profile/UserProfile.jsx
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * The UserProfile Component is a component that renders the user's profile on
+ * PASS
+ *
+ * @memberof Profile
+ * @name UserProfile
+ * @returns {React.JSX.Element} The UserProfile Component
+ */
+const UserProfile = () => {
```
timbot1789: This `UserProfile` is very similar to the `ClientProfile` component. The user profile actually displays a bit more information, and IMO is formatted nicer. How about we just use this one `UserProfile` component, and then if the profile isn't for the logged-in user, we disable editing ability? Then we have a lot less code to manage.

src/utils/network/session-helper.js
leekahung: It's just temporary. The function itself (`getPodUrl`) will get dropped soon once we moved over to using podUrls. For now, I'm going to keep this one in this PR.

src/components/Profile/ClientProfile.jsx
leekahung: Think I'll just leave it here then. I'll include a TODO comment with the comment I have there right now.

src/components/Form/SetAclPermsDocContainerForm.jsx
leekahung: These are just to differentiate between the permission forms if they're in the profile page or the client profile page.

src/components/Form/SetAclPermsDocContainerForm.jsx
leekahung: We'll eventually need a request permissions functions. We don't have to act on them now (we can save it for another PR), but could be a good reminder.

src/components/Form/SetAclPermsDocContainerForm.jsx
leekahung: We could add a comment on the set permission forms. As you mentioned, they're not finalized.

src/components/Modals/UploadDocumentModal.jsx
leekahung: I could do that. I'll have this reverted back.

src/AppRoutes.jsx
leekahung: I'll take another crack at it.

src/contexts/UserDataContext.jsx
leekahung: I'll revert this for now.

src/utils/network/session-helper.js
timbot1789: I think you'd be surprised how long temporary fixes last. If you're determined to keep these forms working, I'd prefer if you put in a lasting fix. This looks like something that can be easily forgotten about until it causes nasty bugs later.

src/AppRoutes.jsx
leekahung: Not having much luck with useSearchParam unfortunately. Managed to get the URL changed with the query (i.e. the URL becomes http://localhost:5173/clients?webId=https%3A%2F%2Fnewtestuser.opencommons.net%2Fprofile%2Fcard%23me), but couldn't manage to get the profile page to actually load.

If I tried to re-route the page with navigator or Link, the seems to query revert to null. The browser url seems to flicker, but I see the page changing to http://localhost:5173/clients?null before going back to http://localhost:5173/clients.

For now, to avoid routing based on `labelId`, I'll use the client's "webId" instead after processing it with `encodedURIComponent` (was thinking to use username, but think webId would be better).

https://github.com/codeforpdx/PASS/assets/14917816/632b9679-83b8-48d8-ae76-9e07c3e45fc5

This seems to work?

src/components/NavBar/NavMenu.jsx
leekahung: Have removed SelectedUserContext.

src/components/Notification/LoadingAnimation.jsx
leekahung: Yeah, think that would be a good idea.

leekahung: Got this to take in an animation component as a children with LinearProgress as default if no children are passed in.

src/model-helpers/DocumentList.js
leekahung: Reverting changes.

src/components/Clients/ClientListTableRow.jsx
leekahung: I can do that. I'll switch the Link to Client instead.

src/components/Clients/ClientListTable.jsx
leekahung: Got this removed and in the children as well.

src/components/Form/SetAclPermsDocContainerForm.jsx
leekahung: I think we could leave the change?

src/components/Profile/UserProfile.jsx
leekahung: Merged the two components together into a component called `ProfileComponent` to differentiate between this and the `Profile` page.

https://github.com/codeforpdx/PASS/assets/14917816/e1c1001c-3948-41db-a6a3-94118dd79405

src/utils/network/session-helper.js
leekahung: I'll make sure to keep a note of this. It'll be important for when we work on Issue #307.

leekahung:
Hey @xscottxbrownx, I know you've approved earlier, but there's quite a bit of changes since there, so I'll re-request a review.

src/components/Documents/DocumentTableRow.jsx
```diff
@@ -30,7 +35,17 @@ const DocumentTableRow = ({ document }) => {
 
   const handleShowDocumentLocal = async (urlToOpen) => {
     const urlFileBlob = await getBlobFromSolid(session, urlToOpen);
-    window.open(urlFileBlob);
+    const newWindow = window.open(urlFileBlob);
```
timbot1789: Are these changes neede for the profile page to work? If not, I think we can save memory optimizations on the documents table for another PR. There are better ways to handle this, and we dont need to hash them out here.

leekahung: No, profile can work without this. It's just that the memory used for the document page that's being opened wouldn't be free up. It'll just be a minor change really, I'll have this in another PR.

leekahung:
@timbot1789 @xscottxbrownx I've recently updated the branch based off of the recent comments. Let me know if there's still anything that needs to be addressed.

src/components/Profile/ProfileComponent.jsx
```diff
@@ -0,0 +1,168 @@
...
+        }}
+      >
+        <Box
+          sx={{
+            display: 'flex',
+            flexDirection: 'column',
+            gap: '10px'
+          }}
+        >
+          <>
```
xscottxbrownx: don't think this fragment is needed/wanted here

src/utils/network/session-helper.js
timbot1789: Let's revert this. This function has dependencies outside of just the permissions forms, and it's too difficult for reviewers to test the full effects of this change. I don't think we have any designs where the permission forms are in the final version of the app anyway, so if they're broken it's not a major issue.

It's also not clear to me what the issue you're trying to fix here is. I'd like to be able to look over it in another PR.

src/contexts/DocumentListContext.jsx
```diff
@@ -50,17 +53,17 @@ export const DocumentListContextProvider = ({ children }) => {
       setDocumentListObject({
         docList: [],
         dataset: createSolidDataset(),
-        containerUrl: `${selectedUser.podUrl}PASS/Documents/`
+        containerUrl: client ? `${client?.podUrl}PASS/Documents/` : `${podUrl}PASS/Documents/`
```
xscottxbrownx, line 56: are these correct?  Is there no `/` between the `podUrl` and `PASS`?

xscottxbrownx (commented):
This all looks good to me, but I'd like another set of eyes on:

`typedefs.js`
`utils` folder files
`test` files

src/utils/network/session-helper.js
leekahung: The issue here is we're no longer using localStorage for oidcIssuer. We're now storing the provider in the environment variable.

This function is mainly used for getting the podUrl from a username which requires the provider. As mentioned, it's just changed so that it wouldn't be broken for the mean time. When we start updating the form to use podUrl directly, this function will be removed.

By the time we transition to use podUrl implicitly based on what we plan for in Figma instead of the form, we don't need to mind this issue at all. Which is why I think it's fine as is.

src/contexts/DocumentListContext.jsx
leekahung: Yes, since we usually have the trailing "/" as part of the `podUrl`, we don't need the extra "/" between `podUrl` and `PASS` for the container URL

timbot1789 (changes requested):
This is definitely improving. I like how you're using a url ID. It saves us having to do a lot of state management. I also like the simplifications made to the clients table. That will make it easier in another PR to replace it with the cards UI/UX drew up. I also really like the design of the `ProfileComponent`. It's solid.

A few more improvements we should make:

1. Combine the `pass.com/clients/:webId` route and `pass.com/profile` route into a single route: `pass.com/profiles/:webId`. The routing doesn't need to distinguish between case manager and client here.
2. In the profile page, check if the `webId` in the URL is equal to the webId in the `SignedInUserContext`. If so, set some kind of `canEdit` bool that we use to toggle the edit features in the view. Some components we can hide entirely. This is really the only time the profile page needs to care if it's displaying a case manager or a client.
3. In the profile page, fetch the profile just using the webId. We can update `fetchProfileInfo` to take a webId directly instead of taking the session. Then we can pass the returned `profileInfo` to different components to display the data.
5. Remove the `request permissions` forms. They're non-functional, and aren't representative of any designs UI/UX has given us. We can talk to UI/UX later about a good way to request permissions for things.
6. Add unit tests

I think if we have these things, we'll have a really strong feature.

src/components/Profile/ProfileComponent.jsx
leekahung: Good catch! I'll have these removed.

leekahung:
> A few more improvements we should make:
> 
> 1. Combine the `pass.com/clients/:webId` route and `pass.com/profile` route into a single route: `pass.com/profiles/:webId`. The routing doesn't need to distinguish between case manager and client here.

Have the two routes use `/profile` as the base, but left `/profile` for the user's profile, while `/profile/:webId` becomes the profile page for a client.

> 2. In the profile page, check if the `webId` in the URL is equal to the webId in the `SignedInUserContext`. If so, set some kind of `canEdit` bool that we use to toggle the edit features in the view. Some components we can hide entirely. This is really the only time the profile page needs to care if it's displaying a case manager or a client.

Have something similar in the form of `client` to act as the `canEdit` role. With React Router Links, the Links carry with them the state for `client` which is either the client object (truthy) if clicking on the client's name from the `Clients` page or null (falsy) if trying to access `Profile` from the nav menu.

> 3. In the profile page, fetch the profile just using the webId. We can update `fetchProfileInfo` to take a webId directly instead of taking the session. Then we can pass the returned `profileInfo` to different components to display the data.

Got `fetchProfileInfo` to utilize `webId` directly.

> 4. Remove the `request permissions` forms. They're non-functional, and aren't representative of any designs UI/UX has given us. We can talk to UI/UX later about a good way to request permissions for things.

Made the permission forms exclusive for the user's profile page for the time being (permission forms remains forms for set permissions). Augmentations for potential expansions for request permissions in the permission forms has been reverted for this PR.

> 5. Add unit tests

Created three additional unit tests for the new Profile components under `test/components/Profile` (each unit test tests for possible renderings for the profile page). Included a new devDependency `@testing-library/jest-dom/extend-expect` to help with testing.

leekahung:
@timbot1789 @xscottxbrownx Alright, think I got the changes I think made. Updated the opening post for the PR. Re-send the review requests.

test/components/Profile/ProfileComponent.test.jsx
```diff
@@ -0,0 +1,93 @@
...
+    expect(editButton).not.toBeNull();
+
+    updateButton = queryByRole('button', { name: 'Update' });
+    expect(updateButton).toBeNull();
+
+    cancelButton = queryByRole('button', { name: 'Cancel' });
+    expect(cancelButton).toBeNull();
+  });
+
+  it('renders no edit button for ProfileInputFields in clientProfile is not null', () => {
```
xscottxbrownx: small typo: `in` should be `if`

test/components/Profile/ProfileImageField.test.jsx
```diff
@@ -0,0 +1,60 @@
...
+  it('renders no button with PersonIcon if clientProfile is not null and has no profile image', () => {
+    const { queryByRole, queryByTestId } = render(<MockProfileComponent mockClientProfile={{}} />);
+    const buttonElement = queryByRole('button');
+    expect(buttonElement).toBeNull();
+
+    const svgElement = queryByTestId('PersonIcon');
+    expect(svgElement).not.toBeNull();
+  });
+
+  it('renders no button with image with clientprofile is not null and has profile image', () => {
```
xscottxbrownx: small typo:
`with clientprofile` should be `if clientProfile`

test/components/Profile/ProfileImageField.test.jsx
```diff
@@ -0,0 +1,60 @@
...
+    expect(svgElement).not.toBeNull();
+  });
+
+  it('renders Remove Img button and image if clientProfile is null, but user has profile image', () => {
+    const { queryByRole, queryByTestId } = render(
+      <MockProfileComponent mockClientProfile={null} />
+    );
+    const buttonElement = queryByRole('button');
+    expect(buttonElement.textContent).toBe('Remove Img');
+
```
xscottxbrownx, line 32: should add an expect for muiAvatar like lines 54 and 55 in this test?

test/components/Profile/ProfileInputField.test.jsx
```diff
@@ -0,0 +1,44 @@
...
+    const { queryByRole } = render(<MockProfileComponent name={mockInputValue} />);
+    const inputElement = queryByRole('textbox');
+
+    expect(inputElement).toBeInTheDocument();
+    expect(inputElement).toHaveAttribute('value', mockInputValue);
+    expect(inputElement).toHaveAttribute('placeholder', mockInputValue);
+    expect(inputElement).toHaveAttribute('disabled');
+  });
+
+  it('renders ProfileInputField with name Alice', () => {
```
xscottxbrownx: don't think what's in this `it()` is accurate

xscottxbrownx (changes requested):
Few comments about things in tests...

test/components/Profile/ProfileComponent.test.jsx
leekahung: Got this fixed.

test/components/Profile/ProfileImageField.test.jsx
leekahung: Got this fixed.

test/components/Profile/ProfileImageField.test.jsx
leekahung: Included line for muiAvatar as part of that test:

```
const muiAvatar = queryByRole('img');
expect(muiAvatar).toHaveAttribute("src", "https://example.com/user.png");
```

test/components/Profile/ProfileInputField.test.jsx
leekahung: Fixed the duplicate it statement to: `renders ProfileInputField with empty string and placeholder for "No value set" if inputValue is null`

xscottxbrownx (approved):
With changes made, looks good to my eyes

src/utils/network/session-helper.js
timbot1789: Everything else in the PR looks good for now, but I'm still pretty confident this change causes a regression. Revert this and I'll approve. We can take a look at this again in a separate PR.

I'm not sure what you mean by the first sentence. `oidcIssuer` is still in local storage. the env is just used for initialization.

It's best practice to keep PRs focused on one thing. It makes it easier to test and review. A large feature PR like this is already pretty demanding. I don't want to delay the rest of this PR to review this largely tangential change.

leekahung: I didn't broke before the recent changes in Development, so I just wanted to patch it up since it's now part of the profile page.

I've found the `setItem` call in the `OidcLoginComponent` Line 17, but it doesn't seem to be setting the `oidcIssuer` correctly after login (see screenshot). Because of that, it's crashing the set permission form, so I thought `oidcIssuer` wasn't set initially, hence the motivation behind the temporary fix for `getPodUrl` and removing the `getItem` in place of the `ENV` call.

Screen Shot 2023-07-25 at 21 23 28 (2)

Think I got this fixed now by moving the setItem outside of `loginHandler` in `OidcLoginComponent`, so I could revert `getPodUrl` back.

Screen Shot 2023-07-25 at 21 33 32 (2)

Screen Shot 2023-07-25 at 21 40 05 (2)

Think this should be all good. Since it's functional again.

leekahung:
Tests breaking. Going to fix those first.

src/components/NavBar/OidcLoginComponent.jsx
```diff
@@ -37,16 +36,18 @@ const OidcLoginComponent = () => {
...
-        type="submit"
+        type="button"
         color="secondary"
         size="large"
         aria-label="Login Button"
         onClick={() => {
           loginHandler();
+          localStorage.setItem('oidcIssuer', oidcIssuer);
         }}
         onKeyUp={(event) => {
```
leekahung: `onKeyUp` doesn't seem to work even after pressing the `Enter` key even before switching the button type from "submit" to "button". Think the event is not being caught by the component. Could address this in another PR.

src/utils/network/session-helper.js
leekahung: Going to move this into a separate PR.

timbot1789 (approved):

PR #322, opened 2023-07-20
Issue 318/refactor documents to profile

### LINKED ISSUE #318: Refactoring Documents to Profile

leekahung:
# PR Details

Included changes to styling for existing profile page components and incorporate parts of `Documents` page into `Profile`. That way we could start phasing out `Documents` in favor of having everything directly in a profile page. The information here should only relates to the user themselves and their own documents in their pod storage.

https://github.com/codeforpdx/PASS/assets/14917816/346a1edf-9bf8-43b4-a895-2377470a41ae

Also created a new `UserDocumentListContext` to separate the document list between user and client. The document list found in profile will only that of the user, not of the client.

https://github.com/codeforpdx/PASS/assets/14917816/b6a8cc2b-0f5a-4ed5-8db5-a7a335446685

The document list in `Documents` will still be using the `DocumentListContext`, thus will still generate the list based on `selectedUser`.

## Future Consideration

Once both profile pages are updated to include elements of the `Documents` page, we could phase out `Documents` completely.

We'll also need to update the UX/UI of the profile page to the existing design on Figma.

### DIFF
--- a/src/components/Documents/DocumentTable.jsx
+++ b/src/components/Documents/DocumentTable.jsx
@@ -37,7 +37,7 @@ const DocumentTable = () => {
   const determineDocumentsTable = documentListObject?.docList?.length ? (
     // render if documents
     <Container>
-      <TableContainer component={Paper} sx={{ marginTop: '3rem', marginBottom: '3rem' }}>
+      <TableContainer component={Paper} sx={{ margin: '1rem 0' }}>
         <Table aria-label="Documents Table">
           <TableHead>
             <TableRow>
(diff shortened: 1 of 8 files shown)


leekahung:
Was planning to have this PR done separately from PR #321. Closing in favor of doing them together.

PR #323, opened 2023-07-20
Navbar - Messages route ----> icon button

### LINKED ISSUE #300: Profile Hamburger menu disappears when navigating to documents page
**Describe the bug**
In our mobile, small screen configuration, we have a hamburger menu on the far right side of the screen that contains profile, notifications, etc...

Screenshot from 2023-07-10 18-14-02

However, this menu disappears when you are on the Documents page:

image

It should not disappear

### LINKED ISSUE #324: Navbar and layout breaks down before limit for mobile
**Bug Description**:
As shown in the clip below, the Navbar and Layout for PASS breaks below a certain screen width and before we even reach what was decided as the lower limit for our project (320px).

https://github.com/codeforpdx/PASS/assets/14917816/7edb2910-230a-4f83-a0cb-3463c8f70ed9

**Expected Behavior**:
Layout and Navbar should be responsive up to our limit, fully filling out the screen.

**Possible Solution**:
Update Navbar, Layout, and any other potential components to be responsive down to 320px screen width.

xscottxbrownx:
### This PR:
1. Moves `Messages` from the left routes section, to the right icons/menu section.
2. Changes `Messages` from a `tab` to an `icon button` similar to `Notifications`.
3. Fixes the `Navbar` width issue when on `Documents` - Tim pointed out in Issue #300.
4. Fixes the `Navbar` layout issue - KaHung pointed out in Issue #324.
5. Removes the mobile only "Navigate To..." dropdown - just leaves the `Tabs` now that `Messages` is gone (and `Documents` will be going away too.)
6. Adds a new dev dependency `css-mediaquery` so we can test for different screen sizes properly with MUI
7. Adds tests for the 3 different navbars, plus their parent component which renders the proper one based on screen size/user's logged in status

---

#### Components
`NavBar.jsx`
`NavbarDesktop.jsx`
`NavbarLinks.jsx`
`NavbarLoggedOut.jsx`
`NavbarMobile.jsx`
`NavMenu.jsx`
#### Tests
`NavBar.test.jsx`
`NavbarDesktop.test.jsx`
`NavbarLoggedOut.test.jsx`
`NavbarMobile.test.jsx`

---

https://github.com/codeforpdx/PASS/assets/71395931/efad5cd1-c02d-4585-9aa2-b7913a790e1e


src/components/NavBar/NavBar.jsx
leekahung, line 1: Minor thing. Think it'll be good if the spacing between the three icons on the right of the NavBar to be even.

leekahung (approved):
The menus are working as stated. Code looks okay. I approve this.

src/components/NavBar/NavBar.jsx
```diff
@@ -80,8 +83,26 @@ const NavBar = () => {
               <Box>
                 <NavbarLinks />
               </Box>
+
               <Box sx={{ flexGrow: 1 }} />
+
+              {/* DESKTOP */}
```
timbot1789: Generally I think it's better to put these kinds of things into the variable names themselves, rather than comments like this. That allows the language itself to validate and enforce this information. So like instead of doing this comment, have something like:

```
<Box sx={{ flexGrow: 1}}/>
  <DesktopLayout/>
  <MobileLayout/>
</Box>
```

timbot1789 (commented):
Could you add a couple test cases to show that the nav bar renders properly? Basically that it properly renders both the mobile and desktop view when set to the proper resolutions. I can give you a hand with writing them if you need.

src/components/NavBar/NavBar.jsx
xscottxbrownx: So, more of a self-documenting code?  How/When do you determine that there's too many child components for one parent?

Like the `Navbar` would have 6 files for it in that case.  Is that fine? Too much?   I'm definitely okay with pulling things out into their own component - should I just make those components in this file?

What are the best practices here?

xscottxbrownx:
> Could you add a couple test cases to show that the nav bar renders properly? Basically that it properly renders both the mobile and desktop view when set to the proper resolutions. I can give you a hand with writing them if you need.

There is a `/test/Navbar` folder with a test for `OIDClogin component` already.  Should I make one `Navbar` test file and move that one into it (along with others I create)?  Or a different file for each thing?

---

Ideas to test so far:
- make sure `Navbar` renders each item of navbar (logo, routes, icons/menu)
- make sure it renders the icons on right (larger than mobile view)
- make sure it renders hamburger menu icon instead of the icons on right (mobile view)

@timbot1789

src/components/NavBar/NavBar.jsx
timbot1789: That's a very good question, and there's no hard and fast rule for this. One of the guides a lot of devs use is the [Single Responsibility Principle](https://stackify.com/solid-design-principles/). A single component should only be responsible for one or two things. If it gets beyond that we probably want to refactor. Currently this navbar has a few responsibilities: 
- checking if the user is logged in,
-  checking if it should render in mobile or desktop view
- actually rendering the mobile and desktop views 
- handling the logout modal.

That's a lot of things. If we can separate some of these responsibilities out into their own components, it can be easier to understand how everything works together.

For the "too many child components for one parent", you're right it sometimes makes sense to combine a couple things together. So like maybe this navbar's responsibility should be figuring out which navbar to display; the login navbar, the desktop navbar, or the mobile navbar. Then there's just some switch logic.

Another rule of thumb is how long a section is getting. Functions generally don't need to be much longer than a dozen or so lines, files don't need to be much longer than a couple hundred. The `NavBar.jsx` file is still only about 200 lines, so if you want to keep the definitions for the mobile layout and the desktop layout in the same file, I'm ok with that. 

Proper software design is a hotly debated topic and there are dozens of books on it.

---

Most IDEs also include automatic refactoring features. [VSCode's React Refactor plugin](https://marketplace.visualstudio.com/items?itemName=planbcoding.vscode-react-refactor) can help you create these components pretty quickly.

xscottxbrownx: Just found this, but haven't tested yet:
Screenshot 2023-07-21 at 10 50 06 AM

https://jestjs.io/docs/tutorial-react#dom-testing
down in the `__tests__/CheckboxWithLabel-test.js` section

Maybe we *don't* need to do this anymore 🤷🏻‍♂️

timbot1789:
> There is a `/test/Navbar` folder with a test for `OIDClogin component` already. Should I make one `Navbar` test file and move that one into it (along with others I create)? Or a different file for each thing?
> 
> Ideas to test so far:
> 
> * make sure `Navbar` renders each item of navbar (logo, routes, icons/menu)
> * make sure it renders the icons on right (larger than mobile view)
> * make sure it renders hamburger menu icon instead of the icons on right (mobile view)
> 
> @timbot1789

Yeah they can go next to the OIDC tests in the navbar folder. And those tests sould good. I'd also add:

- When not logged in, we get the login component in the nav bar

src/components/NavBar/NavBar.jsx
xscottxbrownx: Thanks for all this @timbot1789 

---

I don't mind pulling out and making `NavbarLoggedOut`, `NavbarDesktop`, `NavbarMobile` files, and then rendering the correct version in `NavBar.jsx` if you feel that's the better approach for readability/maintainability.

timbot1789: I think that's a good idea. I was having a hard time reading that file myself. If you think that creates too many files, you can just make them separate components in the NavBar.jsx file itself.

timbot1789 (commented):
I also like how there are more deletions in this PR than additions. If everything works, that's usually a sign of a good design

src/components/NavBar/NavMenu.jsx
```diff
@@ -55,13 +56,28 @@ const NavMenu = ({
...
     >
       <MenuList>
+        {/* messages */}
+        <MenuItem
+          component={Button}
+          startIcon={<EmailIcon />}
+          sx={{ display: { md: 'none' }, color: theme.palette.primary.main, width: '100%' }}
+        >
+          <NavLink
+            to="/PASS/Messages"
```
leekahung: Looks like this wasn't updated yet to the new path name. It's causing the mobile version to have a runaway useEffect.

https://github.com/codeforpdx/PASS/assets/14917816/61d1cf2b-afae-47f9-8330-c15ebb33328d

It looks like we need to make the DocumentsList responsive too, but we can save that for another PR.

xscottxbrownx: Yeah there's a separate issue for the ClientsListTable, so might as well make one for the DocumentsListTable

xscottxbrownx:
Ok, re-review...   

@leekahung @timbot1789

I edited the very top post for what it includes now

timbot1789 (approved):
This looks good to me! Nice work!

leekahung (approved):
The changes and unit tests looks fine. I approve the changes.

PR #326, opened 2023-07-23
325 enhancement add heroku config

### LINKED ISSUE #325: Enhancement: Add Heroku Config
Currently, we have a config for deploying to Github Pages, and a heroku instance.

We don't need the Github Pages deploy any more, and we need to make an update to NPM scripts to correctly launch the app on Heroku.

**Proposed Implementation (if applicable)**:
We can remove the Github Pages related actions from github actions, and we need to add an NPM start script. That's the script heroku uses to launch the app.

timbot1789:
Add config to run Pass from Heroku. Heroku itself is already set up to watch our repository, and is watching master for pushes. Heroku will see the next push to master, and redeploy our code. As long as these changes are in the push to master, the app should launch fine.

Usually, Heroku will be running master. But right now it is running this PR. It will automatically start running master again after then next push. You can see it here: https://passsmartwallet-967e217a2652.herokuapp.com/

Based these changes off the instructions in this stack overflow post: https://stackoverflow.com/questions/69444225/how-do-i-deploy-to-heroku-using-vite

Basically, Heroku runs the vite build process for us, then we serve the app with NGINX.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,22 +36,6 @@ jobs:
         with:
           path: './dist'
 
-  deploy:
-    if: github.event_name == 'push' && github.ref == 'refs/heads/Master'
-    name: Deploy
-    runs-on: ubuntu-latest
-    needs: build
-    permissions:
-      pages: write
-      id-token: write
-    environment:
-      name: github-pages
-      url: ${{ steps.deployment.outputs.page_url }}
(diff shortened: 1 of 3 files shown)


leekahung (approved):
Thanks Tim for setting up the app with Heroku! Just checked out the application site on my phone. It seems to be running fine. I approve.

PR #327, opened 2023-07-23
Free browser memory from blob

leekahung:
This PR include additions to the handleShowDocumentLocal handler function in DocumentTableRow to free up browser memory used for the file when closing the new window with the file.


leekahung:
Not sure if this is an issue for now for PASS. Could revisit in the future. Closing PR.

PR #328, opened 2023-07-24
Cleanup warnings in tests

timbot1789:
We've been getting some annoying warning messages from a couple tests on development. This PR cleans them up.

Before:
image

After:
image

See comments for exact changes.

### DIFF
--- a/src/components/Modals/DeleteClientModal.jsx
+++ b/src/components/Modals/DeleteClientModal.jsx
@@ -26,7 +26,7 @@ import FormSection from '../Form/FormSection';
  */
 
 const DeleteClientModal = ({
-  showDeleteClientModal,
+  showDeleteClientModal = false,
   setShowDeleteClientModal,
   selectedClientToDelete
 }) => {
--- a/test/contexts/SignedInUserContext.test.jsx
+++ b/test/contexts/SignedInUserContext.test.jsx
@@ -14,7 +14,6 @@ import { expect, it, afterEach, describe, vi } from 'vitest';
 import { fetchProfileInfo } from '../../src/model-helpers';
 import { SignedInUserContext, SignedInUserContextProvider } from '../../src/contexts';
 import { RDF_PREDICATES } from '../../src/constants';
-import flushPromises from '../helpers/testHelpers';
 
 const TestConsumer = () => {
   const { podUrl } = useContext(SignedInUserContext);
@@ -59,12 +58,12 @@ describe('SignedInUserContext', () => {
     });
     fetchProfileInfo.mockResolvedValue({ profileInfo: {} });
     getPodUrlAll.mockResolvedValue(['https://example.com/pod/']);
-    const { container } = render(
+    const { findByText } = render(
       <SignedInUserContextProvider>
         <TestConsumer />
       </SignedInUserContextProvider>
     );
-    await flushPromises();
-    expect(container).toMatchSnapshot();
+    const val = await findByText('https://example.com/pod/');
+    expect(val).not.toBeNull();
   });
 });
--- a/test/contexts/__snapshots__/SignedInUserContext.test.jsx.snap
+++ b/test/contexts/__snapshots__/SignedInUserContext.test.jsx.snap
@@ -1,9 +1,0 @@
-// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
-
-exports[`SignedInUserContext > fetches user data if user is logged in 1`] = `
-<div>
-  <div>
-    https://example.com/pod/
-  </div>
-</div>
-`;
--- a/test/helpers/testHelpers.js
+++ b/test/helpers/testHelpers.js
@@ -1,4 +1,0 @@
-// eslint-disable-next-line no-promise-executor-return
-const flushPromises = () => new Promise((r) => setTimeout(r));
-
-export default flushPromises;


src/components/Modals/DeleteClientModal.jsx
```diff
@@ -26,7 +26,7 @@ import FormSection from '../Form/FormSection';
  */
 
 const DeleteClientModal = ({
-  showDeleteClientModal,
+  showDeleteClientModal = false,
```
timbot1789, line 29: The Dialog Element used by MUI expects to always receive a value. If it receives `null` or `undefined` it generates a warning. If `showDeleteClientModal` isn't passed in, this would previously have been `undefined`, which causes the warning message in the test. Now we default it to false.

test/contexts/SignedInUserContext.test.jsx
```diff
@@ -59,12 +58,12 @@ describe('SignedInUserContext', () => {
...
     getPodUrlAll.mockResolvedValue(['https://example.com/pod/']);
-    const { container } = render(
+    const { findByText } = render(
       <SignedInUserContextProvider>
         <TestConsumer />
       </SignedInUserContextProvider>
     );
-    await flushPromises();
-    expect(container).toMatchSnapshot();
+    const val = await findByText('https://example.com/pod/');
```
timbot1789, line 66: I've gotten a bit more familiar with RTL, and I've discovered that the `waitFor` and `findBy` functions do everything I was trying to do with `flushPromises`. This is in general a better approach, so I've removed `flushPromises`. This also removes the weird `act` warning message we were receiving in the test. RTL properly wraps these actions for us.

test/contexts/SignedInUserContext.test.jsx
```diff
@@ -59,12 +58,12 @@ describe('SignedInUserContext', () => {
...
+    expect(val).not.toBeNull();
```
timbot1789, line 67: We were only using the snapshot to check if the podUrl was correctly rendered. We may as well just check it directly. Since we're not using the snapshot anymore, I deleted it.

test/contexts/SignedInUserContext.test.jsx
timbot1789: Since `flushPromises` is no longer being used, and it was the only thing in the `testHelpers` file, I deleted the whole `testHelpers` file.

src/components/Modals/DeleteClientModal.jsx
xscottxbrownx: Can you explain this a little more (really just to better my understanding)... the state is instantiated as `false` to begin with. It's never `null` or `undefined`, only `true` or `false` - or am I confidently incorrect?

OR did you mean to do this in the test file for this component, instead of this src file?

timbot1789: This only happened in the test. The test didn't pass in `showDeleteClientModal={false}` when rendering `DeleteClientModal`. That left `showDeleteClientModal` as `undefined` when passed to the MUI `Dialog` component. Which caused that ugly stack trace in the test output.

leekahung (approved):
Just ran through the tests with the changes, test seems to run just fine without the warnings. I approve.

src/components/Modals/DeleteClientModal.jsx
xscottxbrownx: hmmmm...   
**1)** shouldn't `showDeleteClientModal` be `true` if we want `DeleteClientModal` to render?
**2)** so why wouldn't we do this in the test file instead of the source component file?

Like here on `line 77` and `line 83`:
Screenshot 2023-07-24 at 9 04 23 PM



Again, I'm just trying to understand. I'm just not following the logic here yet.

timbot1789: If we wanted `showDeleteClientModal` to render, then yes, we would. But the relevant test case is lines 34-49:
```js
it('hides/unmounts when showDeleteClientModal is false', () => {
  const removeUser = vi.fn();
  const setModal = () => {};
  const client = { person: 'john' };
  const { queryByLabelText } = render(
    <UserListContext.Provider value={{ removeUser }}>
      <DeleteClientModal setShowDeleteClientModal={setModal} selectedClientToDelete={client} />
    </UserListContext.Provider>
  );

  const deleteButton = queryByLabelText('Delete Client Button');
  const cancelButton = queryByLabelText('Cancel Button');

  expect(deleteButton).toBeNull();
  expect(cancelButton).toBeNull();
});
```
In that one we want the opposite. We _don't_ want to show the modal. So we pass nothing (or `undefined`). The Dialog component coerces the `undefined` to `false`, but also generates a warning because it doesn't want to do the type coercion. It would rather you explicitly set the property to false. 
You can see this for yourself. Remove this change, then run test case `hides/unmounts when showDeleteClientModal is false`. You'll see the warning generate. You can then look through the provided stack trace and see what it's complaining about.

The changes here aren't major. I just found the warnings annoying.

xscottxbrownx: Ok, and we can't just pass it in? `showDeleteClientModal={false}` because then it won't render the component at all?

I guess I'm trying to understand why we cannot do whatever needs to be done in the test file (if this is only an issue for the tests) or why the state that controls whether the modal is rendered or not isn't being passed in, in the first place.

I'm missing something clearly.

I don't disagree that these changes are good or minimal, I don't even disagree that they should be done... I'm just trying to understand the approach.


Also, `App.test.jsx` fails - but I haven't looked into it yet.

timbot1789: We could pass it in. We could pass it in every time we call the component. Or we can set a default value on the component itself once and never have to think about it again. The behavior is equivalent, and doesn't cause any regressions.

And `App.test.jsx` doesn't fail when I run it. It's also passing the test run here on github. It's probably an issue with your local install.

xscottxbrownx (approved):
All questions addressed. Looks good.

PR #330, opened 2023-07-26
Issue 286/new message styling

### LINKED ISSUE #286: UI update for new message
Currently, `NewMessage.jsx` uses custom styling that does not match the Material UI theme of the app.

Replace custom styling with MUI as well as refactoring the component into a modal.

**Additional context**
The source of the custom styling, `MessageStyles.js`, is also imported in `MessagePreview.jsx` and `MessageFolder.jsx`. Those will also have to be updated before `MessageStyles.js` can be fully removed.

andycwilliams:
Refactoring the new message function into a modal and applying MUI styling.

Opening this as a draft for now due to being away for a while and wanting to iron out any kinks. I'll have more time to sort through the most recent changes later, but wanted to update everyone what I'm working on right now.


src/components/Modals/NewMessageModal.jsx
```diff
@@ -0,0 +1,210 @@
...
+            name="title"
+            id="title"
+            onChange={(e) => handleChange(e)}
+            required
+            label="Subject"
+            inputProps={{ maxLength: '24' }}
+            fullWidth
+          />
+
+          <TextField
```
timbot1789, line 151: I don't think old message is supposed to be an editable field. It's just supposed to display previous messages in the chain.

src/pages/Messages.jsx
```diff
@@ -67,14 +69,14 @@ const Messages = () => {
...
 
   return (
     <Box sx={{ display: 'grid', gridTemplateRows: '80px 1fr' }}>
       <Box sx={{ display: 'flex', padding: '20px 30px 10px' }}>
         <Button
           variant="contained"
-          onClick={() => setShowForm(!showForm)}
+          onClick={() => setShowModal(!showModal)}
           startIcon={<CreateIcon />}
           sx={{
```
xscottxbrownx: this full `sx` should be replaced with `color="secondary"` - will do the same, built in feature of `Button`

src/components/Modals/NewMessageModal.jsx
```diff
@@ -0,0 +1,215 @@
...
+      ...message,
+      [e.target.id]: e.target.value
+    });
+  };
+
+  // Handles submit (awaiting functionality for this)
+  const handleSubmit = async (e) => {
+    e.preventDefault();
+
+    if (!message.title) {
```
xscottxbrownx, line 65: I think we can make these errors a little more user friendly/readable, dropping the "value for the".

src/components/Modals/NewMessageModal.jsx
```diff
@@ -0,0 +1,215 @@
...
+            onChange={(e) => handleChange(e)}
+            required
+            label="Subject"
+            inputProps={{ maxLength: '24' }}
+            fullWidth
+          />
+          <TextField
+            margin="normal"
+            value={oldMessage && originalMessage}
+            type="previousMessage"
```
xscottxbrownx: this `type` isn't a valid type

src/components/Modals/NewMessageModal.jsx
xscottxbrownx: Also, I would say that "Previous Message" shouldn't be in this at all, unless we are replying to a message.   

This `TextField` should have `sx={{ display: oldMessage ? 'block (or whatever)' : 'none' }}`

src/components/Modals/NewMessageModal.jsx
```diff
@@ -0,0 +1,215 @@
...
+            name="previousMessage"
+            id="previousMessage"
+            label="Previous Message"
+            // helperText="Previous message"
+            // variant="filled"
+            multiline
+            rows={3}
+            InputProps={{
```
xscottxbrownx, line 162: Would have to test (when you have an old message and reply to it), but I would think using the prop `disabled` would be better than this `read-only`.  
Personally, don't like that you can still click on/focus on this input with read-only, but I admittedly don't know what is the acceptable standard for accessibility in this case.

src/components/Modals/NewMessageModal.jsx
```diff
@@ -0,0 +1,215 @@
...
+            margin="normal"
+            value={message.message}
+            type="text"
+            name="message"
+            id="message"
+            onChange={(e) => handleChange(e)}
+            multiline
+            rows={6}
+            label="Enter Message"
+            placeholder="Add a description here"
```
xscottxbrownx: Not sure if "description" is accurate for this `placeholder` - or if placeholder is needed at all. None of the other inputs have it, seem to use the label as a direction/placeholder basically.

src/components/Modals/NewMessageModal.jsx
```diff
@@ -0,0 +1,215 @@
...
+              variant="contained"
+              type="submit"
+              color="primary"
+              startIcon={<CheckIcon />}
+              fullWidth
+            >
+              Submit
+            </Button>
+          </DialogActions>
+          {error && <div>{error}</div>}
```
xscottxbrownx, line 203: I would say maybe make the error message red, and the success message green so they stand out more. But, I think Russell is working on the status notification messaging --> snackbar feature.

So, maybe make sure this modal uses the same notification components as the others - that way it'll work when that is updated/merged??

src/components/Messages/MessagePreview.jsx
```diff
@@ -6,7 +6,7 @@ import Button from '@mui/material/Button';
 import Typography from '@mui/material/Typography';
 import { StyledDate, StyledHeader, StyledPreview } from './MessageStyles';
 // Component Imports
-import NewMessage from './NewMessage';
+import { NewMessageModal } from '../Modals';
 
 /**
  * @typedef {import("../../typedefs.js").messagePreviewProps} messagePreviewProps
```
xscottxbrownx, line 13: I'm still terrible with jsDoc, but I think this is now a member of `Messages`

src/components/Messages/MessagePreview.jsx
```diff
@@ -60,7 +69,14 @@ const MessagePreview = ({ message, folderType }) => {
           </Box>
         </Box>
       )}
-      {replyMessage && <NewMessage oldMessage={message} closeForm={() => setReplyMessage(false)} />}
+      {replyMessage && (
+        <NewMessageModal
+          showModal={showModal}
+          setShowModal={setShowModal}
+          oldMessage={message}
+          closeForm={() => setReplyMessage(false)}
```
xscottxbrownx: I can't find anything about `closeForm`, I don't think it's a valid prop - but could be wrong.

I believe what is intended from this line is to run this function when the `NewMessageModal` closes.  IF that is true, this can go in the `onClose` function within `NewMessageModal.jsx` - line 110.

src/components/Modals/NewMessageModal.jsx
andycwilliams: Yeah my idea was to wait until the status notification is reworked more, but I can definitely take another look at that.

andycwilliams:
Okay, still needs work (such as how clicking inside the modal opens and closes the messages in the background) and there are a couple of console errors to work through.

For me it's a bit of a toss-up between `disabled` and `read-only`, though I probably lean towards the latter. I think it should be immediately distinguishable from input that the user needs to provide, but still legible. `disabled` makes it a bit tougher to read than necessary, in my opinion. I'm trying a different variant for `TextField` for this as well.

Here it is with `read-only`:


And here it is with `disabled` (with bonus errors for Grammarly that I've never seen before):

xscottxbrownx:
> Okay, still needs work (such as how clicking inside the modal opens and closes the messages in the background) and there are a couple of console errors to work through.
> 
> For me it's a bit of a toss-up between `disabled` and `read-only`, though I probably lean towards the latter. I think it should be immediately distinguishable from input that the user needs to provide, but still legible. `disabled` makes it a bit tougher to read than necessary, in my opinion. I'm trying a different variant for `TextField` for this as well.
> 


read-only looks good.

src/components/Messages/MessagePreview.jsx
leekahung, line 1: There seems to be an issue with the `Reply` button and `NewMessageModal`. Was able to open the message modal when clicking on `Reply` for the first time. However, when canceling out of a reply, the message modal no longer would appear in subsequent clicks to `Reply`, it simply closes the accordion for that message (see clip).

https://github.com/codeforpdx/PASS/assets/14917816/31ab2e06-cdf7-41f7-ba32-d064950d59f2

src/components/Messages/MessageFolder.jsx
```diff
@@ -28,6 +28,7 @@ import { EmptyListNotification, LoadingAnimation } from '../Notification';
  * @param {messageFolderProps} Props - Component props for MessageFolder
  * @returns {React.JSX.Element} React component for MessageFolder
  */
+
```
leekahung: The removal of the spacing was intentional. It helps JSDoc associate the comment with the component directly below it.

https://github.com/codeforpdx/PASS/assets/14917816/dc7dda30-cef2-476d-a7dd-31448728bf01

In fact, think we should also do that for the other components being modified and remove the spaces there.

timbot1789 (changes requested):
If I close a message and reopen it, I can no longer reply to it:

[Screencast from 07-27-2023 01:57:06 PM.webm](https://github.com/codeforpdx/PASS/assets/37914436/d0f10818-e12c-497e-a488-8b674fe39c89)

The video seems to be excluding my cursor, but I'm clicking on the reply button and it's just closing the message.

src/components/Messages/MessagePreview.jsx
andycwilliams: I'm missing something obvious but what's the fix for this?

leekahung: It's related to the states `showModal` and `replyMessage` and their associated set functions. To be honest, it looks like we really just need one of them.

The state `replyMessage` is supposed to be used as `showModal` and `setReplyMessage` is supposed to be used as `setShowModal`. The new state is a duplicate of those but is not configured correctly for rendering the new modal, hence the bug.

The rendering condition for the message component uses `replyMessage` but the state never resets after component closes. This was solved previously by using the closeForm prop which contains the condition for reseting `replyMessage`.

src/components/Modals/NewMessageModal.jsx
```diff
@@ -0,0 +1,220 @@
...
+import Dialog from '@mui/material/Dialog';
+import DialogActions from '@mui/material/DialogActions';
+import TextField from '@mui/material/TextField';
+import Typography from '@mui/material/Typography';
+// Utility Imports
+import { sendMessageTTL, getMessageTTL } from '../../utils';
+// Context Imports
+import { MessageContext, SignedInUserContext } from '../../contexts';
+
+/**
```
leekahung, line 19: There seems to be a mismatch between JSDoc here and JSDoc in Development.

Assuming this was supposed to be a copy of `NewMessage`, there might have been a problem during conflict resolution or git rebase. The typedef for the component prop for this component has been moved to typedefs.js. On top of that, the props have changed for this component, so we should update that.

leekahung (commented):
See comment

src/components/Messages/MessagePreview.jsx
andycwilliams: I believe this is working now. But there's another error that I can't pinpoint yet. Sometimes my messages will duplicate and yield this in the console. They're not permanent and disappear when I refresh the page.

src/components/Modals/NewMessageModal.jsx
andycwilliams: How should it be updated? I haven't messed with the `typedefs.js` file yet.

leekahung: The JSDoc over in Development for `NewMessage.jsx` looks like this:

```
/**
 * @typedef {import("../../typedefs.js").newMessageProps} newMessageProps
 */

/**
 * New Message Component - Component that allows user to write
 *  a message to another user from their inbox
 *
 * @memberof Inbox
 * @name NewMessage
 * @param {newMessageProps} Props - Props used for NewMessage
 * @returns {React.JSX.Element} React component for NewMessage
 */
```

I think you could simply copy and paste it in there.

src/components/Modals/NewMessageModal.jsx
```diff
@@ -0,0 +1,220 @@
...
+ * @name NewMessageModal
+ * @param {newMessageModalProps} Props - Props used for NewMessageModal
+ * @returns {React.JSX.Element} React component for NewMessageModal
+ */
+const NewMessageModal = ({ showModal, setShowModal, oldMessage = '' }) => {
+  const { session } = useSession();
+  const { outboxList, setOutboxList } = useContext(MessageContext);
+  const { podUrl } = useContext(SignedInUserContext);
+  const [originalMessage, setOriginalMessage] = useState(oldMessage.message);
+  const [replyMessage, setReplyMessage] = useState(false);
```
leekahung: If `replyMessage` will not be used, can be removed along with `setReplyMessage`.

src/components/Messages/MessagePreview.jsx
leekahung: The error here is caused by the extra space set between the original `RE:${oldMessage.title}` in the `NewMessageModal`.

https://github.com/codeforpdx/PASS/assets/14917816/5a5dc162-37a4-4b04-bca6-65c69d9d50be

The extra space ended up creating a slug that's not accessible to PASS causing the problem. Think I'll create an issue for a fix to make this more robust. In the mean time, restore that line to use `RE:${oldMessage.title}` instead of `RE: ${oldMessage.title}`.

src/components/Modals/NewMessageModal.jsx
andycwilliams: Alright I've copied it over. Is this alright? Or should it just be these two and not include this:
```
/**
 * messageFolderProps is an object that stores the props for the MessageFolder
 * component
 *
 * @typedef newMessageModalProps
 * @type {object}
 * @property {() => void} closeForm - The function used to trigger NewMessageModal to
 * close
 * @memberof typedefs
 */
```

src/components/Modals/NewMessageModal.jsx
andycwilliams: Should be done now. I'm still unsure what's causing the messages to be clicked when the reply version of the modal is open. It doesn't seem to happen when the regular, new message version is.

src/components/Modals/NewMessageModal.jsx
leekahung: You would have to update the JSDoc typedefs for the props since you're passing in different one from what the component uses originally

andycwilliams: Do we still need the `closeForm` function though? Isn't using MUI Modals making that unnecessary?

leekahung: You would need to include `showModal`, `setShowModal`, and `oldMessage` as `@property` inside `newMessageProps` in `typedefs.js` in place of `closeForm`. You would also have to rename the typedef to `newMessageModalProps` to match with the type being used to define the props in the modal.

From there you would simply need to import `newMessageModalProps` from that file as shown from my earlier comment. You don't need to define the typedef again in the modal. What was shown in my previous comment should be the correct JSDoc for the new modal, you just need to update the names.

On top of that, it would seem there's a typo in the description for `newMessageProps`, which should be `newMessageModalProps` and not `messageFoldProps` and it should be props for the `NewMessageModal` component, not the `MessageFolder` component.

src/components/Modals/NewMessageModal.jsx
leekahung: I'm still seeing the line in Line 48 and Line 71. Perhaps the changes were not saved or committed?

src/components/Modals/NewMessageModal.jsx
```diff
@@ -0,0 +1,222 @@
...
+/**
+ * messageFolderProps is an object that stores the props for the MessageFolder
+ * component
+ *
+ * @typedef newMessageModalProps
+ * @type {object}
+ * @property {() => void} closeForm - The function used to trigger NewMessageModal to
+ * close
+ * @memberof typedefs
+ */
```
leekahung: Line 23 to 32 is duplicating typedefs from `typedefs.js`. Can be removed.

src/components/Modals/NewMessageModal.jsx
```diff
@@ -0,0 +1,222 @@
...
+import DialogActions from '@mui/material/DialogActions';
+import TextField from '@mui/material/TextField';
+import Typography from '@mui/material/Typography';
+// Utility Imports
+import { sendMessageTTL, getMessageTTL } from '../../utils';
+// Context Imports
+import { MessageContext, SignedInUserContext } from '../../contexts';
+
+/**
+ * @typedef {import("../../typedefs.js").newMessageProps} newMessageProps
```
leekahung: The new prop name should be `newMessageModalProps`

src/components/Modals/NewMessageModal.jsx
andycwilliams: Oops. Should be done now.

src/components/Modals/NewMessageModal.jsx
andycwilliams: Alright, this should be done now. Fingers crossed.

src/typedefs.js
```diff
@@ -157,13 +157,15 @@ const React = require('react');
...
- * @exports newMessageProps
- * @typedef {object} newMessageProps
- * @property {() => void} closeForm - The function used to trigger NewMessage to
- * close
+ * @exports newMessageModalProps
+ * @typedef {object} newMessageModalProps
+ * @property {object} showModal - Boolean for showing message modal
+ * @property {React.Dispatch<React.SetStateAction<boolean>>} setShowModal
+ * - React set function for showModal
+ * @property {string} oldMessage - The previous message when using the modal to reply
```
leekahung: `oldMessage` is actually an object if passed in or an empty string if nothing is passed in to the parameter.

andycwilliams: Like this? 

Also, I'm still troubleshooting the matter of the messages being opened and closed when clicking on the reply modal.

src/typedefs.js
```diff
@@ -157,21 +157,24 @@ const React = require('react');
...
+ * newMessageModalProps is an object that stores the props for the NewMessageModal
  * component
  *
- * @exports newMessageProps
- * @typedef {object} newMessageProps
- * @property {() => void} closeForm - The function used to trigger NewMessage to
- * close
+ * @exports newMessageModalProps
+ * @typedef {object} newMessageModalProps
+ * @property {object} showModal - Boolean for showing message modal
```
leekahung: `showModal` is a boolean

andycwilliams:
When using the reply version of the modal, clicking around also opens and closes the messages in the background.

The issue is with the `handleClick` function on line 38 in `MessagePreview.jsx`.

Since that file is due for restyling anyway, we could leave this to be addressed at that point or have it be part of this PR as well.

leekahung (approved):
Have looked over it again and tested locally. Think it should be okay for merging once you decide to open the PR from a draft.

One relevant thing to note. Was wondering if you could include a function to close the message modal a few seconds after submitting a message similar to how the other modals behave after submission.

timbot1789 (approved):

PR #331, opened 2023-07-26
Issue 301 revamp status notifications

### LINKED ISSUE #301: Revamp Status Notification system
### Is your feature request related to a problem? Please describe.
Currently, all status notifications shown to the user are within the modals/forms that the user makes the action.  These should be it's own component/feature.

---

### Describe the solution you'd like
Most of the team agrees that using MUI's `snackbar` component seems to be a great approach.
We will still need some sort of indication that an action is occurring to the user (loading spinner/animation, etc.) before the snackbar with the end result is displayed.

---

### Additional context
Example of current notification system here on bottom of modal:
Screenshot 2023-07-11 at 6 42 13 AM

Example of MUI snackbar:

https://github.com/codeforpdx/PASS/assets/71395931/bd0c160c-bfd1-4069-b7a8-a8c39e581575

russfraze:
I created a reusable component called BasicNotification in the Notifications folder that renders a mui snackbar.  It accepts a severity prop which will determine the color and icon that appears in the notification. Style overrides have been added to the theme so that these colors can be configurable in the future. 

In the current state of this PR, you can see the snackbar when adding and deleting a client. It delivers the same messages that appear in the modal, and the bottom left of screen. This messaging system utilizes the already in place statusReducer and the notification helper to set the message in the logic of each component its being used in.  A mui spinner also appears in the modal as an action is being performed. 

Looking for feedback on this approach.


src/components/Modals/AddClientModal.jsx
```diff
@@ -92,6 +94,7 @@ const AddClientModal = ({ showAddClientModal, setShowAddClientModal }) => {
 
     runNotification(
       `Adding "${userObject.givenName} ${userObject.familyName}" to client list...`,
+      'succes',
```
xscottxbrownx: may be best to save `success` for when it's actually successful (line 120)? and maybe these (when it's trying to do what the user asked) are better suited for either `warning` or `info` colors?

src/components/Modals/AddClientModal.jsx
```diff
@@ -114,6 +118,7 @@ const AddClientModal = ({ showAddClientModal, setShowAddClientModal }) => {
     } finally {
       runNotification(
         `"${userObject.givenName} ${userObject.familyName}" added to client list`,
+        'error',
```
xscottxbrownx: should this be `success` instead? Adding the client was successful yes? It was still green for me even with this though.

src/components/Modals/DeleteClientModal.jsx
```diff
@@ -32,12 +33,15 @@ const DeleteClientModal = ({
...
   const { removeUser } = useContext(UserListContext);
+  const [showSpinner, setShowSpinner] = useState(false);
 
   // Event handler for deleting client from client list
   const handleDeleteClient = async (event) => {
     event.preventDefault();
+    setShowSpinner(true);
     runNotification(
       `Deleting "${selectedClientToDelete?.person}" from client list...`,
+      'success',
```
xscottxbrownx: may be best to save `success` for when the action was actually successful (line 55)? and maybe these (when it's trying to do what the user asked) are better suited for either `warning` or `info` colors?

src/components/Notification/BasicNotification.jsx
```diff
@@ -0,0 +1,27 @@
+import React, { forwardRef, useState, useEffect } from 'react';
+import Snackbar from '@mui/material/Snackbar';
+import MuiAlert from '@mui/material/Alert';
```
xscottxbrownx: change to `import Alert from '@mui/material/Alert';`

and get rid of lines 5-7

add to the `Alert` component on line 20:
`variant="filled" elevation={6}`
will give you the same result

this also allows you to get rid of `forwardRef` import as well

src/components/Notification/BasicNotification.jsx
```diff
@@ -0,0 +1,27 @@
...
+  const [open, setOpen] = useState(false);
+
+  useEffect(() => {
+    if (message !== '') {
+      setOpen(true);
+    }
+  }, [message]);
+
+  return (
+    <Snackbar open={open}>
```
xscottxbrownx: I think if you can add the slide transition on this (either the `up` or `left`) it will look super good...
https://mui.com/material-ui/react-snackbar/#change-transition

xscottxbrownx:
### This is the look I believe we all talked about, liked, and agreed upon.

My big picture thoughts:

- Need to figure out how to have the `BasicNotification` live outside of the modal, so it will stay rendered even when modal disappears.
- Probably also set the `autoclose` prop of the `Snackbar` so it goes away after certain time.  Having a clickable close icon MAY also be desired.
- I also think the spinner is a good idea. Not sure if this is exact placement for it, but haven't given it any thought. But quickly, I think both `BasicNotification` and `Spinner` might be best at the highest level - like `App.jsx`, and so we don't have to add them to a ton of files. They just always listen for these "statuses" of "action being performed" & "notification".

---

Ultimately I think this is a great direction.  You'll see a lot of small comments on the code from me ⬆️ to address/think about.

src/components/Modals/AddClientModal.jsx
xscottxbrownx: actually another option to the `warning` or `info` colors would be taking off the `variant="filled` prop (see my comment below on `BasicNotification.jsx` - but haven't thought about how this would be implemented... some sort of ternary on this prop maybe.

timbot1789 (commented):
I think the overall design is good, but I'm having issues testing it because the Add Client Modal seems to be bugged out right now.

Since we're on the topic of the add client modal and getting general first impressions, I have a first impressions PR for a new user registration flow that would actually remove this modal. Could you take a look and see how that meshes with what you're thinking? Or maybe move the demo over to the add documents modal?

Link to PR: https://github.com/codeforpdx/PASS/pull/332

src/theme.js
```diff
@@ -29,6 +29,29 @@ const theme = createTheme({
       main: '#64748B',
       contrastText: '#fff'
     }
+  },
+  // color properties TBD for status snackbar severieties
+  components: {
```
timbot1789, line 34: Could you go over your thoughts for this theme a bit more? I'm a little confused how it's conceptually different from the "status" section higher up.

timbot1789 (commented):
I also like the idea of a single basic notification that's used everywhere, but I think we can simplify the architecture a bit. Currently in this design it seems the notification component needs to be placed on every element that will be firing off notifications. I think we actually only need one global notification across the entire app.

We can place a `<Notifications />` component at the base of the App (maybe in the layout component?) which by default is hidden. Then we have a hook that allows other forms to fire off relevant notifications as they want. So `AddClientModal`, instead of having a `BasicNotification` at its base, just looks something like this:

```js
const AddClientModal = () => {
  const notify = useNotification();

  const submitClient = async (e) => {
    client = e.target.data;
    try {
      await addClientToPod(client)
      notify(severity: 'success', message: 'client successfully added to pod')
    } catch (error) {
      notify(severity: 'error', message: `Error when adding client to pod. Reason: ${error.message}`)
    }
  }

  return <Dialog>
    <Form onSubmit={submitClient}>
      ...Form Junk
    </Form>
  </Dialog>
}
```

Kind of like what's described in this tutorial: https://hackernoon.com/using-reactjs-and-context-api-to-build-a-custom-toast-notification-component

russfraze:
I'll be taking all these comments on board. Creating another iteration of this PR soon.

src/components/Notification/BasicNotification.jsx
russfraze: This is a great change. Was there some documentation that lead you to make this change? I just grabbed the starter code from the Mui Customizable section and started editing that but I'm trying to get a better grasp on the why.

xscottxbrownx: At the bottom of each MUI component's documentation is a link to that particular API and what props it can take... the `variant` prop was there, and I think I just tried the `elevation` prop - it's not real clear but a lot of the props are "system-wide" utility ones that can be used.  Here's the `Alert` api doc: https://mui.com/material-ui/api/alert/

src/theme.js
xscottxbrownx: The way I understand it,  the top section of `theme` is mostly `palette`(line 5) - which is general colors.
this new section is `components` where we can override the base theme of any components we want, so we don't have to say `color='primary'` every time. It will just put that primary color on it automatically.

So here, it looks like Russ is overriding the theme colors used for specifically the `MuiAlert`.  So we can make the specific type of `Alerts` any color we wish by setting it once here and they will automatically be that. 

So if UX team says success alerts should be purple, and error alerts should be pink - we set that here and anytime we put `<Alert severity='success'>` in the codebase, it will be purple. No need to do `<Alert severity='success' color='purple'>`

xscottxbrownx:
> I also like the idea of a single basic notification that's used everywhere, but I think we can simplify the architecture a bit. Currently in this design it seems the notification component needs to be placed on every element that will be firing off notifications. I think we actually only need one global notification across the entire app.
> 
> We can place a `<Notifications />` component at the base of the App (maybe in the layout component?) which by default is hidden. Then we have a hook that allows other forms to fire off relevant notifications as they want. So `AddClientModal`, instead of having a `BasicNotification` at its base, just looks something like this:
> 
> ```js
> const AddClientModal = () => {
>   const notify = useNotification();
> 
>   const submitClient = async (e) => {
>     client = e.target.data;
>     try {
>       await addClientToPod(client)
>       notify(severity: 'success', message: 'client successfully added to pod')
>     } catch (error) {
>       notify(severity: 'error', message: `Error when adding client to pod. Reason: ${error.message}`)
>     }
>   }
> 
>   return <Dialog>
>     <Form onSubmit={submitClient}>
>       ...Form Junk
>     </Form>
>   </Dialog>
> }
> ```
> 
> Kind of like what's described in this tutorial: https://hackernoon.com/using-reactjs-and-context-api-to-build-a-custom-toast-notification-component

Yes, this... I think this is much better approach.   I would also debate this for the spinner too honestly, but we can all discuss that later.

leekahung:
I've noticed that the new snackbars follows a similar pattern as before with StatusNotifications with two messages (one for the action being performed and one for action completion/error).

Was wondering if we could perhaps separate these into their own snackbar (i.e., have snackbar stack on top of one another) or streamline the notifications so that it only shows one notification, like for adding clients, show only "<client_name> added" for success or "Add <client_name> failed" for error.

russfraze:
I think I got this working the way we want. I've implemented it in the DeleteClient and AddClient modals again when clients are added and deleted. 

I streamlined it to only have a success message on Ka Hung's suggestion. I think that works better than having two notifications stacked on top, especially two different kinds (an info for the process running and a success when its finished). The deleting.. or adding.. message pops up only a split second before the success, and the user only needs to know if what they did worked or not, in my opinion. 

Let me know if this approach is looking good and I can delete some of the unused code from the old messaging system, and implement the notifications everywhere one is needed.

andycwilliams:
Haven't looked closely at the code yet but just wanted to mention quickly that it seems like this is an older version of Development. Have you pulled and merged the latest one?

src/components/Modals/DeleteClientModal.jsx
```diff
@@ -30,35 +32,28 @@ const DeleteClientModal = ({
...
     } catch (e) {
       runNotification(`Client deletion failed. Reason: ${e.message}`);
     } finally {
-      runNotification(
-        `"${selectedClientToDelete?.person}" deleted from client list...`,
-        5,
-        state,
-        dispatch
-      );
+      notify.addNotification('success',`"${selectedClientToDelete?.person}" deleted from client list...`)
```
andycwilliams: The `deleted` notification has an ellipsis at the end whereas the `added` one does not. Not vital but I'd personally like to have it be consistent. I prefer the version without an ellipsis.

russfraze: I pushed this and then noticed it wasn't up to date, then pulled from Development and I have merge conflicts. I'm going to try and resolve in the morning.

andycwilliams: Alright. I'll postpone reviewing until that's ready.

src/theme.js
timbot1789: This will make it a bit confusing to reuse the colors across different components. We may use these warning / error / info colors in e.g. form validations.

MUI palettes suggest having `warning`, `error`, and `info` fields as part of the palette itself. I think we can structure the theme that way, and all our cases should be covered.

Take a look at this link: https://mui.com/material-ui/customization/palette/#default-colors

timbot1789 (commented):
I think you have a solid outline here Russell. Nice work.

It seems that a lot of the comments in this PR have to do with specifics around the implementation in AddClientsModal and DeleteClientsModal. I don't think it's necessary to delay the notifications components themselves to deal with the implementation on the individual modals.

Can you split this up into multiple PRs? I'm thinking probably 3:

- The notifications component + tests
- The Add Client modal + tests
- The delete client modal + tests

Then we can just review each one, and merge the code in as each smaller piece gets completed.

russfraze:
@timbot1789 Yes, I will work on only the notification components and corresponding tests on this branch moving forward. I've reverted the delete and add client modals back to their state in the latest Development branch. 

I'm currently attempting to refactor the code in some of my components to get rid of the circular dependency errors I'm getting when I try to push this branch.

russfraze:
@timbot1789 I could use some advice. 

I'm getting errors on this brach:
Screen Shot 2023-08-01 at 1 42 18 PM

I understand that these are circular dependency errors, the first one for example is happening because BasicNotification is imported by NotificationContainer (the component that maps and renders BasicNotifications) and then the NotificationContainer is imported by my NotificationContext. The problem occurs because the original BasicNotification also imports the context so that is has access to the delete function. 

I'm not sure how to refactor this code so that it gets rid of the error but still works.

leekahung:
> I'm not sure how to refactor this code so that it gets rid of the error but still works.

Hey @russfraze, we could start dealing with the issue starting with the errors in `NotificationContext.jsx`, we don't need to pass the `NotificationContainer.jsx` here. We could have it instead in `Layout.jsx` as part of the application layout, where you could pass in the `state` from that context with `useContext(NotificationContext)`.

You could also remove the useMemo warning inside `NotificationContext.jsx` by passing the provider value with a memoized object similar to what's done in the other contexts.

Both of these would clear up all errors within `NotificationContext.jsx`.

Next, you could include the export of `NotificationContainer.jsx` as part of `src/component/Notification/index.js`, so that you could use that for `Layout.jsx`.

With these fixes it should be cleared of dependency cycle warnings. All you'll be left with just one lint error within `notificationReducer.js`, but that's a separate kind of error.

Screen Shot 2023-08-01 at 16 32 05

src/components/Form/FormSection.jsx
```diff
@@ -38,6 +47,7 @@ const FormSection = ({ title, state, statusType, defaultMessage, file = null, ch
       {title}
     </Typography>
     {children}
+    {showSpinner && <CircularProgress />}
```
leekahung, line 50: Quick question, what could this spinner be used for?

And if there's a reason for this, how would this be different from utilizing the `LoadingAnimation` component with the circular spinner?

src/reducers/notificationReducer.js
```diff
@@ -0,0 +1,18 @@
+const notificationReducer = (state, action) => {
+  switch (action.type) {
+    case 'ADD_NOTIFICATION':
+      return {
+        ...state,
+        notifications: [...state.notifications, action.payload]
+      };
+    case 'DELETE_NOTIFICATION':
+      const updatedNotifications = state.notifications.filter(
```
leekahung: Lint error could be removed by moving `state.notifications.filter((notification) => notification.id !== action.payload)` to inside the return call without declaring a new `const` inside the switch case.

https://github.com/codeforpdx/PASS/assets/14917816/1d62ed4b-12a7-49c4-85ab-d261a24941c2

russfraze: @leekahung Thanks Ka Hung. That helped me get all errors fixed.

leekahung:
Quick thing, be sure to keep your branch up to date with the base branch in Development.

Screen Shot 2023-08-02 at 18 37 52

src/layouts/Layout.jsx
leekahung, line 1: One thing to note, I'm no longer seeing the snackbar popping up like before. I'm wondering if it's due to `state.notification` not being updated.

src/layouts/Layout.jsx
```diff
@@ -4,13 +4,18 @@ import React from 'react';
...
+import useNotification from '../hooks/useNotification';
 // Component Imports
 import { NavBar } from '../components/NavBar';
 import { InactivityMessage } from '../components/Notification';
 import Footer from '../components/Footer/Footer';
+import NotificationContainer from '../components/Notification/NotificationContainer';
 
 const Layout = ({ ariaLabel, children }) => {
   const { session } = useSession();
+  const notify = useNotification();
```
leekahung: You could directly call the `state` from the `useNotificiation` hook with destructuring:

`const { state } = useNotification();`

src/contexts/NotificationContext.jsx
```diff
@@ -0,0 +1,36 @@
+import React, { createContext, useReducer, useMemo } from 'react';
+import notificationReducer from '../reducers/notificationReducer';
+
+export const NotificationContext = createContext();
```
leekahung: The `createContext` needs to be initialized with an initial state:

```
const initialState = { notifications: [] };
export const NotificationContext = createContext(initialState);
```

See `UserListContext.jsx` as an example. Or simply set the initial state without declaring it:

```
export const NotificationContext = createContext({ notification: [] });
```

See `MessageContext.jsx` as an example.

src/layouts/Layout.jsx
russfraze: I took the implementation out of the add and delete client modals. I think thats why the snackbar is not showing up. I put it back in to the add client modal so you can see notifications working when you add a client now. I'm apologies for not noting that change.

src/components/Modals/AddClientModal.jsx
```diff
@@ -36,6 +37,7 @@ const renderWebId = (username) => {
 
 const AddClientModal = ({ showAddClientModal, setShowAddClientModal }) => {
   const { state, dispatch } = useStatusNotification();
+  const notify = useNotification();
```
leekahung: You could destructure the `addNotification` function directly from the `useNotification` hook without needing to called it from `notifiy.addNotification`.

src/components/Notification/BasicNotification.jsx
```diff
@@ -0,0 +1,37 @@
+import React, { useRef, useEffect } from 'react';
+import Stack from '@mui/material/Stack';
+import Snackbar from '@mui/material/Snackbar';
+import Alert from '@mui/material/Alert';
+import useNotification from '../../hooks/useNotification';
+
+const BasicNotification = ({ severity, message, id }) => {
+  const notify = useNotification();
```
leekahung: Similar with `addNotification`, you could destructure the `remove` function from `useNotification`.

src/components/Modals/AddClientModal.jsx
```diff
@@ -117,6 +119,7 @@ const AddClientModal = ({ showAddClientModal, setShowAddClientModal }) => {
         state,
         dispatch
       );
+      notify.addNotification('success', `this is a test`);
```
leekahung: Rather than using the statement: `this is a test`, we could use the notification message from the notification state from the hook.

leekahung:
Hey Russell, was wondering about the progress to this draft. Let us know what you think.

russfraze:
@leekahung I have time to work on this today and tomorrow. I'm trying to add tests to the NotificationsContainer component and the context. I will be at the work session tomorrow night if that would be a good time to regroup on where I'm at. If you have something specific you would like me to address before then, let me know.

leekahung:
> @leekahung I have time to work on this today and tomorrow. I'm trying to add tests to the NotificationsContainer component and the context. I will be at the work session tomorrow night if that would be a good time to regroup on where I'm at. If you have something specific you would like me to address before then, let me know. 

Don't really have anything specific, but check out the feedback that's not yet been resolved.

russfraze:
@leekahung when I try to destructure the remove function for the useNotification in basicNotification, it makes my test fail which confuses me because the function still works (still removes snackbar after 8 seconds) which means it must be importing. 

Screen Shot 2023-08-09 at 3 19 11 PM

src/components/Notification/BasicNotification.jsx
russfraze: @leekahung when I try to destructure the remove function for the useNotification in basicNotification, it makes my test fail which confuses me because the function still works (still removes snackbar after 8 seconds) which means it must be importing. 

Screen Shot 2023-08-09 at 3 19 11 PM

src/components/Modals/AddClientModal.jsx
```diff
@@ -117,6 +119,7 @@ const AddClientModal = ({ showAddClientModal, setShowAddClientModal }) => {
         state,
         dispatch
       );
+      addNotification('success', `this is a test`);
```
leekahung: I think strings like this is a test would be useful in the unit tests or console logs for debugging, but I'm still a bit confused as to why we're not just using the notification stored in `useNotification`?

src/theme.js
russfraze: Sorry for the late reply on this. Would the answer be to keep the Mui alert overides for the convenience of just being able to set an alert type and have it be the correct color, but then also storing the status colors in the status object for use in other places? The colors would be repeated, but it would be nice not to have to set the colors on every alert as well.

src/components/Modals/AddClientModal.jsx
russfraze: Can you help me understand what you mean by this? The string is just testing that the addNotification function passes the string its given and the notification displays it. The useNotification hook just gives access to the notification context. Are you saying hard code a notification in the notifications array in the context? I think there is something simple I am missing here.

leekahung: Earlier when you started to implement the snackbar, you were able to display the different notifications being used for the modal when adding and removing clients into the snackbar. Messages like "Alice Test added to client list" or "Deleting Alice Test from client list" with the name being pulled from the user you're going to delete or add, so not exactly hardcoded.

I'm a little confused as to why we've switch to utilizing test strings like this. Could you not just use what was in the `runNotification`? In this case with `AddClientModal`, I believe `addNotification` is a replacement for `runNotification`.

leekahung: See clip for what I've meant:

https://github.com/codeforpdx/PASS/assets/14917816/47521df8-a40c-4c38-bdc3-d7d372297722

In this case, I've also included the changes for `DeleteClientModal.jsx` to show the delete notification using the snackbar.

test/components/Notification/BasicNotification.test.jsx
```diff
@@ -0,0 +1,69 @@
...
+  );
+
+  act(() => {
+    vi.advanceTimersByTime(8000);
+  });
+
+  expect(remove).toHaveBeenCalledOnce();
+});
+
+it('calls remove with the correct id', () => {
```
timbot1789, line 50: This test looks really good as is. Nice work. 

The `act` call is an interesting one, but seems to be doing exactly what you intend, so I'm fine with it.

You may also consider looking over react testing library's `waitFor` feature:
https://testing-library.com/docs/dom-testing-library/api-async/

They may be easier to work with than `act`

timbot1789 (approved):
I think all this code looks good Russell. Can you move it out of `draft`? Then we can approve and merge it.

russfraze:
I'm working on deleting some of the components from the older notification system.

src/components/Form/SetAclPermissionForm.jsx
```diff
@@ -65,8 +67,15 @@ const SetAclPermissionForm = () => {
         state,
         dispatch
       );
+      addNotification(
+        'info',
```
leekahung: Not too important at the moment with regards to coloring since that could be handled later, but I think we could use the color for 'success' for this snackbar.

src/components/Form/SetAclPermsDocContainerForm.jsx
```diff
@@ -62,8 +64,15 @@ const SetAclPermsDocContainerForm = () => {
         state,
         dispatch
       );
+      addNotification(
+        'info',
```
leekahung: Similar comment to the other permission form about using 'success'

leekahung (approved):
Have just tried out the new changes. They look pretty good. Just some minor comments, but not too urgent.

Outside of the test that needs to be patched up for the checks, I'm good with approve this.

With regards to removing the older notification system, like what we've discussed with @xscottxbrownx, think we could have this done in a separate issue/PR if you'd like.

xscottxbrownx:
@timbot1789 I tried and tried but not seeing the issue or solution really. Can you check into these tests/code?  It seems that now the `DeleteContactModal` test is erroring, but I see no issues.

src/components/Form/SetAclPermissionForm.jsx
xscottxbrownx: Adding onto this. I was also confused - but that's due to the code that was already there.   

So the fix for this is yes to make it `'success'` but also change the words to `Gave` and `Revoked` - I didn't initially realize this is after it is successfully completed.

src/components/Form/SetAclPermsDocContainerForm.jsx
xscottxbrownx: Adding onto this. I was also confused - but that's due to the code that was already there.   

So the fix for this is yes to make it `'success'` but also change the words to `Gave` and `Revoked` - I didn't initially realize this is after it is successfully completed.

xscottxbrownx: The more I think about it... this line: 
`${permissions.read ? 'Give' : 'Revoke'} permission to ${permissionState.podUrlToSetPermissionsTo} for Documents Container.`
        
Should be:
`${permissions.read ? 'Gave' : 'Revoked'} permission ${permissions.read ? 'to' : 'from'} ${permissionState.podUrlToSetPermissionsTo} for Documents Container.`

src/components/Form/SetAclPermissionForm.jsx
xscottxbrownx: The more I think about it... this line:
`${permissions.read ? 'Give' : 'Revoke'} permission to ${permissionState.podUrlToSetPermissionsTo} for ${docType}.`

Should be:
`${permissions.read ? 'Gave' : 'Revoked'} permission ${permissions.read ? 'to' : 'from'} ${permissionState.podUrlToSetPermissionsTo} for ${docType}.`

src/components/Modals/DeleteContactModal.jsx
```diff
@@ -57,8 +59,13 @@ const DeleteContactModal = ({
         state,
         dispatch
       );
+      addNotification(
+        'success',
+        `"${selectedContactToDelete?.person}" deleted from contact list...`
```
xscottxbrownx: maybe just period instead of ellipsis

russfraze:
This should be good to go aside from the test error. If anyone had any ideas on what to do about that... holler.

timbot1789:
> This should be good to go aside from the test error. If anyone had any ideas on what to do about that... holler.

I'm guessing these tests started failing after a rebase? You probably brought in changes from a refactor of the clients/contacts page that weren't fully compatible. It looks like you need to wrap the components in the `DeleteContactModal.test.jsx` in your new notification provider.

russfraze:
With this merge, the app now has a new notification system using the MUI snackbar component.

timbot1789:
Sweet. Love it

PR #332, opened 2023-07-26
Add new user registration flow

### LINKED ISSUE #213: Make the "Add User" section also create a Web ID and pod for the new user
Currently, we are only able to effectively register users who have created a Web ID and pod through a separate workflow. This will make it difficult for case managers to register clients, as it means the case manager can't complete the signup themselves. They'll need to make the client perform a few steps.

Ideally, PASS should be able to create a pod and a web ID for the client. That way, the case manager just needs to click a "Register Client" button after filling in a few fields on a form, and the system creates everything for them.

Another alternative would be to have the case manager select an identity provider, and then route them to the identity provider's own signup field. This is the most general case, and we may need to implement it at some point. But for the purposes of a demo and an MVP, we can assume we are using an identity provider and solid server that allows us to create web IDs and Pods through code.

**Additional Context**
We'll also need to sign into the user's pod to generate the ACL permissions we need, and initialize any other data for PASS usage.

timbot1789:
I've been working on a user registration flow for a while, and I've brought it up in dev meetings a couple times. Want to put up a draft PR to get people's thoughts on this design for user registration. 

This isn't finalized, and I don't want to merge this PR as-is. If everyone's OK with this direction, I'll split this up into a series of smaller PRs that have unit tests. If not, I'm happy to change any aspect of the design

---

Idea here is to create a new registration flow that can create pods for users. Specific goal is to create a single registration flow that:

1) Is usable by both Case Managers and private individuals. That way we don't need to maintain multiple different flows
2) Creates new pods and web IDs for people who don't have any already
3) Allows people to use already existing pods issued by other pod providers
4) Sets up the individual to work with the case manager. This means adding the individual to the case manager's list of contacts, and granting the case manager permission to access different parts of their pod.
5) Launches into a data collection flow where we can ask for key pieces of information from an individual (name, DOB, veteran status, etc). This isn't present in this demo, but we can dynamically start the flow after successful pod registration.
6) Is quick to use and simple to understand, so the case manager can set the pod up and then transfer the individual to other services

---

### Starting the Flow 
The registration flow can be accessed one of two ways:

Case Managers can access it themselves by clicking the 'Add Client' button on the 'Clients' page:
image

Clients can register themselves by getting a 'signup' link that they can use to register with a specific case worker. This link could be sent in a text message, posted on a publicly available site, or communicated any way the organization finds best. Each case manager has their own signup link, which is currently in their profile:
image

### Registration Page

Both methods take you to the same registration page. Clicking the 'Add Client' button opens the page in a new tab (or window depending on browser settings)

image

The wording for the page currently assumes you are joining as an individual, but we could customize it to give text for a case manager.
image

Note: we go to a separate page instead of a modal because I expect this to continue into a long, multiscreen registration flow. Such a flow could be annoying to do in a modal.

### How to register

The registration form presents 2 ways to register:

1) Create a new pod. This form requests the bare minimum information needed by the pod server for it to initialize a pod. I have set this form up currently to work with community solid server, which requires an email and a password.

image

Note: different pod providers can require different information to create a pod. If we want to create a pod against something other than community solid server, we'll need a dedicated form for that.

2) Sign up with an existing pod. This requires less information. Simply provide your original pod provider URL, and you will be navigated to the pod provider's login screen.
image

### Registering

Both methods will result in PASS logging into the client's pod:
image

From there PASS will perform 3 setup steps:

1) Create a PASS container in the client's pod
2) Grant the case manager read-write access to the PASS container, and default read-write access to everything inside it.
3) Add the client to the case manager's client list

If all three steps succeed, a message is displayed saying that the client has successfully registered with PASS. At this point the client can sign into pass and interact with it to their heart's content. And the case manager can communicate with the client through PASS.

In the registration itself, we could use the same check we use to display the successful registration message to instead launch the view into a data collection flow where we can get any information we want from the client.

---

Notes:

- The UI is not finalized. I am not suggesting we use any particular aspect of it. I just drew up one that looked clean enough and communicates every aspect of the flow. We can get UI/UX involved with designing a strong UI.
- The code itself is not finished. There is no error handling or edge case validation. Although I do expect the final product to have a similar structure. Namely, I expect we will need to handle the session with the newly-created pod manually in code.


xscottxbrownx:
### Question...

Do we need to setup these tiers (organization, caseworker, client) first to really get this flow implemented the way you describe?

---

Seems like a great opportunity for a version of this: https://mui.com/material-ui/react-stepper/

timbot1789:
> Do we need to setup these tiers (organization, caseworker, client) first to really get this flow implemented the way you describe?

Sort of? I think the short answer is yes, but we don't need to set up specific details. This PR assumes 2 things:

1. A pod server exists to which PASS can connect and create Pods on. For the purposes of this demo (and probably through MVP), that's CSS. So necessarily, there must be an agent (group, person, organization) who set up that server. You'll notice in the demo I added an env variable `VITE_SOLID_POD_SERVER`, which is supposed to represent the location of that server.
2. There is a person or agent with a pod that the client can connect to. This agent does not necessarily need to have a pod on the same server as `VITE_SOLID_POD_SERVER`.

Since this registration process is just about making a pod and connecting it with another pod, I don't think we need to worry about what those pods are beyond the fact that they exist. We don't need to know about case managers or organizations in detail. We just need to know that they exist.

---
> Seems like a great opportunity for a version of this: https://mui.com/material-ui/react-stepper/

And yeah, that looks like a great library for this!

leekahung:
Have looked over the registration flow again and think I've stated most of my suggestions and inputs during the meeting today with regards to a notification for un-supported Solid servers (non-CSS servers) that couldn't utilize the new "Set up your pod" registration flow.

timbot1789:
Closing. First PR will allow for pod creation

PR #333, opened 2023-07-26
Fix oidcIssuer localStorage

leekahung:
Found a bug related to localStorage.setItem for `oidcIssuer` not being triggered upon login. This PR fixes that by moving it outside of the handler function.


timbot1789 (approved):
Yeah, looks like this has never worked properly, and you just recently surfaced it. We're not setting the OidcIssuer properly on development either. This looks good to me.

PR #338, opened 2023-07-27
Add onkeyup

radhey30:
Added onKeyUp functionality on input field with the help of useRef hook.
Fixes #334

PR #340, opened 2023-07-27
Fixed Login onKeyUp

### LINKED ISSUE #334: Login onKeyUp not working
**Bug Description**:
At the moment, we have the `onKeyUp` attribute set up for the login button on PASS with the "Enter" key (OidcLoginComponent.jsx). However, it never gets triggered. Only the onClick works as intended.

Not urgent for MVP, but could be a good quality of life fix.

**Expected Behavior**:
To trigger login with the "Enter" key from the login page.

radhey30:
On login page, pressing 'enter' triggers login.
Added onkeyup event handler on textfield instead of button.
Fixees #338

### DIFF
--- a/src/components/NavBar/OidcLoginComponent.jsx
+++ b/src/components/NavBar/OidcLoginComponent.jsx
@@ -24,6 +24,12 @@ const OidcLoginComponent = () => {
         variant="filled"
         value={oidcIssuer}
         onChange={(e) => setOidcIssuer(e.target.value)}
+        onKeyUp={(event) => {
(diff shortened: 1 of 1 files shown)


timbot1789:
This fix works in principle, but we should trigger login when the `onKeyUp` event happens on the login button. Not the text field. I don't think a user would expect that hitting enter on the input field would cause them to log in. Usually that would cause them to move to a new field.

timbot1789 (approved):

leekahung (approved):
This looks fine to me. I approve.

PR #341, opened 2023-07-27
251/update inactivity message

### LINKED ISSUE #251: Update the inactivity message and add force logout on inactivity

### LINKED ISSUE #339: Inactivity Notification no longer appears
**Bug Description**:
Currently we have an inactivity notification set in place through the InactivityMessage component (InactivityMessage.jsx). However, one of the states that causes notification rendering `activeUser` no longer get set to `true` since `loggedIn` is no longer being stored in localStorage. This causes the notification to never pop-up even if `showPopup` is `true`.

**Potential Solution**:
We could simply replace the localStorage.getItem("loggedIn") with `session.info.isLoggedIn`.

brancwill:
I've updated the inactivity message with `useSession`, and created the timer function with a forced logout.

InactivityMessage


timbot1789 (approved):

xscottxbrownx (approved):
Looks good to my eyes

leekahung (commented):
Think this should be good with regards to the forced time-out, but I was wondering if we could perhaps limit the continue to just the `Continue` button itself.

If a user could trigger continue simply be clicking outside the modal, there's not really a point of the button.

brancwill:
@leekahung Done. I removed the onClose function so the user is forced to either hit `Continue` or `Logout` to close the message.

src/components/Notification/InactivityMessage.jsx
```diff
@@ -62,27 +60,43 @@ const InactivityMessage = () => {
...
+  /* Starts a five minute timer that is displayed to the user when the 
+  inactivity window pops up. If said timer runs out, the user is 
+  forcibly logged out. */
+  useEffect(() => {
+    if (showPopup) {
+      logoutTimer.current = setInterval(() => {
+        if (secondsToLogout > 0) {
+          setSecondsToLogout((prev) => prev - 1);
+        } else if (secondsToLogout === 0) {
+          logout();
```
leekahung, line 73: Think we could clear the localStorage after log out since there's no need for PASS-related items after that. It was part of the handleLogout handler function previously.

Other than that and needing to update the branch, think it's ready to be merged.

timbot1789 (approved):
Looks good to my eyes. Thanks a lot Brandon!

src/components/Notification/InactivityMessage.jsx
brancwill: Cool, I've done both

brancwill:
@timbot1789 No problem, thank you

src/components/Notification/InactivityMessage.jsx
```diff
@@ -91,7 +106,7 @@ const InactivityMessage = () => {
               variant="outlined"
               color="error"
               endIcon={<LogoutIcon />}
-              onClick={handleLogout}
+              onClick={() => logout()}
```
leekahung: Sorry Brandon, I should have been more clear in my last comment. The log out button for this component should also contain the same call for localStorage.clear() as the useEffect.

I'll approve for now, since everything else looks good. Thanks for the updates.

leekahung (approved):
Approving since most of it is here already. Just one small thing (see comment).

src/components/Notification/InactivityMessage.jsx
brancwill: Oof, sorry I should've thought about that. I've added it now, thanks for approving

PR #344, opened 2023-07-27
Update pod server script

timbot1789:
We've been getting a lot of questions about running local pod servers, so I want to add better documentation about it. And also make it easier to do, since the scripts right now are confusingly named. Here are the changes:

- Updated README with a section about setting up the pod servers
- Removed `env_templates` folder and replaced it with a single `env.template` file that new devs can copy.
- Updated server scripts to be easier to use and tell apart. Now the command `npm run pod` will launch a server that stores server files to the local file system. `npm run pod:temp` will run a server that only exists in memory. It doesn't store anything, and nothing will persist between startups.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,25 @@ PASS is currently in Development heading towards [Minimum Viable Product](./docs
    ```
 4. PASS should launch at `http://localhost:5173`. You can now visit that url, and sign into a pod hosted at the OIDC provider of your choice.
 
+- ### Setting up a Development Pod Server
+    
+   PASS is able to connect to any solid-spec compliant pod server. However, for testing, it's recommended that you run a server locally. PASS provides tools to make this easy to do.
+
+1. Clone and install dependencies. `npm i`
+
+2. In the project's root directory, copy the `env.template` file into a `.env` file. In bash you can use this command:
+    ```bash
+    cp env.template .env
+    ```
+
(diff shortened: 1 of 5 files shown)


leekahung (approved):
The instructions for starting a local server looks straightforward to me. I'm good with approving it.

leekahung:
Oh, one small thing. It wouldn’t affect the underlying script (and I'm still fine approving with this being `pod`), but I was thinking we could use `server` instead of `pod` since we are running a local Solid server that hosts a local pod.

timbot1789:
> Oh, one small thing. It wouldn’t affect the underlying script (and I'm still fine approving with this being `pod`), but I was thinking we could use `server` instead of `pod` since we are running a local Solid server that hosts a local pod.

In my mind, that's a little confusing because we have a few servers. We have `npm run dev` which runs the vite development server, `npm run preview` which runs preview server, so what server does `npm run server` run? It seems kinda vague to me.

leekahung:
> In my mind, that's a little confusing because we have a few servers. We have `npm run dev` which runs the vite development server, `npm run preview` which runs preview server, so what server does `npm run server` run? It seems kinda vague to me.

We could modify it so it's server:solid or something akin to that if we want something more targeted, `npm run server:solid`.

timbot1789:
How about `podserver`

leekahung:
> How about `podserver`

Yeah, that's good too. I like it

xscottxbrownx (approved):

PR #345, opened 2023-07-27
Issue 329 update pass color scheme

### LINKED ISSUE #329: Update PASS color scheme
To date, PASS has used a placeholder color scheme. It is fine for development but does not follow UX guidelines and will ultimately need an update. The PASS UX team is creating a theme that better suits this app's specific needs.

Replace the current color scheme with the one developed by the UX team.

**Rationale**:
A guiding principle for the app is to be trauma-informed. Colors can have effects that affect the user, such as calming them or causing additional stress. The new color scheme will incorporate that philosophy.

**Proposed Implementation (if applicable)**:
`theme.js` contains the styling which applies to the whole app. Only that file will need to be changed.

**Additional context**:
The different palette options in `theme.js` have fields for `light` and `dark`. Those are not currently part of the MVP. Updating `main` should be the focus.

russfraze:
I updated all the main colors in the theme to correspond to the colors found in the Style Guide section of the UX team's Figma document. The instructions indicate to ignore the dark and light colors for now, but it appeared that the Pass uses a color labeled "slight" in the primary category. This color is used in the table rows on the client page and documents list. I replaced that color with the "primary container" color in the primary section of the style guide. 

The colors in the style guid all have a main color, a lighter color marked as "container", a darker color and other variations as well. These aren't keys found in the theme, so if those colors are to be added in the future, please let me know. 

There are elements in the app that are still not within the new color palette, like the pagination buttons, and elements of some of the modals. Also there are some colors in the footer that are not legible and might need to be switched.

### DIFF
--- a/src/theme.js
+++ b/src/theme.js
@@ -7,7 +7,7 @@ const theme = createTheme({
       light: '#039686',
       main: '#6750a4',
       dark: '#004d3e',
-      slight: '#8fc2bb',
+      slight: '#eaddff',
       contrastText: '#fff'
     },
     secondary: {
--- a/src/theme.js
+++ b/src/theme.js
@@ -5,25 +5,25 @@ const theme = createTheme({
   palette: {
     primary: {
       light: '#039686',
-      main: '#017969',
+      main: '#6750a4',
       dark: '#004d3e',
       slight: '#8fc2bb',
       contrastText: '#fff'
     },
     secondary: {
       light: '#b32126',
-      main: '#961020',
+      main: '#625b71',
       dark: '#790111',
       contrastText: '#fff'
     },
     tertiary: {
       light: '#dbc584',
-      main: '#DEBC59',
+      main: '#7d5260',
       dark: '#dbb032',
       contrastText: '#fff'
     },
     status: {
-      danger: '#e53e3e'
+      danger: '#b3261e'
     },
     neutral: {
       main: '#64748B',


timbot1789:
Looks like you opened this PR against `Master`. Can you update it to be against `Development`?

timbot1789 (commented):
Looking at how this theme runs, it seems the UI isn't able to clearly support it yet:

image

It especially kills the text colors in the footer, making them pretty illegible. And the buttons IMO don't have enough contrast.

Could you add this as a second theme in the themes file? Then we can switch between them while developing so we can get the UI to work with both. Maybe even add a button or dropdown selector to change theme.

src/theme.js
```diff
@@ -5,25 +5,25 @@ const theme = createTheme({
   palette: {
```
timbot1789, line 5: I think we can make this file something like
```js
import { createTheme } from '@mui/material/styles';

const greenTheme = {
 ...all the green theme values
}

const purpleTheme = {
  ...all the purple theme values
}

export default createTheme(greenTheme)
```

leekahung:
I very much like having a purple/gray color scheme, but was thinking we could perhaps have a lighter shade for gray so that we could have a larger contrast.

Think we should also consider updating the color scheme for the hover state so that it uses a different shade (similar to what's done with the pagination buttons in the `Messages` page) as oppose to a different color.

As for the gold(?)ish color, I agree with Tim that we should use a different color. The low contrast makes it hard to read. We could perhaps use just white, much like what is presented in the surrounding texts and icons.

andycwilliams (commented):
To build on previous comments, I do think the gray looks nice enough with the purple, but I wonder if it makes buttons look disabled or as though they cannot be clicked. This could be especially confusing for whenever we do have buttons that are not able to be clicked for whatever reason (such as another action needed to be accomplished first).

And yeah the dark gold is too difficult to distinguish. White would look fine enough but maybe even just a lighter shade of gold would be worth trying.

PR #349, opened 2023-07-29
Issue 307/use webID to set permissions

### LINKED ISSUE #307: Change permission forms to use pod URLs explicitly instead of usernames

xscottxbrownx:
## This PR is the first step of what is needed for the "set permissions" forms ( Closes #307 )

---
## This PR:
1.  Changes from having a user type in a username on these forms to typing in a podUrl
2. The form file itself converts that podUrl to a webId before passing it into the functions from `session-core.js`
3. Will error and not submit form until inputs are filled
4. podUrl input will error if user types in their own podUrl and not submit form

---
### Components
`SetAclPermissionForm.jsx`
`SetAclPermsDocContainerForm.jsx`

### Tests
`session-core.test.js`

### Other Files
`session-core.js`

---
Decided to go the route of leaving the button enabled, but fields will error until filled out.
This was due to all other forms/modals being this way.  In the `Upload Document Modal` the `upload` button is disabled, but once you choose a file it is enabled whether or not the rest of the required fields are filled in - it relies on MUIs checking of the required attribute.  So, all current forms are now similar in this aspect.

---
**1.** Need to move these into modals/dialogs instead so they are not showing on the profile page until an onClick
          - Current idea is to make a `Set Permission for Container` button next to the `Add Document` button to trigger the container modal & add an icon or link in the document table row to trigger the document modal.
**2.** Need to update `userslist` or something similar to create a list of `contacts`
**3.** Use that list of contacts to replace the text input of podUrl with a select input of the user's contacts.

---

### DIFF
--- a/src/components/Form/SetAclPermissionForm.jsx
+++ b/src/components/Form/SetAclPermissionForm.jsx
@@ -52,7 +52,7 @@ const SetAclPermissionForm = () => {
     const permissions = event.target.setAclPerms.value
       ? { read: event.target.setAclPerms.value === 'Give' }
       : undefined;
-    const webIdToSetPermsTo = `${permissionState.podUrlToSetPermissionsTo}/profile/card#me`;
+    const webIdToSetPermsTo = `${permissionState.podUrlToSetPermissionsTo}profile/card#me`;
 
     try {
       await setDocAclPermission(session, docType, permissions, podUrl, webIdToSetPermsTo);
--- a/src/components/Form/SetAclPermissionForm.jsx
+++ b/src/components/Form/SetAclPermissionForm.jsx
@@ -149,7 +149,7 @@ const SetAclPermissionForm = () => {
             >
               {permissionState.permissionType
                 ? `${permissionState.permissionType} Permission`
-                : 'Give or Revoke Permission'}
+                : 'Set Permission'}
             </Button>
           </FormControl>
         </form>
--- a/src/components/Form/SetAclPermissionForm.jsx
+++ b/src/components/Form/SetAclPermissionForm.jsx
@@ -11,7 +11,7 @@ import MenuItem from '@mui/material/MenuItem';
 import Select from '@mui/material/Select';
 import TextField from '@mui/material/TextField';
 // Utility Imports
-import { getPodUrl, runNotification, setDocAclPermission } from '@utils';
+import { runNotification, setDocAclPermission } from '@utils';
 // Context Imports
 import { SignedInUserContext } from '@contexts';
 // Component Imports
@@ -32,7 +32,7 @@ const SetAclPermissionForm = () => {
   const { podUrl } = useContext(SignedInUserContext);
   const [docType, setDocType] = useState('');
   const [permissionState, setPermissionState] = useState({
-    username: '',
+    podToSetPermissionsTo: '',
     permissionType: ''
   });
 
@@ -51,9 +51,9 @@ const SetAclPermissionForm = () => {
     const permissions = event.target.setAclPerms.value
       ? { read: event.target.setAclPerms.value === 'Give' }
       : undefined;
-    const podUsername = event.target.setAclTo.value;
+    const podToSetPermissionsToURL = event.target.setAclTo.value;
 
-    if (!podUsername) {
+    if (!podToSetPermissionsToURL) {
       runNotification('Set permissions failed. Reason: PodURL not provided.', 5, state, dispatch);
       setTimeout(() => {
         clearInputFields();
@@ -61,7 +61,7 @@ const SetAclPermissionForm = () => {
       return;
     }
 
-    if (getPodUrl(podUsername) === podUrl) {
+    if (podToSetPermissionsToURL === podUrl) {
       runNotification(
         'Set permissions failed. Reason: Current user Pod cannot change container permissions to itself.',
         5,
@@ -91,11 +91,11 @@ const SetAclPermissionForm = () => {
     }
 
     try {
-      await setDocAclPermission(session, docType, permissions, podUsername);
+      await setDocAclPermission(session, docType, permissions, podToSetPermissionsToURL);
 
       runNotification(
         `${permissions.read ? 'Give' : 'Revoke'} permission to ${
-          permissionState.username
+          permissionState.podToSetPermissionsTo
         } for ${docType}.`,
         5,
         state,
@@ -140,9 +140,11 @@ const SetAclPermissionForm = () => {
             <TextField
               id="set-acl-to"
               name="setAclTo"
-              value={permissionState.username}
-              onChange={(e) => setPermissionState({ ...permissionState, username: e.target.value })}
-              placeholder={permissionState.username}
+              value={permissionState.podToSetPermissionsTo}
+              onChange={(e) =>
+                setPermissionState({ ...permissionState, podToSetPermissionsTo: e.target.value })
+              }
+              placeholder={permissionState.podToSetPermissionsTo}
               label="Enter PodURL"
               required
             />
--- a/src/components/Form/SetAclPermissionForm.jsx
+++ b/src/components/Form/SetAclPermissionForm.jsx
@@ -51,19 +51,11 @@ const SetAclPermissionForm = () => {
     const permissions = event.target.setAclPerms.value
       ? { read: event.target.setAclPerms.value === 'Give' }
       : undefined;
-    const podToSetPermissionsToURL = event.target.setAclTo.value;
+    const podUrlToSetPermissionsTo = event.target.setAclTo.value;
 
-    if (!podToSetPermissionsToURL) {
-      runNotification('Set permissions failed. Reason: PodURL not provided.', 5, state, dispatch);
-      setTimeout(() => {
-        clearInputFields();
-      }, 3000);
-      return;
-    }
-
-    if (podToSetPermissionsToURL === podUrl) {
+    if (podUrlToSetPermissionsTo === podUrl) {
       runNotification(
-        'Set permissions failed. Reason: Current user Pod cannot change container permissions to itself.',
+        'FAILED TO SET PERMISSIONS. REASON: Current user Pod cannot change container permissions to itself.',
         5,
         state,
         dispatch
@@ -74,24 +66,8 @@ const SetAclPermissionForm = () => {
       return;
     }
 
-    if (permissions === undefined) {
-      runNotification('Set permissions failed. Reason: Permissions not set.', 5, state, dispatch);
-      setTimeout(() => {
-        clearInputFields();
-      }, 3000);
-      return;
-    }
-
-    if (!docType) {
-      runNotification('Search failed. Reason: No document type selected.', 5, state, dispatch);
-      setTimeout(() => {
-        dispatch({ type: 'CLEAR_PROCESSING' });
-      }, 3000);
-      return;
-    }
-
     try {
-      await setDocAclPermission(session, docType, permissions, podToSetPermissionsToURL);
+      await setDocAclPermission(session, docType, permissions, podUrl, podUrlToSetPermissionsTo);
 
       runNotification(
         `${permissions.read ? 'Give' : 'Revoke'} permission to ${
@@ -102,7 +78,7 @@ const SetAclPermissionForm = () => {
         dispatch
       );
     } catch (error) {
-      runNotification('Set permissions failed. Reason: File not found.', 5, state, dispatch);
+      runNotification('FAILED TO SET PERMISSIONS. REASON: File not found.', 5, state, dispatch);
     } finally {
       setTimeout(() => {
         clearInputFields();
--- a/src/components/Form/SetAclPermissionForm.jsx
+++ b/src/components/Form/SetAclPermissionForm.jsx
@@ -10,6 +10,7 @@ import InputLabel from '@mui/material/InputLabel';
 import MenuItem from '@mui/material/MenuItem';
 import Select from '@mui/material/Select';
 import TextField from '@mui/material/TextField';
+import Typography from '@mui/material/Typography';
 // Utility Imports
 import { runNotification, setDocAclPermission } from '@utils';
 // Context Imports
@@ -53,19 +54,6 @@ const SetAclPermissionForm = () => {
       : undefined;
     const podUrlToSetPermissionsTo = event.target.setAclTo.value;
 
-    if (podUrlToSetPermissionsTo === podUrl) {
-      runNotification(
-        'FAILED TO SET PERMISSIONS. REASON: Current user Pod cannot change container permissions to itself.',
-        5,
-        state,
-        dispatch
-      );
-      setTimeout(() => {
-        clearInputFields();
-      }, 3000);
-      return;
-    }
-
     try {
       await setDocAclPermission(session, docType, permissions, podUrl, podUrlToSetPermissionsTo);
 
@@ -95,6 +83,18 @@ const SetAclPermissionForm = () => {
     >
       <Box display="flex" justifyContent="center">
         <form onSubmit={handleAclPermission} autoComplete="off">
+          <Typography
+            variant="subtitle2"
+            mb={2}
+            sx={{
+              width: '250px',
+              whiteSpace: 'nowrap',
+              overflow: 'hidden',
+              textOverflow: 'ellipsis'
+            }}
+          >
+            File Name: LongAssFileNameToTestIfThisIsWorkingCorrectly
+          </Typography>
           <FormControl required fullWidth sx={{ marginBottom: '1rem' }}>
             <InputLabel id="permissionType-label">Select One</InputLabel>
             <Select
@@ -123,6 +123,12 @@ const SetAclPermissionForm = () => {
               placeholder={permissionState.podToSetPermissionsTo}
               label="Enter PodURL"
               required
+              error={permissionState.podToSetPermissionsTo === podUrl}
+              helperText={
+                permissionState.podToSetPermissionsTo === podUrl
+                  ? 'Cannot modify your permissions to your own pod.'.toUpperCase()
+                  : ''
+              }
             />
           </FormControl>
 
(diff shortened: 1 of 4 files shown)


xscottxbrownx:
## Comments/thoughts/ideas/direction is open to anybody and everybody.

---

@andycwilliams you might have the most experience with MUI (especially forms and modals) to see what I'm doing that is causing these 2 forms to be different sizes.

@leekahung your eyes are definitely needed to check on my work about username -> podUrl, especially in `session-core.js` & probably the `session-core.test.js`

@timbot1789 would appreciate a second set of eyes on the `session-core.test.js`

src/utils/network/session-core.js
```diff
@@ -46,15 +45,19 @@ import {
...
  * document type, if exists, or null
  */
-export const setDocAclPermission = async (session, fileType, permissions, otherPodUsername) => {
+export const setDocAclPermission = async (
+  session,
+  fileType,
+  permissions,
+  podToSetPermissionsToURL
+) => {
   const containerUrl = getContainerUrl(session, 'Documents', INTERACTION_TYPES.SELF);
```
leekahung: Think we could also replace the function call for `getContainerUrl` too with `${<podUrl_of_user>}PASS/Documents/` (see example from `setDocContainerAclPermission` in session-core). We just need to pass the user's own podUrl into `setDocAclPermission` as an additional parameter.

leekahung:
> 1. On these 2 forms, I made one button show as "active" or "disabled" differently from the other.
>    
>    * The top form has the button start as disabled, and is only clickable after filling out both inputs. This idea stemmed from the similar step-by-step approach of the `UploadDocumentModal`. **BUT** this isn't _really_ necessary as both the inputs are marked as `required` meaning the form won't submit until they are both filled out anyway.
>    * The bottom form works on _that_ concept. No need to start as "disabled" because the `required` prop of the `inputs`

Don't have too strong of an opinion on either approach. Both approaches utilize the same concept of needing the user to complete before it can proceed (first option forces user to fill everything out before button can be activated to proceed; second approach forces user to fill everyone out before button let's users proceed). Think this is a matter of how much we want to lean on MUI's form validation.

src/components/Form/SetAclPermissionForm.jsx
```diff
@@ -50,17 +51,17 @@ const SetAclPermissionForm = () => {
     const permissions = event.target.setAclPerms.value
       ? { read: event.target.setAclPerms.value === 'Give' }
       : undefined;
-    const podUsername = event.target.setAclTo.value;
+    const podToSetPermissionsToURL = event.target.setAclTo.value;
 
-    if (!podUsername) {
-      runNotification('Set permissions failed. Reason: Username not provided.', 5, state, dispatch);
+    if (!podToSetPermissionsToURL) {
```
leekahung: Considering we're going to utilize MUI to help with our validation (making it a required field to input the podUrl of the other user), I think we could remove this `if` statement.

src/utils/network/session-core.js
xscottxbrownx: > Think we could also replace the function call for `getContainerUrl` too with `${<podUrl_of_user>}PASS/Documents/` (see example from `setDocContainerAclPermission` in session-core). We just need to pass the user's own podUrl into `setDocAclPermission` as an additional parameter.

Just for clarity, that should be the podUrl of the logged in user, not the podUrl of the person they want to give/revoke permissions to/from correct?

src/components/Form/SetAclPermissionForm.jsx
xscottxbrownx: > Considering we're going to utilize MUI to help with our validation (making it a required field to input the podUrl of the other user), I think we could remove this `if` statement.

really the only `if` that would need to stay is the check to see if the user put in their own podUrl, correct?  All the others will be handled by form validation.

timbot1789: That could actually be a form validation as well.

timbot1789: We generally use notifications to display server or other asynchronous errors to the user. "I tried to submit your form, but the server rejected it because you're not authorized / it's currently down for maintenance / it's a teapot." Errors the client can figure out itself can usually be displayed more directly.

src/components/Form/SetAclPermissionForm.jsx
```diff
@@ -109,51 +112,55 @@ const SetAclPermissionForm = () => {
...
+
+          <FormControl fullWidth sx={{ marginBottom: '1rem' }}>
             <TextField
               id="set-acl-to"
               name="setAclTo"
-              value={username}
-              onChange={(e) => setUsername(e.target.value)}
-              placeholder={username}
-              label="Search Username"
+              value={permissionState.podToSetPermissionsTo}
```
timbot1789: 2 things:

1. I think it's going to be very difficult for users to remember the pod urls of every person they are interacting with. I think it would be easier for users if they can instead select from a dropdown of people from their userlist (like if this were a 'contact' list like gmail lets you do with known emails). Then we automatically figure out the proper webIds and info to set.
2. Although solid uses pod urls for document upload, for permissions solid uses the webId. So this shouldn't be passing around pod urls, it should be passing around webIds.

timbot1789 (commented):
For a cleaner overall permissions design, an idea that I had was to have 2 different views:

1. A view from each individual document that shows who has access to a document, and what permissions they have. This can be a separate table / modal where the user can revoke access to each individual user
2. A view from the case manager's list (let's imagine each client has a list of associated case managers) where the user can see what documents each individual case manager has access to. Then from that view, the client can change what documents the case manager can access

This is out of scope of this PR though.

src/components/Form/SetAclPermissionForm.jsx
xscottxbrownx: > 2 things:
> 
> 1. I think it's going to be very difficult for users to remember the pod urls of every person they are interacting with. I think it would be easier for users if they can instead select from a dropdown of people from their userlist (like if this were a 'contact' list like gmail lets you do with known emails). Then we automatically figure out the proper webIds and info to set.
> 2. Although solid uses pod urls for document upload, for permissions solid uses the webId. So this shouldn't be passing around pod urls, it should be passing around webIds.

1. agreed, but where does a list of users come from? Should you not be able to grant anybody you want permission to whatever you want?    If we want a userslist, we would need some way of creating that - like maybe requesting friends or connections as in other social media.... ?    I was under the impression that the webId was not standardized, meaning any server/provider could format it really however they'd like and that's why we were moving to podUrls.

2. If that is the case, then this issue should just be closed? "Change permission forms to use podUrls explicitly, instead of usernames"

xscottxbrownx:
> For a cleaner overall permissions design, an idea that I had was to have 2 different views:
> 
> 1. A view from each individual document that shows who has access to a document, and what permissions they have. This can be a separate table / modal where the user can revoke access to each individual user
> 2. A view from the case manager's list (let's imagine each client has a list of associated case managers) where the user can see what documents each individual case manager has access to. Then from that view, the client can change what documents the case manager can access
> 
> This is out of scope of this PR though.

Agreed, this was the next step - but needed to get the giving/revoking permissions working the way we wanted it to first.

src/components/Form/SetAclPermissionForm.jsx
leekahung: For 2, you could get WebIds from pod URLs by attaching `profile/card#me` similar to how it's done with messages (see `sendMessageTTL` in session-core). You're still using podUrls explicitly and having them converted into webIds in the process. They're already being done in the two functions used for the permission forms.

src/utils/network/session-core.js
leekahung: Yeah, the logged in user, which is why it'll be included as an additional parameter. One will be used to build the path for the user's Documents container (`podUrl`), the other will be used to build the other person's webId (`podToSetPermissionsToUrl`, where the webId would be `${podToSetPermissionsToUrl}profile/card#me`).

src/components/Form/SetAclPermissionForm.jsx
timbot1789: 1. We have userslist.ttl file in the pod, and a UserslistContext in the react front end. Looking over those should answer your question
2. I didn't look over the issue itself too closely, but I think we can just word the issue better. The intent of the issue is good. We don't need a concept of usernames, so we should remove it from the permissions components. I believe the webid is stored for each user in the userlist.

What's the difference between pod urls and web ids? Pod urls for storage, web ids for identity.

timbot1789: I added a related issue about enhancing the users list here:

https://github.com/codeforpdx/PASS/issues/351

andycwilliams:
> ## Comments/thoughts/ideas/direction is open to anybody and everybody.
> @andycwilliams you might have the most experience with MUI (especially forms and modals) to see what I'm doing that is causing these 2 forms to be different sizes.

The culprit is `DocumentSelection.jsx`, which is being imported into `SetAclPermissionForm.jsx` but not `SetAclPermsDocContainerForm.jsx`. It adds another `<Box>` to the former, which alters the sizing. The latter doesn't import it so it doesn't ultimately receive the same amount of styling.

One workaround is to add `<br>` between the inputs, which may not be very elegant but fixed it when I tested it out.

---
> Bottom form improvements to come:
> - the [FileName] in title ideally would be replaced by the actual filename depending on which one the user clicked on.
> - the Select Document Type would also be unnecessary once we can connect it to the specific file.

If it's going to show the file name at the top, there should be a character cutoff point in case the name is extremely long. There's such a limit already in `UploadDocumentModal.jsx`:

```
sx={{
  width: '200px',
  whiteSpace: 'nowrap',
  overflow: 'hidden',
  textOverflow: 'ellipsis'
}}
```
---
On the `active`/`disabled` question, I'm pretty agnostic. Disabling the button looks a bit cleaner in my opinion, but MUI's default validation is also good.

xscottxbrownx:
@andycwilliams no rush - but if you can give more detail on where/which file you are putting the `<br />` and finding it fixes the two different size form components...  I can't seem to figure it out, stupidly frustrating 😂

src/utils/network/session-core.js
```diff
@@ -44,17 +41,23 @@ import {
...
+export const setDocAclPermission = async (
+  session,
+  docType,
+  permissions,
+  podUrl,
+  webIdToAssignPermissionsTo
+) => {
+  const containerUrl = `${podUrl}PASS/Documents/`;
+  const documentUrl = `${containerUrl}${docType.replace("'", '').replace(' ', '_')}/`;
+  const webId = webIdToAssignPermissionsTo;
```
leekahung: If we're going to use webIds for the parameters to `setDocAclPermission` and `setDocContainerAclPermission`, we should just pass it into `setDocAclForUser` directly without needing to use `const webId`.

src/components/Form/SetAclPermissionForm.jsx
xscottxbrownx: For now the input is asking user to type in the podURL, then that form component converts it to a webID - and sends that as param/argument.

When we switch over to having a dropdown/selection menu we can update however we wish.

src/components/Form/SetAclPermsDocContainerForm.jsx
```diff
@@ -48,50 +49,21 @@ const SetAclPermsDocContainerForm = () => {
...
-    }
-
-    if (permissions === undefined) {
-      runNotification('Set permissions failed. Reason: Permissions not set.', 5, state, dispatch);
-      setTimeout(() => {
-        clearInputFields();
-      }, 3000);
-      return;
-    }
+    const webIdToSetPermsTo = `${permissionState.podUrlToSetPermissionsTo}/profile/card#me`;
```
leekahung: At the moment, the pod URLs stored/saved in PASS includes a trailing `/`. Users could slip through our validation by mistake and revoke their own permissions. Also, the extra `/` would make the incorrect webId if users type their pod URLs with a trailing `/`.

https://github.com/codeforpdx/PASS/assets/14917816/3510f0c6-0164-4490-a169-1a3da1741b90

src/components/Form/SetAclPermissionForm.jsx
```diff
@@ -50,56 +52,21 @@ const SetAclPermissionForm = () => {
...
-    }
-
-    if (!docType) {
-      runNotification('Search failed. Reason: No document type selected.', 5, state, dispatch);
-      setTimeout(() => {
-        dispatch({ type: 'CLEAR_PROCESSING' });
-      }, 3000);
-      return;
-    }
+    const webIdToSetPermsTo = `${permissionState.podUrlToSetPermissionsTo}/profile/card#me`;
```
leekahung: See comment in `SetAclPermsDocContainerForm` related to the trailing `/`

src/components/Form/SetAclPermsDocContainerForm.jsx
xscottxbrownx: So I now understand/see that there is a `/` at the end of the stored podUrl...
and pushed up fix so it creates correct webID without an extra.

src/components/Form/SetAclPermissionForm.jsx
```diff
@@ -50,56 +52,21 @@ const SetAclPermissionForm = () => {
...
+        `${permissions.read ? 'Give' : 'Revoke'} permission to ${
+          permissionState.podUrlToSetPermissionsTo
+        } for ${docType}.`,
         5,
         state,
         dispatch
       );
     } catch (error) {
-      runNotification('Set permissions failed. Reason: File not found.', 5, state, dispatch);
+      runNotification('FAILED TO SET PERMISSIONS. REASON: File not found.', 5, state, dispatch);
```
leekahung: A bit curious about the reason behind the change in case?

src/components/Form/SetAclPermissionForm.jsx
```diff
@@ -109,51 +76,80 @@ const SetAclPermissionForm = () => {
...
+            mb={2}
+            sx={{
+              width: '250px',
+              whiteSpace: 'nowrap',
+              overflow: 'hidden',
+              textOverflow: 'ellipsis'
+            }}
+          >
+            <span style={{ color: 'grey' }}>File Name: </span>
+            LongAssFileNameToTestIfThisIsWorkingCorrectly
```
leekahung: Think we can remove this test line if the truncation is working properly.

src/components/Form/SetAclPermissionForm.jsx
```diff
@@ -109,51 +76,80 @@ const SetAclPermissionForm = () => {
...
+          <FormControl fullWidth sx={{ marginTop: '2rem' }}>
+            <Button
+              variant="contained"
+              disabled={permissionState.podUrlToSetPermissionsTo === podUrl || state.processing}
+              type="submit"
+              color="primary"
+            >
+              {permissionState.permissionType
+                ? `${permissionState.permissionType} Permission`
+                : 'Give or Revoke Permission'}
```
leekahung: I think we could leave the original text as `Set Permission` and have the word "Set" changed into "Give" or "Revoke" rather than `Give or Revoke`.

src/components/Form/SetAclPermsDocContainerForm.jsx
```diff
@@ -48,50 +49,21 @@ const SetAclPermsDocContainerForm = () => {
...
+        `${permissions.read ? 'Give' : 'Revoke'} permission to ${
+          permissionState.podUrlToSetPermissionsTo
+        } for Documents Container.`,
         5,
         state,
         dispatch
       );
     } catch (error) {
-      runNotification('Set permissions failed. Reason: File not found.', 5, state, dispatch);
+      runNotification('FAILED TO SET PERMISSIONS. REASON: File not found.', 5, state, dispatch);
```
leekahung: Same comment as above for casing.

src/components/Form/SetAclPermsDocContainerForm.jsx
```diff
@@ -101,51 +73,55 @@ const SetAclPermsDocContainerForm = () => {
...
-                label="Revoke"
-                id="set-acl-perm-revoke"
-                name="setAclPerms"
-              />
-            </RadioGroup>
             <Button variant="contained" disabled={state.processing} type="submit" color="primary">
-              Set Permission
+              {permissionState.permissionType
+                ? `${permissionState.permissionType} Permission`
+                : 'Give or Revoke Permission'}
```
leekahung: Same comment as above for "Set" and "Give or Revoke"

src/components/Form/SetAclPermissionForm.jsx
xscottxbrownx: Just for matter of importance, like using different <h> tags...  I can switch it back if preferred. I'm not married to the idea.

src/components/Form/SetAclPermsDocContainerForm.jsx
xscottxbrownx: Just for matter of importance, like using different <h> tags...  I can switch it back if preferred. I'm not married to the idea.

src/components/Form/SetAclPermissionForm.jsx
xscottxbrownx: Don't want a placeholder with comment to know where to put it when we have the context?

src/components/Form/SetAclPermissionForm.jsx
leekahung: No preference for either, but was wondering if we could make the whole notification line uppercase if we're going to use this. Otherwise, we can revert this back.

src/components/Form/SetAclPermissionForm.jsx
leekahung: A TODO comment would be fine.

src/components/Form/SetAclPermsDocContainerForm.jsx
leekahung, line 1: The restrictions/guards that was set in place previously to prevent users from removing their own access to `Documents` seems to been removed.

https://github.com/codeforpdx/PASS/assets/14917816/0a6cd36e-fbc9-40f3-83dd-124a0f551ad5

I had to delete the container and regenerate it just to have access to it again.

leekahung: I thought MUI validation would prevents us from submitting afterwards, but it doesn't seem to be the case?

leekahung (approved):
This looks good to me. Just tested out the functionality as well. I'm okay with approving.

andycwilliams (approved):
Looks good to me

timbot1789 (approved):
looks good to me!

PR #354, opened 2023-07-31
Fixed template in AddClientModal

brancwill:
The "Add Client" button doesn't currently work because of a missing `/` in the `renderWebId` function. I've added the `/`, and now it works fine.

### DIFF
--- a/src/components/Modals/AddClientModal.jsx
+++ b/src/components/Modals/AddClientModal.jsx
@@ -18,8 +18,6 @@ import { runNotification } from '@utils';
 import { UserListContext } from '@contexts';
 // Model Imports
 import { createUser } from '../../model-helpers/User';
-// Constants Imports
-import { ENV } from '../../constants';
 // Component Imports
 import { FormSection } from '../Form';
 
@@ -32,8 +30,8 @@ import { FormSection } from '../Form';
  */
 
 const renderWebId = (username) => {
-  const oidcProvider = ENV.VITE_SOLID_IDENTITY_PROVIDER.split('//')[1];
-  return `https://${username}.${oidcProvider}/profile/card#me`;
+  const baseUrl = new URL(localStorage.getItem('oidcIssuer'));
+  return new URL(`${username}/profile/card#me`, baseUrl);
 };
 
 const AddClientModal = ({ showAddClientModal, setShowAddClientModal }) => {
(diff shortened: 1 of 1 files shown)


xscottxbrownx:
That's strange, the `oidcProvider` definitely has a `/` at the end.

Oh maybe that's just my machine's `Development` branch. 


I don't feel very strongly about this, but maybe remove `template` and just return:
`https://${username}.${oidcProvider}/profile/card#me`
??

brancwill:
@xscottxbrownx I can do that

src/components/Modals/AddClientModal.jsx
```diff
@@ -33,7 +33,7 @@ import { FormSection } from '../Form';
 
 const renderWebId = (username) => {
   const oidcProvider = ENV.VITE_SOLID_IDENTITY_PROVIDER.split('//')[1];
-  const template = ['https://', `.${oidcProvider}profile/card#me`];
+  const template = ['https://', `.${oidcProvider}/profile/card#me`];
```
timbot1789: We can actually make this fix a little better if we change both lines 35 and 36

We shouldn't be loading the oidcProvider from the ENV. We store the oidcProvider used during login in localstorage to fetch later. We can also construct these Urls using javascript's builtin `URL` object to avoid needing to deal with these dashes.

```suggestion
const renderWebId = (username) => {
  const baseUrl = new URL(localStorage.getItem('oidcProvider'));
  return new URL(`${username}/profile/card#me`, baseUrl);
}
```

This will technically return a webId in a different format than before, but it's the format that's used by the community solid server we're using as a dev server, so it shouldn't matter that much.

brancwill: @timbot1789 That changes the outcome url's structure. Is that ok?

timbot1789: Yeah that'll be fine. It's the format used by our dev server. And if you want to use a different format, you can manually edit the field

brancwill: Cool, I made the change

leekahung:
Hold off for the merger. Please wait for approval before merging. (see contributing guidelines for submitting a pull request: https://github.com/codeforpdx/PASS/blob/Development/docs/CONTRIBUTING.md)

leekahung:
@brancwill I've reverted the changes for now, but the branch used for this PR should be restored. You can start these fixes in another PR. Apologies for this.

PR #353, opened 2023-07-31
Version Update from v0.6.0 to v0.7.0

leekahung:
Version update here will be from v0.6.0 to v0.7.0 for the Master branch. Will create a new v0.7.0 Release after we got this merged into Master.

Major updates since last version includes:
- Switching from HashRouter to BrowserRouter (#281)
- Switching hosting service from GitHub Pages to Heroku (#281, #326)
- Converting Documents Page components into Profile Page components (#321)
- Moving Messages Route to Icon instead of a tab (#323)
- Styling updates to ClientList (#275), Message Pagination (#303), Inactivity Notification (#341)
- Documentation updates (#317)
- Script updates (#344)

Fixes includes:
- oidcIssuer in LocalStorage (#333)
- Client refresh and url redirection (#281)
- Login onKeyUp (#340)
- renderWebId in AddClientModal (#367)

PR #356, opened 2023-08-01
Cleaning up Development for merge to Master

leekahung:
This PR is related to suggested changes in PR #353.

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,11 +1,11 @@
 // React Imports
 import React from 'react';
 import { BrowserRouter } from 'react-router-dom';
-import { SessionProvider } from '@contexts';
 // Material UI Imports
 import { ThemeProvider } from '@mui/material/styles';
 import CssBaseline from '@mui/material/CssBaseline';
 // Context Imports
+import { SessionProvider } from '@contexts';
 import UserDataContextProvider from './contexts/UserDataContext';
 // Theme Imports
 import theme from './theme';
--- a/src/components/NavBar/OidcLoginComponent.jsx
+++ b/src/components/NavBar/OidcLoginComponent.jsx
@@ -3,7 +3,9 @@ import React, { useState } from 'react';
 // Custom Hook Imports
 import { useSession } from '@hooks';
 // Material UI Imports
-import { TextField, Box, Button } from '@mui/material';
+import Box from '@mui/material/Box';
+import Button from '@mui/material/Button';
+import TextField from '@mui/material/TextField';
 // Constants Imports
 import { ENV } from '../../constants';
 
--- a/src/components/NavBar/OidcLoginComponent.jsx
+++ b/src/components/NavBar/OidcLoginComponent.jsx
@@ -1,6 +1,6 @@
 // React Imports
 import React, { useState } from 'react';
-// Inrupt Library Imports
+// Custom Hook Imports
 import { useSession } from '@hooks';
 // Material UI Imports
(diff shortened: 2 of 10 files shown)


Jared-Krajewski (approved):

src/App.jsx
```diff
@@ -1,11 +1,11 @@
 // React Imports
 import React from 'react';
 import { BrowserRouter } from 'react-router-dom';
-import { SessionProvider } from '@contexts';
 // Material UI Imports
 import { ThemeProvider } from '@mui/material/styles';
 import CssBaseline from '@mui/material/CssBaseline';
 // Context Imports
+import { SessionProvider } from '@contexts';
 import UserDataContextProvider from './contexts/UserDataContext';
```
xscottxbrownx, line 9: might as well make this one `@contexts` too

src/components/NavBar/OidcLoginComponent.jsx
```diff
@@ -1,6 +1,6 @@
 // React Imports
 import React, { useState } from 'react';
-// Inrupt Library Imports
+// Custom Hook Imports
 import { useSession } from '@hooks';
 // Material UI Imports
 import { TextField, Box, Button } from '@mui/material';
```
xscottxbrownx: We've been importing MUI individually for performance - which we haven't actually tested how much of a performance hit importing it this way is, but the docs say the other way is better for performance.

src/components/Notification/InactivityMessage.jsx
```diff
@@ -1,18 +1,19 @@
...
 import Dialog from '@mui/material/Dialog';
 import DialogActions from '@mui/material/DialogActions';
 import DialogContent from '@mui/material/DialogContent';
 import DialogContentText from '@mui/material/DialogContentText';
 import DialogTitle from '@mui/material/DialogTitle';
-import CheckIcon from '@mui/icons-material/Check';
 import LogoutIcon from '@mui/icons-material/Logout';
-
+// Component Imports
 import LogoutButton from '../Modals/LogoutButton';
```
xscottxbrownx, line 15: import from the index?

xscottxbrownx (changes requested):

src/App.jsx
leekahung: `UserDataContext` is not used as an export from `contexts/index.js` to avoid dependency cycles discovered by the linter. It's likely complaining for importing from index.js from one of the sibling contexts when you could simply import the sibling directly.

Screen Shot 2023-08-01 at 15 32 47

But this lint error vanishes if we import the context explicitly as shown here.

src/components/NavBar/OidcLoginComponent.jsx
leekahung: Got this changed to individual imports.

src/components/Notification/InactivityMessage.jsx
leekahung: Similar dependency cycle warning as before with `UserDataContext`.

Screen Shot 2023-08-01 at 15 38 03

xscottxbrownx: Hmmm, so it seems these `index.js` file structures may not be really worth the trouble...  I'm also not sure why `LogoutButton` is in `Modals` folder either.

None of this is important, just thoughts.

xscottxbrownx (approved):
With new changes, looks good

Issue #77, opened 2023-04-06
Multi-file Upload


leekahung:
**Is your feature request related to a problem? Please describe.**
Current version of PASS supports single file uploads to Solid via HTML input element

**Describe the solution you'd like**
HTML input allows for multi-file uploads with the multiple attribute. We just need to update these elements and refactor all relevant components and functions

leekahung:
Opening new branch for #77, issue-77/multi-file-uploads; any contributors can use this branch to deal with that new feature

leekahung:
Function for multi-file upload is viable as proof-of-concept. Will keep as back log at this time.

leekahung:
Currently working on new system for uploading files on to Solid Pods. See PR #249.

leekahung:
Think what we've established in recent PR #249 helped us upload multiple files to Solid Pods one at a time and could assign metadata to them individually.

Think there might still be a case where someone wishes to upload a set of related files have them all use the same metadata description, but the concept would need to be flushed out first.

PR #357, opened 2023-08-01
Issue 335/device camera uploads

### LINKED ISSUE #335: [Open Discussion] Device camera uploads
**Feature Description**:
Leaving this as an open discussion for the idea about uploading documents/images from a device after capturing an image of it. It's been a feature that's been discussed with the PASS team and other collaborators that it could be a useful feature to have, but we have yet to implement it.

Currently, PASS allows for file uploads directly from the device's/browser's file explorer. But there could be times when users simply want to capture an image and upload that image into storage instead.

**Possible Solution**:
One way we could go about it is to create an option to upload images from device to PASS on "Upload Documents" and utilize the `getUserMedia` function from the `MediaDevices` web API.

Another way to go about this is to use input's native capture attribute from HTML to trigger the device's camera.

leekahung:
This PR implements image capture from mobile devices and refactors UploadDocumentModal with a new UploadButtonGroup sub component.

Had previously experimented with MediaDevice API, but the HTML input approach appears much simpler and are used for this purpose. Currently, only devices with screen widths smaller than 768px has this functionality.

Had tested this implementation on a fork to see if implementation works for mobile devices (see fork below).

https://github.com/codeforpdx/PASS/assets/14917816/ea934ddf-27d2-48bc-ba34-ac3cbc455b85

Note the styling was unaltered for the second button, I simply wanted to check if the implementation works. For styling, it looks like the following with the second button now utilizing Capture image for its wording:

Screen Shot 2023-08-01 at 0 13 59 (2)

The first button still opens up a file explorer.


xscottxbrownx:
I'll dig into the code in a little, but love the look and idea of this.  I also like that it still pulls up an "image selection" if you click the `capture image` on a larger device - BUT it would be real ideal if it could use the laptop camera (if one existed.)

src/components/Modals/UploadButtonGroup.jsx
```diff
@@ -0,0 +1,50 @@
...
+
+const UploadButtonGroup = ({ file, setFile }) => (
+  <Box sx={{ display: 'flex', padding: '8px 8px 0', boxSizing: 'border-box' }}>
+    <Button
+      variant={file ? 'outlined' : 'contained'}
+      component="label"
+      color="primary"
+      id="upload-doctype"
+      name="uploadDoctype"
+      onChange={(e) => setFile(e.target.files[0])}
```
xscottxbrownx, line 17: Can you explain this line? I'm missing something here... can we upload multiple files?  Trying to understand the `e.target.files[0]`

leekahung: `e.target.files` is an object that could store multiple files objects (you would have to turn on the multiple attribute in `<input>`), `e.target.files[0]` is simply the file object stored at the key 0. For individual uploads, we could access the file we want with `e.target.files[0]`.

Screen Shot 2023-08-01 at 15 14 17

leekahung:
> I'll dig into the code in a little, but love the look and idea of this. I also like that it still pulls up an "image selection" if you click the `capture image` on a larger device - BUT it would be real ideal if it could use the laptop camera (if one existed.)

Yeah, I did try out the idea with `<input>` and `capture` but it doesn't seem to work on desktop. It always seem to end up opening the file explorer.

The only way I manage to make the desktop camera working is if I were to use the `MediaDevice` API and `getUserMedia`. But at that point, it would seem like we would need to create a local GUI for the camera on PASS.

src/components/Modals/UploadButtonGroup.jsx
xscottxbrownx: Ok. Wouldn’t we always want to upload every file that the user selects to upload?

leekahung: We could (and from there we could fully utilize the `e.target.file` object), but we don't have the `multiple` attribute turned on at the moment (see Issue #77). And without the `multiple` attribute, users cannot select more than one file from the file explorer.

Clip without `multiple` attribute for `<input>`

https://github.com/codeforpdx/PASS/assets/14917816/a71a7314-58c5-44eb-af7e-a0041dc26f2c

Clip with `multiple` attribute for `<input>`

https://github.com/codeforpdx/PASS/assets/14917816/ff6211d5-f952-408d-a288-4d554d162587

On top of that, we haven't really discussed a way to support bulk upload yet with PASS and Solid. Like "how would we handle the metadata? Should it be just for the bulk upload itself? Or we want meta data for each file in the bulk?" and "how would we want to upload the files as a bulk upload (as a container of files within `Documents`? Have PASS doe multiple single file uploads onto `Documents`?", etc.

timbot1789 (approved):
This looks good. I think as a followup we can add a field that lets  users edit file names

src/components/Modals/UploadDocumentModal.jsx
```diff
@@ -189,6 +172,7 @@ const UploadDocumentModal = ({ showModal, setShowModal }) => {
                 onClick={clearInputFields}
                 fullWidth
                 helpertext="Please enter your name"
```
xscottxbrownx: Not sure where this `helpertext` on the CANCEL button came from. I don't think it's needed.

xscottxbrownx (changes requested):
Check into my comment please

src/components/Modals/UploadDocumentModal.jsx
leekahung: Good catch! Yeah, I don't think it's needed for this button. I'll remove it.

xscottxbrownx (approved):
Changes made, approved!

PR #360, opened 2023-08-01
Issue 329 update pass color scheme

russfraze:
This branch now has both color schemes in the theme. To switch from one to the other, just change the name of the export in theme.js to greenTheme or purpleTheme and uncomment the theme if commented out. For reference I'm including a screen grab of the purple theme. 

Should we send a screen grab to the UX team? 
Screen Shot 2023-08-01 at 2 52 21 PM

### DIFF
--- a/src/theme.js
+++ b/src/theme.js
@@ -7,7 +7,7 @@ const theme = createTheme({
       light: '#039686',
       main: '#6750a4',
       dark: '#004d3e',
-      slight: '#8fc2bb',
+      slight: '#eaddff',
       contrastText: '#fff'
     },
     secondary: {
--- a/src/theme.js
+++ b/src/theme.js
@@ -32,36 +32,36 @@ const greenTheme = createTheme({
   }
 })
 
-const purpleTheme = createTheme({
-  typography: { fontFamily: 'Roboto, sans-serif' },
-  palette: {
-    primary: {
-      light: '#039686',
-      main: '#6750a4',
-      dark: '#004d3e',
-      slight: '#eaddff',
-      contrastText: '#fff'
-    },
-    secondary: {
-      light: '#b32126',
-      main: '#625b71',
-      dark: '#790111',
-      contrastText: '#fff'
-    },
-    tertiary: {
-      light: '#dbc584',
-      main: '#7d5260',
-      dark: '#dbb032',
-      contrastText: '#fff'
-    },
-    status: {
-      danger: '#b3261e'
-    },
-    neutral: {
-      main: '#64748B',
-      contrastText: '#fff'
-    }
-  }
-});
+// const purpleTheme = createTheme({
+//   typography: { fontFamily: 'Roboto, sans-serif' },
+//   palette: {
+//     primary: {
+//       light: '#039686',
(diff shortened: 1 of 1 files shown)


leekahung:
Yeah, I think we should give the UX a look over the color scheme for some feedback.

timbot1789 (approved):

leekahung (changes requested):
A bit confused about the imports, but shouldn't `App.jsx` be importing "greenTheme" and not "theme" since the export name got changed? It'll still be `theme` when using the `useTheme` hook via `ThemeProvider`, but it should be `greenTheme` in `App.jsx`.

timbot1789:
> A bit confused about the imports, but shouldn't `App.jsx` be importing "greenTheme" and not "theme" since the export name got changed? It'll still be `theme` when using the `useTheme` hook via `ThemeProvider`, but it should be `greenTheme` in `App.jsx`.

It's a `default` export, so this will work fine. You can actually assign any name to a default import that you want. Internally,
```
import module from "./file";
```
is mostly syntactic sugar for
```
import default as module from "./file"
```

timbot1789 (commented):
https://www.figma.com/file/BfsGjGbFUVvtqNq18sykoY/Pass-Project?type=design&node-id=1216-2986&mode=design

It looks like UX has added an updated purple color scheme to figma. Could you take a look at that and update the PR?

https://www.figma.com/file/BfsGjGbFUVvtqNq18sykoY/Pass-Project?type=design&node-id=1216-2986&mode=design

russfraze:
@timbot1789 I'm looking in the Figma file style guide and the panel labeled Color Guidance. The light theme seems to be what was already there so I am thinking the pallet name dark theme is the new one? 

I think I need some guidance on how these colors are supposed to match up to the values we have in place in the theme. In each category, we have keys named light, main, dark, contrastText. In these themes there is an unlabeled top row, a on primary, primary container, and on primary container.  

I need to know which rows in these color palettes match to the key we have in place. 
Screen Shot 2023-08-03 at 1 24 30 PM

andycwilliams:
> @timbot1789 I'm looking in the Figma file style guide and the panel labeled Color Guidance. The light theme seems to be what was already there so I am thinking the pallet name dark theme is the new one?
> 
> I think I need some guidance on how these colors are supposed to match up to the values we have in place in the theme. In each category, we have keys named light, main, dark, contrastText. In these themes there is an unlabeled top row, a on primary, primary container, and on primary container.
> 
> I need to know which rows in these color palettes match to the key we have in place. Screen Shot 2023-08-03 at 1 24 30 PM

It would be best to check in with the UX team directly. Can you access their thread on Discord?

They also just posted some new palette ideas to In Progress page, which are easier to make sense of. It seems like they haven't fully settled on one yet.

And like @timbot1789 suggested, we could add a developer-only option to quickly swap between themes and see what works and what doesn't. In fact, that may be best since we can then show UX what they actually look like in the current state of the app.

leekahung:
In discussion with @xscottxbrownx, going to close existing PR for color scheme until further consolidation of ideas from Figma and the UX team.

Issue #351, opened 2023-07-29
Add relationship information to contacts


timbot1789:
**Feature Description**:
Currently, we are able to store contacts, but we want to be able to distinguish between clients/caseworkers/coworkers/ etc.

**Proposed Solution**:

Add a couple more fields to each contacts entry that contains information on the relationship.

```js
{
  person: <webId>, //required: the webId of the connected person
  pod: <podUrl>, // optional: the person's preferred pod url. If not present, we fetch the first pod url on the profile document. If that's not present, we derive it from the webid
  relationship: <client / case manager / associated organization>, // optional: an enum value describing how the user is related to this person
  relationship_status: <active / archived / etc>, // optional: an enum describing if the user is actively working with the person, or if this is a dormant or past relationship
}
```

ofu997:
I'd be interested in this

PR #362, opened 2023-08-02
Rewrite clients with react query

timbot1789:
This is one part of a rewrite of the "clients" feature to support having a similar feature for clients: https://github.com/codeforpdx/PASS/issues/351. This PR implements 3 main features:

1. Renames the 'client' feature to 'contacts' within the codebase. This will let the feature better reflect the design proposed in the issue -- have one universal page that can store all types of contacts for all types of users (e.g. clients, case managers, coworkers, organizations)
2. Rewrites the 'contacts' page to use react query instead of contexts. This is to serve as a demo of how it can work in PASS. It seems to work substantially better than our current system of contexts and helper functions. We've been able to remove a substantial amount of code from the system and push it into react query instead.
3. Added `podUrl` into the session context so we don't need to use `SignedInUserContext` as extensively.
4. Removed the `userlist.ttl` header thing, as it wasn't being used for anything


.eslintrc.js
```diff
@@ -38,7 +38,8 @@ module.exports = {
           ['@contexts', './src/contexts'],
           ['@pages', './src/pages'],
           ['@utils', './src/utils'],
-          ['@components', './src/components']
+          ['@components', './src/components'],
+          ['@constants', './src/constants']
```
timbot1789, line 42: I've been just adding these to the list as I encounter them. Not directly relevant to any changes.

timbot1789 (commented):
The change list is long, but a lot of it is do to renaming and deletions of unnecessary files. The most significant change I did was combining the `Clients` page and the `ClientsList` component into one `Contacts` page.

src/App.jsx
```diff
@@ -1,11 +1,12 @@
 // React Imports
 import React from 'react';
 import { BrowserRouter } from 'react-router-dom';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
```
timbot1789, line 4: We need to wrap the app in a provider for react query.

src/components/Contacts/ContactListTable.jsx
```diff
@@ -58,12 +53,11 @@ const ClientListTable = ({ setSelectedClientToDelete, setShowDeleteClientModal }
...
-            <ClientListTableRow
-              key={client.webId}
-              client={client}
-              setShowDeleteClientModal={setShowDeleteClientModal}
-              setSelectedClientToDelete={setSelectedClientToDelete}
+          {sortedContacts?.map((contact) => (
+            <ContactListTableRow
+              key={contact.webId}
+              contact={contact}
+              deleteContact={deleteContact}
```
timbot1789, line 66: Minor refactor: The client table doesn't need to worry about setting the modal. It just needs to tell its parent which client it wants to delete.

src/contexts/MessageContext.jsx
```diff
@@ -32,7 +31,7 @@ export const MessageContext = createContext([]);
 export const MessageContextProvider = ({ children }) => {
   const { podUrl } = useContext(SignedInUserContext);
   const [loadMessages, setLoadMessages] = useState(true);
-  const { session } = useSession();
+  const { session } = useContext(SessionContext);
```
timbot1789, line 34: This removes a dependency cycle between the hooks and contexts

src/components/Contacts/ContactListTable.jsx
leekahung, line 1: I'm not too sure what's happening but it seems like a new "Null" row is generated in the new `Contacts` list.

Clip of PR: 
Screen Shot 2023-08-02 at 1 34 29 (2)
Screen Shot 2023-08-02 at 1 37 33 (2)

Clip of Development:
Screen Shot 2023-08-02 at 1 35 30 (2)

src/components/Contacts/ContactListTableRow.jsx
```diff
@@ -14,48 +14,42 @@ import { DocumentListContext } from '@contexts';
...
-          to={`/profile/${encodeURIComponent(client.webId)}`}
-          state={{ client }}
+          to={`/profile/${encodeURIComponent(contact.webId)}`}
+          state={{ contact }}
           style={{ textDecoration: 'none', color: theme.palette.primary.dark }}
         >
-          <Button sx={{ textTransform: 'capitalize' }} onClick={() => setClient(client)}>
-            {client.person}
+          <Button sx={{ textTransform: 'capitalize' }} onClick={() => setContact(contact)}>
+            {`${contact.person}`}
```
leekahung: Minor nitpick (it's fine as it is), but think we could straight-up use the `String` function for this given we're just trying to force a string from this value.

src/contexts/DocumentListContext.jsx
```diff
@@ -31,7 +31,7 @@ export const DocumentListContextProvider = ({ children }) => {
   const [loadingDocuments, setLoadingDocuments] = useState(true);
   const { podUrl } = useContext(SignedInUserContext);
   const [client, setClient] = useState(null);
-  const { session } = useSession();
+  const { session } = useContext(SessionContext);
```
leekahung, line 34: I've noticed the switch with `SessionContext` for the `session` object. We can technically phase out the custom hook in favor `useContext(SessionContext)` in future changes, correct? Since `useSession` simply return everything that's already declared in `SessionContext`.

src/contexts/UserDataContext.jsx
```diff
@@ -1,30 +1,21 @@
...
  * @memberof contexts
  * @function UserDataContextProvider
- * @param {React.JSX.Element} children - The wrapped components that consumes
+ * @param {React.JSX.Element} children - The components that consume the data context
+ * @returns {React.JSX.Element} - A react app
  * Context from Provider
  */
-
 const UserDataContextProvider = ({ children }) => (
-  <SignedInUserContextProvider>
```
leekahung, line 21: I believe `SignedInUserContextProvider` is still in use? It's just `UserListContextProvider` being replaced.

src/hooks/useSession.js
leekahung, line 1: See earlier comment earlier about `useSession` and `SessionContext`. Aren't the object being return here the same if not similar to the one provided by `SessionContext`? Or there's a reason to utilize both in the codebase?

src/pages/Contacts.jsx
```diff
@@ -0,0 +1,160 @@
...
+  const [showDeleteContactModal, setShowDeleteContactModal] = useState(false);
+  const [selectedContactToDelete, setSelectedContactToDelete] = useState(null);
+  const queryClient = useQueryClient();
+
+  const handleSelectDeleteContact = (contact) => {
+    setSelectedContactToDelete(contact);
+    setShowDeleteContactModal(true);
+  };
+
+  const parseContacts = (data) => {
```
leekahung: I've noticed some of the utility functions are being moved from other files and into the pages and component file from the deleted `User.js` and `UserList.js` from `src/model-helpers`.

Should we not try to separate the two and mainly keep handler functions/hooks within components? Think if we leave the utility functions outside in a separate file like before, it'll be easier to isolate and unit test. We could simply remove the functions that's no longer in use like `updateUserActivity` and rename `UserList.js` to `ContactsList.js`.

src/contexts/UserDataContext.jsx
timbot1789: fixed

src/contexts/DocumentListContext.jsx
timbot1789: The main difference in this PR between `useContext(SessionContext)` and `useSession` is that `useSession` filters and modifies some of the data from `sessionContext`. E.G. `SessionContext` exposes the full solid profile document, which includes all the user's pod urls, however I modified `useSession` to filter that down to just one pod Url so that each component doesn't need to compute the pod url to use. 

I think that could be a useful distinction. Most components can access `useSession` just to get an authenticated fetch and a location to save and load, but components that need to manipulate the profile can use the `SessionContext` directly.

That's basically the distinction we've been using (somewhat unknowingly). I think if we find that distinction unnecessary, we can look at reworking SessionContext to better fit our needs.

src/components/Contacts/ContactListTableRow.jsx
timbot1789: Oh yeah this code is dumb. I'll fix it.

src/components/Contacts/ContactListTable.jsx
timbot1789: Oh that's probably the old header in the `userList.ttl`. I removed it from this PR because it wasn't being used for anything. I'll add a note about that to the PR description.

src/hooks/useSession.js
timbot1789: Commented here: https://github.com/codeforpdx/PASS/pull/362#discussion_r1282214435

src/pages/Contacts.jsx
timbot1789: Yeah the main reason I removed those is because React query took over a lot of the code that was being written in those files. React Query also uses a different paradigm of 'queries' and 'mutations' that I found difficult to map directly onto the 'model' paradigm I was originally going for.

Either way, you're right we can probably factor these out a bit more. However, I don't think it will look exactly like the old `userList` file

timbot1789: Moved these all into a custom hook

xscottxbrownx:
> The change list is long, but a lot of it is do to renaming and deletions of unnecessary files. The most significant change I did was combining the `Clients` page and the `ClientsList` component into one `Contacts` page.

I haven't dug into any of this yet, BUT:

I'm not sure we want a `Contacts` page for the caseworkers.  I think they need quick dedicated access to their Clients.  I'm unsure what other `contacts` caseworkers would have - that they'd have the authority to set perms to.

Like, let's say they have other caseworkers in their `contacts` list, do we see them trying to give them perms to a `clients` files? Would they have authority to do so?


But I haven't thought about this much honestly. I know we need to have changes to the `contactsList` so we can add caseworkers, organizations, etc - and not only have clients.  But, I'm unsure about changing any UI of the clients page due to this yet.

Could you explain a little about this thought process?

timbot1789:
@xscottxbrownx I think you're conflating two behaviors: permission setting and connection. We can handle permissions setting separately. None of the changes in this PR affect permissions, or limit how we can approach it.

Goal of this PR is to allow us to reuse code across multiple very similar features. A client will need to contact multiple case managers and possibly other clients (like family members). A case manager may need to talk with not just clients, but other related case managers and organizations. This ability to 'contact' people is functionality that we can reuse in several places. 

> But, I'm unsure about changing any UI of the clients page due to this yet. Could you explain a little about this thought process?

Thought process for updating the UI is because right now we don't have a way to actually differentiate between a client and a case worker, so I wanted to create a UI that would be equally usable by both. I think a more custom UI could be useful, but for now my thought is to just filter contacts by type, and make a table on the contacts page for each type. Then we can customize the tables to better suit each type. If we decide that certain contact types need different, more dedicated flows, then we can factor those tables out into separate flows.

Distinguishing between a client and a case worker is also non-trivial. What happens if we have a case worker on food stamps? They're a case worker for one organization, but a client for another.

src/components/Contacts/ContactListTable.jsx
```diff
@@ -1,36 +1,32 @@
...
- * ClientListTable Component - Component that generates table of clients from data within ClientList
+ * ContactListTable Component - Component that generates table of clients from data within ContactList
  *
- * @memberof Clients
- * @name ClientListTable
- * @param {clientListTableProps} Props - Props for ClientListTableRow
- * @returns {React.JSX.Element} The ClientListTableRow Component
+ * @memberof Contacts
+ * @name ContactListTable
+ * @param {clientListTableProps} Props - Props for ContactListTableRow
```
xscottxbrownx: param name `clientListTableProps` -> `contactListTableProps` ?

src/components/Modals/DeleteContactModal.jsx
```diff
@@ -12,70 +12,68 @@ import DialogContent from '@mui/material/DialogContent';
...
   const { state, dispatch } = useStatusNotification();
-  const { removeUser } = useContext(UserListContext);
 
   // Event handler for deleting client from client list
-  const handleDeleteClient = async (event) => {
+  const handleDeleteContact = async (event) => {
     event.preventDefault();
     runNotification(
-      `Deleting "${selectedClientToDelete?.person}" from client list...`,
+      `Deleting "${selectedContactToDelete?.person}" from client list...`,
```
xscottxbrownx: from `contact` list ?

src/components/Modals/DeleteContactModal.jsx
```diff
@@ -12,70 +12,68 @@ import DialogContent from '@mui/material/DialogContent';
...
     );
     try {
-      await removeUser(selectedClientToDelete);
-    } catch (e) {
-      runNotification(`Client deletion failed. Reason: ${e.message}`);
-    } finally {
+      await deleteContact(selectedContactToDelete);
       runNotification(
-        `"${selectedClientToDelete?.person}" deleted from client list...`,
+        `"${selectedContactToDelete?.givenName}" deleted from client list...`,
```
xscottxbrownx: 1) from `contact` list ?
2) only first name? What if they have multiple contacts with first name? Would going back to full name give the user more confidence that only what they wanted to delete was deleted?

src/components/Modals/DeleteContactModal.jsx
```diff
@@ -12,70 +12,68 @@ import DialogContent from '@mui/material/DialogContent';
...
         5,
         state,
         dispatch
       );
+    } catch (e) {
+      runNotification(`Client deletion falied. Reason: ${e.message}`);
```
xscottxbrownx: `Contact` deletion failed ?

src/components/Modals/DeleteContactModal.jsx
```diff
@@ -12,70 +12,68 @@ import DialogContent from '@mui/material/DialogContent';
...
         state={state}
         statusType="Status"
         defaultMessage="To be deleted..."
       >
-        <form onSubmit={handleDeleteClient} autoComplete="off">
+        <form onSubmit={handleDeleteContact} autoComplete="off">
           <DialogContent>
             <DialogContentText id="dialog-description">
-              {`Are you sure you want to delete "${selectedClientToDelete?.person}" from your client list?`}
+              {`Are you sure you want to delete "${selectedContactToDelete?.person}" from your client list?`}
```
xscottxbrownx: from your `contact` list ?

src/hooks/index.js
```diff
@@ -1,11 +1,12 @@
 import useField from './useField';
```
xscottxbrownx, line 2: I'd prefer the imports and exports be alphabetical, so it's easier to find what you need to adjust/delete/replace/etc - and so it's consistent with all the other files.  Even though, it doesn't matter one bit :)

src/pages/Contacts.jsx
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Contacts Component - Component that generates Contacts Page for PASS
+ *
+ * @memberof Pages
+ * @name Contacts
+ */
+
+const Contacts = () => {
+  localStorage.setItem('restorePath', '/contacts');
+  // state for AddContactModal component
```
xscottxbrownx: Not sure we need the comment for this state, if we aren't going to be consistent and label the other states as well. I think the name of the state is self-documenting enough honestly.

src/pages/Contacts.jsx
```diff
@@ -0,0 +1,85 @@
...
+          <ContactListTable
+            contacts={data}
+            deleteContact={(contact) => handleSelectDeleteContact(contact)}
+          />
+        ) : (
+          <EmptyListNotification type="Contacts" />
+        )}
+      </Box>
+
+      {/* modal/popup renders when showAddContactModal state is true */}
```
xscottxbrownx: this comment too. either consistent with other modal, or no longer necessary.

xscottxbrownx (changes requested):
Would also like another set of eyes double-checking the `contexts` and `hooks` files.

src/components/Contacts/ContactListTable.jsx
timbot1789: fixed

src/components/Modals/DeleteContactModal.jsx
timbot1789: fixed

src/components/Modals/DeleteContactModal.jsx
timbot1789: fixed

src/components/Modals/DeleteContactModal.jsx
timbot1789: fixed

src/hooks/index.js
timbot1789: fixed

src/pages/Contacts.jsx
timbot1789: fixed

src/pages/Contacts.jsx
timbot1789: fixed

src/components/Modals/DeleteContactModal.jsx
timbot1789: Fixed

timbot1789:
@xscottxbrownx @leekahung Thanks for the eyes on this! I think it's ready for another pass.

leekahung:
Hey @timbot1789 had started reviewing the PR again. Before diving into the codebase itself, I've started with a visual check with `npm run dev` for the new Contacts page. However, it looks like there could still be some issue with the contacts list as seen in the error from the clip below. The page crashes every time it enters it.

https://github.com/codeforpdx/PASS/assets/14917816/9cbef5cd-13eb-47a5-a278-b0da9d366c38

Other parts of the site remains unaffected, but again, it'll break once it re-enters the contacts page.

https://github.com/codeforpdx/PASS/assets/14917816/b70682e9-9bdc-49b2-b758-38133d91e6a4

timbot1789:
@leekahung delete your existing userlist.ttl and try testing again. Let me know if the issue persists

leekahung:
> @leekahung delete your existing userlist.ttl and try testing again. Let me know if the issue persists

That fixes it.

src/components/Contacts/ContactListTableRow.jsx
```diff
@@ -14,48 +14,42 @@ import { DocumentListContext } from '@contexts';
 import { StyledTableCell, StyledTableRow } from '../Table/TableStyles';
 
 /**
- * @typedef {import("../../typedefs.js").clientListTableRowProps} clientListTableRowProps
+ * @typedef {import("../../typedefs.js").contactListTableRowProps} contactListTableRowProps
```
leekahung: We could change the typedef in `typedefs.js` as well for `clientListTableRowProps` to `contactListTableRowProps` and their relevant properties.

src/components/Modals/AddContactModal.jsx
```diff
@@ -14,45 +14,35 @@ import InputAdornment from '@mui/material/InputAdornment';
...
-// Context Imports
-import { UserListContext } from '@contexts';
-// Model Imports
-import { createUser } from '../../model-helpers/User';
 // Component Imports
 import { FormSection } from '../Form';
 
 /**
- * AddClientModal Component - Component that allows users to add other user's
+ * AddContactModal Component - Component that allows users to add other user's
```
leekahung: Just noticed this now, but I think we could move the JSDoc here to above `AddContactModal` instead of `renderWebId`. In this order, JSDoc mistaken `renderWebId` for `AddContactModal`. Would probably need to include typedef for the React props passed into `AddContactModal`

Screen Shot 2023-08-10 at 16 11 50

Could add a separate JSDoc comment for `renderWebId`.

src/hooks/useContactsList.js
```diff
@@ -0,0 +1,115 @@
...
+    contact.givenName = getStringNoLocale(thing, RDF_PREDICATES.givenName);
+    contact.familyName = getStringNoLocale(thing, RDF_PREDICATES.familyName);
+    contact.person = getStringNoLocale(thing, RDF_PREDICATES.Person);
+    contacts.push(contact);
+  });
+  return contacts;
+};
+
+/**
+ *
```
leekahung, line 56: Could always flush out the JSDoc out later, hook seems to be working as intended.

leekahung (commented):
I like the changes, just a few minor comments here and there, think it should be good to go soon for my part. I look forward to working with React Query in the project.

src/components/Contacts/ContactListTable.jsx
```diff
@@ -1,36 +1,32 @@
...
+import ContactListTableRow from './ContactListTableRow';
 import { StyledTableCell } from '../Table/TableStyles';
 
 // ===== MAKE CHANGES HERE FOR TABLE HEADER / COLUMN TITLES =====
-const columnTitlesArray = ['Client', 'Pin', 'Delete'];
+const columnTitlesArray = ['Contact', 'Pin', 'Delete'];
 
 /**
- * @typedef {import("../../typedefs.js").clientListTableProps} clientListTableProps
+ * @typedef {import("../../typedefs.js").contactListTableProps} contactListTableProps
```
leekahung: The typedef in `typedefs.js` would need to be changed as well along with the new props for `ContactListTable`

src/components/Modals/DeleteContactModal.jsx
```diff
@@ -12,70 +12,68 @@ import DialogContent from '@mui/material/DialogContent';
...
 /**
- * DeleteClientModal Component - Component that allows users to delete other user's
+ * DeleteContactModal Component - Component that allows users to delete other user's
  * Pod URLs from a user's list stored on their own Pod
  *
- * @memberof Clients
- * @name DeleteClientModal
+ * @memberof Contacts
+ * @name DeleteContactModal
  */
```
leekahung: Think we could remove the space between the JSDoc and `DeleteContactModal` so that the JSDoc could associate with the component below it. Would probably need to include `@returns` and `@params` for the component and React props, respectively.

src/pages/Contacts.jsx
```diff
@@ -0,0 +1,83 @@
...
+import { ContactListTable } from '@components/Contacts';
+import { LoadingAnimation, EmptyListNotification } from '@components/Notification';
+
+/**
+ * Contacts Component - Component that generates Contacts Page for PASS
+ *
+ * @memberof Pages
+ * @name Contacts
+ */
+
```
leekahung: Could remove the spacing between the JSDoc and `Contacts` page. Would need to include a `@returns` for JSDoc.

src/components/Contacts/ContactListTable.jsx
timbot1789: Is it possible to move these into a file closer to the code, or enable automated validation of these? Devs will not be able to keep track of these types.

timbot1789: Either way, I'll update things for this PR.

src/components/Modals/DeleteContactModal.jsx
timbot1789: fixed

src/pages/Contacts.jsx
timbot1789: fixed

src/components/Contacts/ContactListTable.jsx
timbot1789: moved the typedef into the ContactListTable and updated it.

src/components/Contacts/ContactListTableRow.jsx
timbot1789: fixed

src/components/Modals/AddContactModal.jsx
timbot1789: fixed

src/hooks/useContactsList.js
timbot1789: fixed

src/components/Contacts/ContactListTable.jsx
timbot1789: Would still like to see a followup to enable automated validation of JSDoc. I don't see much value in using JSDoc if we don't have some automated requirements for adding it.

leekahung: > Is it possible to move these into a file closer to the code, or enable automated validation of these? Devs will not be able to keep track of these types.

Yes, we could move the typedefs just above the type we're defining instead of importing it from one location.

As for keeping track of these types, I believe there's an eslint plugin we could use for JSDoc called `eslint-plugin-jsdoc`.

leekahung: There's a way to make it do typescript like validation I believe(?) (see screenshot below from eslint-plugin-jsdoc documentation), but I have not yet messed with the setting outside of the default ones currently set up for the project:

Screen Shot 2023-08-11 at 12 28 10 (2)

timbot1789: Cool. We can look into implementing this in a separate issue.

src/components/Modals/AddContactModal.jsx
```diff
@@ -12,49 +12,43 @@ import FormControl from '@mui/material/FormControl';
...
 };
 
-const AddClientModal = ({ showAddClientModal, setShowAddClientModal }) => {
+/**
+ * AddContactModal Component - Component that allows users to add other user's
+ * Pod URLs from a user's list stored on their own Pod
+ *
+ * @memberof Contacts
+ * @name AddContactModal
+ * @param {Function} addContact  - function to add a contact
```
leekahung: The param for the component should be a React Prop. I believe you've implemented them correctly in the components in the new `src/components/Contacts`

src/components/Modals/DeleteContactModal.jsx
```diff
@@ -12,70 +12,77 @@ import DialogContent from '@mui/material/DialogContent';
...
+
+/**
+ * DeleteContactModal Component - Component that allows users to delete other user's
  * Pod URLs from a user's list stored on their own Pod
  *
- * @memberof Modals
- * @name DeleteClientModal
+ * @memberof Contacts
+ * @name DeleteContactModal
+ * @param {DeleteContactModalParams} root0 - params
```
leekahung: Instead of using `root0`, think you could use `Props`.

leekahung (commented):
Outside of those few comments with the JSDoc, I think the rest of the files are fine. Think it should be good to go on my end. Not sure for @xscottxbrownx.

src/components/Modals/DeleteContactModal.jsx
```diff
@@ -12,70 +12,77 @@ import DialogContent from '@mui/material/DialogContent';
...
+ * @property {Function} deleteContact - method that deletes contact
+ */
+
+/**
+ * DeleteContactModal Component - Component that allows users to delete other user's
  * Pod URLs from a user's list stored on their own Pod
  *
- * @memberof Modals
- * @name DeleteClientModal
+ * @memberof Contacts
```
xscottxbrownx: Is it not a member of Modals?

src/components/Modals/index.js
```diff
@@ -10,4 +10,4 @@ import UploadDocumentModal from './UploadDocumentModal';
  * @namespace Modals
  */
 
-export { AddClientModal, DeleteClientModal, LogoutModal, NewMessageModal, UploadDocumentModal };
+export { AddContactModal, NewMessageModal, DeleteContactModal, LogoutModal, UploadDocumentModal };
```
xscottxbrownx: I'd prefer alphabetical order, but not an issue.  Just find it easier for future maintainability/changes.

xscottxbrownx (approved):
No issues for me that have any significance

src/components/Modals/AddContactModal.jsx
timbot1789: fixed

src/components/Modals/DeleteContactModal.jsx
timbot1789: fixed

src/components/Modals/DeleteContactModal.jsx
timbot1789: fixed

src/components/Modals/index.js
timbot1789: fixed

leekahung (approved):
Think this should be good. Approving.

PR #370, opened 2023-08-08
Issue 358/render permissions forms

### LINKED ISSUE #358: Render permissions forms on profile page as modals after an onClick event only
Both "set permissions" forms are rendered on the Profile Page currently.

---
Ideally, these would be modals/dialogs that popup after an onClick event attached to a button or icon.

---
### Rationale:
**1.** To clean up the profile page and bring the focus to the user's info and documents only.
**2.** To have the "set permission to a specific file" be more obviously linked with a specific file - modal renders after an onClick of something in that file's row on the documents table.
**3.** To be more consistent with rest of app ("Add Client", "Add Document", etc)
**4.** Second step in getting these forms into the final version we actually want them to be.

---
Follow similar structure of the `AddClientModal` or `UploadDocumentModal` files - using what we've already built into the "set permissions forms" files.

brancwill:
---
name: Pull Request
about: Create a pull request to merge your code into the PASS codebase
title: ''358/render-permission-forms"
labels: ''Enhancement"
assignees: ''brancwill"

---

I changed the `SetAclPermissionForm` and `SetAclPermsDocContainerForm` into modals. The former is now opened by a button that I've attached to each file in the Document Table to open the modal with the `File Name` set to said file's name. The latter is opened by a button next to the `Add Document` button.

---
## This PR:
**1.**    

---
### Components
`SetAclPermissionsModal.jsx` (formerly `SetAclPermissionsForm.jsx`)
`SetAclPermsDocContainerModal.jsx` (formerly `SetAclPermsDocContainerModal.jsx`)
`DocumentTable.jsx`
`DocumentTableRow.jsx`
`Profile.jsx`
`index.js` (in Forms and in Modals folder)

### Tests
List file names here

### Other Files
List file names here

---
Screenshot1
Screenshot2

---

---
@xscottxbrownx and I spoke about potentially condensing the table through either a menu containing the document actions (setting permissions, deletion, etc.) or through an accordion (@leekahung 's suggestion). If this is something we'd like to do, I can either add to this PR, or start another one.

---
**1.**  I'm pretty unfamiliar with MUI and JSDoc, so I copy and pasted a lot. With JSDoc in particular, I wasn't sure if I should add Modals to `SetAclPermissionsModal` and `SetAclPermsDocContainerModal`'s `@memberof` or not. Feedback would be appreciated, I'm happy to tweak anything.


src/components/Documents/DocumentTable.jsx
```diff
@@ -15,15 +15,20 @@ import { StyledTableCell } from '../Table/TableStyles';
 import DocumentTableRow from './DocumentTableRow';
 import { EmptyListNotification, LoadingAnimation } from '../Notification';
 
+/**
+ * @typedef {import("../../typedefs.js").documentTableProps} documentTableProps
```
leekahung, line 19: You would want to create a typedef in `typedefs.js` for `documentTableProps` since we don't have one yet for this component.

To see how to create the typedef for the React prop, you could look at one of the prop definitions in `typedefs.js` (see `footerProps` in Line 287).

As you would see there, `@property` is a decorator that help define objects in JSDoc, namely for the object being described within the JSDoc.

src/components/Documents/DocumentTableRow.jsx
```diff
@@ -27,7 +28,7 @@ import DOC_TYPES from '../../constants/doc_types';
  * @param {documentTableRowProps} Props - Props for DocumentTableRow
  * @returns {React.JSX.Element} The DocumentTableRow component
  */
-const DocumentTableRow = ({ document }) => {
+const DocumentTableRow = ({ document, handlePermissions }) => {
```
leekahung: Would need to update the typedef for `documentTableRowProps` to include `handlePermissions`

src/components/Documents/DocumentTableRow.jsx
```diff
@@ -64,6 +65,11 @@ const DocumentTableRow = ({ document }) => {
           <FileOpenIcon />
         </IconButton>
       </StyledTableCell>
```
leekahung, line 67: Not too urgent of a fix in this PR, but was wondering when we'll be phasing out the styled-components here with MUI components.

src/components/Form/index.js
```diff
@@ -1,12 +1,12 @@
 import DocumentSelection from './DocumentSelection';
 import FormSection from './FormSection';
-import SetAclPermissionForm from './SetAclPermissionForm';
-import SetAclPermsDocContainerForm from './SetAclPermsDocContainerForm';
+import SetAclPermissionModal from '../Modals/SetAclPermissionModal';
```
leekahung: Since they're already inside `src/components/Modals` why not have them imported from the `index.js` file there?

src/components/Modals/SetAclPermissionModal.jsx
```diff
@@ -0,0 +1,200 @@
...
+import Select from '@mui/material/Select';
+import SettingsIcon from '@mui/icons-material/Settings';
+import TextField from '@mui/material/TextField';
+import Typography from '@mui/material/Typography';
+// Utility Imports
+import { runNotification, setDocAclPermission } from '@utils';
+// Context Imports
+import { SignedInUserContext } from '@contexts';
+// Component Imports
+import DocumentSelection from '../Form/DocumentSelection';
```
leekahung: The imports in Line 22 and 23 could be simplified with named imports `import { DocumentSelection, FormSection } from './Form'` given we're using the `index.js` file from `src/components/Form`

src/components/Modals/SetAclPermissionModal.jsx
```diff
@@ -0,0 +1,200 @@
...
+import FormSection from '../Form/FormSection';
+
+/**
+ * @typedef {import("../../typedefs.js").setAclPermissionModalProps} setAclPermissionModalProps
```
leekahung: Similar to the comment on the `DocumentTable`, we would need to create a new typedef for `setAclPermissionModalProps` for the component in `typedefs.js` for these imports. (see comment before for guidance)

src/components/Modals/SetAclPermissionModal.jsx
```diff
@@ -0,0 +1,200 @@
...
+  const handleDocType = (event) => {
+    setDocType(event.target.value);
+  };
+
+  const clearInputFields = () => {
+    dispatch({ type: 'CLEAR_PROCESSING' });
+  };
+
+  const handleCloseModal = () => {
+    setPermissionState({
```
leekahung: I think we can have `setPermissionState` and `setDocType` be moved to `clearInputFields` rather than a handler function for closing the modal.

src/components/Modals/SetAclPermsDocContainerModal.jsx
```diff
@@ -0,0 +1,172 @@
...
+      setTimeout(() => {
+        clearInputFields();
+      }, 3000);
+    }
+  };
+
+  return (
+    <Dialog open={showModal} onClose={handleCloseModal}>
+      <FormSection
+        title="Permission for Container"
```
leekahung: While it's correct that's it's permission for a container on Solid, think it'll be clearer if we specific what, like "Permissions to Documents"

src/pages/Profile.jsx
```diff
@@ -89,24 +106,43 @@ const Profile = () => {
...
+            <Container sx={{ display: 'flex', justifyContent: 'center', gap: 2 }}>
+              <Button
+                variant="contained"
+                color="primary"
+                size="small"
+                aria-label="Set Container Permissions Button"
+                startIcon={<SettingsIcon />}
+                onClick={() => setShowAclPermsDocContainerModal(true)}
+              >
+                Container Permissions
```
leekahung: Think we can rephrase this to use something else other than `Container Permissions`. Not really sure what at the moment though.

src/components/Modals/SetAclPermsDocContainerModal.jsx
```diff
@@ -0,0 +1,172 @@
...
+import MenuItem from '@mui/material/MenuItem';
+import Select from '@mui/material/Select';
+import SettingsIcon from '@mui/icons-material/Settings';
+import TextField from '@mui/material/TextField';
+// Utility Imports
+import { runNotification, setDocContainerAclPermission } from '@utils';
+// Context Imports
+import { SignedInUserContext } from '@contexts';
+// Component Imports
+import FormSection from '../Form/FormSection';
```
leekahung: Line could be changed to `import { FormSection } from './Form'`

src/components/Modals/SetAclPermsDocContainerModal.jsx
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * @typedef {import("../../typedefs.js").setAclPermsDocContainerModalProps} setAclPermsDocContainerModalProps
```
leekahung: Similar to the comment on the `DocumentTable`, we would need to create a new typedef for `setAclPermsDocContainerModalProps` for the component in `typedefs.js` for these imports. (see comment before for guidance)

src/components/Modals/SetAclPermsDocContainerModal.jsx
```diff
@@ -0,0 +1,172 @@
...
+    podUrlToSetPermissionsTo: '',
+    permissionType: ''
+  });
+
+  const clearInputFields = () => {
+    dispatch({ type: 'CLEAR_PROCESSING' });
+  };
+
+  const handleCloseModal = () => {
+    setPermissionState({
```
leekahung: Think we can move `setPermissionState` set function to `clearInputFields` instead of `handleCloseModal`.

src/pages/Profile.jsx
```diff
@@ -89,24 +106,43 @@ const Profile = () => {
...
+              </Button>
+              <Button
```
leekahung: While I understand moving the `Container Permissions` button into the conditional rendering only for the `Profile` page for the user, why was the `Add Document` buttons moved into here as well? Case workers should be able to help clients upload into their Pods if they have permission.

src/pages/Profile.jsx
```diff
@@ -89,24 +106,43 @@ const Profile = () => {
...
+                color="secondary"
+                size="small"
+                aria-label="Add Client Button"
+                startIcon={<AddIcon />}
+                onClick={() => setShowAddDocModal(true)}
+              >
+                Add Document
+              </Button>
+            </Container>
+            <SetAclPermsDocContainerModal
```
leekahung: Given that the set permission "forms" are now modals that depends on the `showModal` prop to render, I don't think they need to be kept inside this rendering condition. Think we can have these moved to where `UploadDocumentsModal` and `DocumentTable` are.

xscottxbrownx:
@Jared-Krajewski I don't see any difference between this PR template and the Issue templates, but you can see that the extra stuff used to help GitHub was displayed...

Screenshot 2023-08-08 at 7 17 53 AM

If you have time, could you take a look and see what may be wrong?

src/pages/Profile.jsx
```diff
@@ -33,14 +38,26 @@ const Profile = () => {
 
   // Documents related states
   const { session } = useSession();
-  const [showModal, setShowModal] = useState(false);
+  const [showAddDocModal, setShowAddDocModal] = useState(false);
+  const [showAclPermsDocContainerModal, setShowAclPermsDocContainerModal] = useState(false);
+  const [showAclPermissionModal, setShowAclPermissionModal] = useState(false);
+  const [documentName, setDocumentName] = useState('filename.jpg');
```
timbot1789: I wouldn't put a false filename in here as a default. It seems like something that could easily lead to errors.

src/components/Form/index.js
brancwill: OK, I'll get rid of them here and leave them in the Modal folder one. I'm not familiar with importing components into these different `index.js` files. Could you give me the proper term to search so I can learn more about why/how to do it?

src/pages/Profile.jsx
brancwill: All Document Permissions?

src/components/Modals/SetAclPermsDocContainerModal.jsx
brancwill: What's the difference? Not in a disrespectful way, I don't actually know what the difference is.

src/pages/Profile.jsx
brancwill: Sorry, I moved the whole container in there without thinking.

brancwill:
OK, I've made the changes

src/components/Documents/DocumentTableRow.jsx
timbot1789: I don't think we need to worry about this here. You can put up another issue about it. Would be a good first issue.

timbot1789 (commented):
I know I'm a little late to the party with my review, but I do have one question: why 2 separate modals? It seems they do largely the same thing. The `SetAclPermissionsModal` sets permissions on one dataset, and `SetAclPermsDocContainerModal` sets permissions on a different dataset. It seems to me like we could just have one modal, `SetAclPermissionsModal` that just takes in the requested dataset, displays current dataset permissions, and lets you edit them. The signature for both of them can be:

`const SetAclPermissionsModal = (showModal, setShowModal, dataset)`

Then all permissions settings are done directly on the `dataset` parameter.

src/typedefs.js
```diff
@@ -293,4 +306,26 @@ const React = require('react');
...
 
+/**
+ * setAclPermissionModalProps is an object that stores the props for the setAclPermissionModal component.
+ *
+ * @exports setAclPermissionModalProps
+ * @typedef {object} setAclPermissionModalProps
+ * @property {boolean} showModal - Boolean for showing setAclPermissionModal.
+ * @property {React.Dispatch<React.SetStateAction<boolean>>} setShowModal
+ * - React set function for setting showModal state
+ * @property {(string) => void} handlePermissions
```
leekahung: The last prop for `SetAclPermissionModal` seems to be a string(?) called `documentName` and not the handler function `handlePermissions`.

src/components/Documents/DocumentTableRow.jsx
leekahung: I believe's there's already a related issue set in Issue #219.

src/components/Modals/SetAclPermsDocContainerModal.jsx
leekahung: Think this is fine too. Wasn't expecting `setShowModal` to move from the handler function as well.

src/components/Modals/SetAclPermissionModal.jsx
leekahung: Was not expecting `setShowModal` to move as well, but guess this works.

src/pages/Profile.jsx
leekahung: Think that sounds alright.

src/components/Modals/SetAclPermsDocContainerModal.jsx
leekahung: One's a named import importing from `src/components/Form/index.js` the other is a default import importing from `src/components/Form/FormSection.jsx`.

Depending on whether or not there's going to be additional changes or modifications from components inside the "src/component/Form", it might be more useful to use the named import approach. Seems like you've already changed it to the named import. Which should be fine.

andycwilliams:
I definitely like having the forms no longer just awkwardly placed under the document table.

I would suggest not using the settings icon for this since we already have it for the settings page in the navbar menu.

Some ideas for alternatives:

- LockIcon (and then LockOpenIcon if the permission has been granted, maybe)
- ShieldIcon 
- PrivacyTipIcon 
- SecurityIcon from

https://mui.com/material-ui/material-icons/

I'm not sure if the `<Box>` on line 99 of `SetAclPermissionModal.jsx` actually does anything. Might be able to remove it. Same with line 95 of `SetAclPermsDocContainerModal.jsx`.

Also, now that all of the forms are in modals and no longer directly on a webpage, we can maybe remove `boxShadow: 2` on line 34 of `FormSection.jsx`. With this refactoring, it just creates an extra unnecessary line around the border for all modals.

src/pages/Profile.jsx
brancwill: Actually, what about "Set Permission for All Documents"?

brancwill:
OK, I'll get all of these new changes incorporated soon. My weekend starts on Sunday, so I'll definitely have time then, if not sooner.

src/components/Modals/SetAclPermissionsModal.jsx
leekahung, line 1: While the individual permission modal looks alright, if the modal for all documents is going to be selected for all documents, I don't think we need to specify document type. The whole container is going to be set regardless of what was chosen.

Screen Shot 2023-08-16 at 15 15 54 (2)

src/components/Modals/SetAclPermissionsModal.jsx
```diff
@@ -0,0 +1,216 @@
...
+              ) : null}
+
+              <br />
+              <FormControl fullWidth sx={{ display: 'flex', gap: 2, flexDirection: 'column' }}>
+                <Button
+                  variant="contained"
+                  disabled={permissionState.podUrlToSetPermissionsTo === podUrl || state.processing}
+                  type="submit"
+                  color="primary"
+                  startIcon={<ShieldIcon />}
```
leekahung: While the shield is nice, I think the previous iteration with the gear looks better. It's certainly something I think about when it comes to a settings option or menu.

src/components/Documents/DocumentTableRow.jsx
```diff
@@ -66,8 +66,8 @@ const DocumentTableRow = ({ document, handlePermissions }) => {
...
       <StyledTableCell align="center">
-        <IconButton type="button" onClick={() => handlePermissions(name)}>
-          <SettingsIcon />
+        <IconButton type="button" onClick={() => handleModal('document', name)}>
+          <ShieldIcon />
```
leekahung, line 70: Similar to the case with the Shield icon in the set permission button, but I think the previous iteration with the gear icon looks better in this case.

src/pages/Profile.jsx
```diff
@@ -113,10 +115,10 @@ const Profile = () => {
               color="primary"
               size="small"
               aria-label="Set Container Permissions Button"
-              startIcon={<SettingsIcon />}
-              onClick={() => setShowAclPermsDocContainerModal(true)}
+              startIcon={<ShieldIcon />}
```
leekahung, line 118: Similar to above, think the gear icon from `SettingsIcon` looks better for this case.

src/typedefs.js
```diff
@@ -226,8 +226,8 @@ const React = require('react');
  *
  * @exports documentTableProps
  * @typedef {object} documentTableProps
- * @property {(string) => void} handlePermissions
- * - Function that sets the correct File Name for the SetAclPermissions Modal, and opens it.
+ * @property {(string) => void} handleModal
```
leekahung: I've just check the handler function for handleModal. It looks likes in two parameters, `type` and `fileName`. I believe `type` is also a string(?). If that's the case, the type for handleModal should be updated to be `(type: string, fileName: string) => void` similar to how it's like in TypeScript. See clip below:

https://github.com/codeforpdx/PASS/assets/14917816/9ea0310a-18b1-4b5f-a131-f21c86eacb4f

src/typedefs.js
```diff
@@ -238,7 +238,7 @@ const React = require('react');
  * @exports documentTableRowProps
  * @typedef {object} documentTableRowProps
  * @property {File} document - File object containing the document
- * @property {(string) => void} handlePermissions
+ * @property {(string) => void} handleModal
```
leekahung: Can the same type defined above for handleModal in this line.

src/components/Form/index.js
leekahung: I'm not too sure about the name for this, but basically `index.js` acts as a relay for existing exports in the existing directory and allows it to export them as named imports. It let's us import the path for the directory other than the path for the file.

src/pages/Profile.jsx
leekahung: That should be fine.

src/components/Modals/SetAclPermissionsModal.jsx
brancwill: Sorry, had the wrong condition for the render. It's fixed now

src/components/Modals/SetAclPermissionsModal.jsx
brancwill: I changed it because of what @andycwilliams said above. What should I do here?

leekahung:
Hey @brancwill, I've noticed that there's commits from other PRs mixed into this one. I'm thinking there could be a mix up when trying to resolve merge conflicts for this branch. Let me know if you need some help. I'll be available on Discord. 👍

brancwill:
There was some confusion with merge conflicts, so I'm made a new branch and I'm making another PR

PR #378, opened 2023-08-11
Create User Registration Page

### LINKED ISSUE #213: Make the "Add User" section also create a Web ID and pod for the new user

timbot1789:
## Create User Registration Page

---
## This PR:
1. Creates a new page, Signup, where users can create pods, and optionally connect themselves with a case manager
2. Adds a new ENV variable, `VITE_SOLID_POD_SERVER` which is the default server to add new pods to.
3. Grants each case manager a unique version of this page, which they can share with potential clients. Any client that registers through a case manager's page will automatically have that case manager added to their contact list.
4. Initializes the newly-created pod with a PASS container
5. If they connected to a case manager, registration will:
  - Give the case manager read-write access to the user's PASS container
  - Add the case manager to the user's contacts list

- Each PASS user now has a signup link available in their profile document
image
- Clicking on that will launch the signup page in a new tab
  
image
 This page will allow new users to sign up for PASS, and connect them with whoever is referenced in the `webId` query param in the page URL. If you remove the query param from the URL, you will be able to register for a pod without being connected to a case manager

- Once a new user fills out the top form, their pod will be created. Then they will be redirected to the pod provider and asked to log in:
image
image

- Once they log in, PASS will complete pod initialization, and the signup form will be replaced with a message (NOTE: this is a dev testing view. we should replace this in the future with a user information collection flow):
image

- In the background, PASS will initialize a PASS container, as well as a PASS contacts list. If the user registered with a case manager, PASS will grant the case manager read-write permissions to the pod, and will add the case manager to the user's contact list. (NOTE: This PR does not currently add the client to the case manager's contact list)
image
image
image

---
Currently, the registration form only works when connected to a Community Solid Server.

---
## Future Steps:
- Add a button to the landing page to allow a person to create a pod without registering for pass
- Add another form to allow users to use existing pods
- Send a notification to case managers that a person is requesting to register with them
- Replace the "You have successfully registered a pod" card with an information collection flow.


src/AppRoutes.jsx
```diff
@@ -28,6 +29,14 @@ const AppRoutes = () => {
         path="/"
         element={session.info.isLoggedIn ? <Navigate to={path} replace /> : <Home />}
       />
+      <Route
+        path="/signup"
+        element={
+          <SessionProvider>
```
timbot1789, line 35: As the new user signup can be launched from a case manager's session, we want to set up a separate session for the newly-created pod to act in. This child `SessionProvider` "shadows" the parent `SessionProvider`, hiding the case manager's session from this flow without logging the case manager out.

leekahung:
Would signing in a client with an existing pod be using a separate flow and be done in a different PR?

timbot1789:
> Would signing in a client with an existing pod be using a separate flow and be done in a different PR?

It will be in a separate PR, yes

src/pages/Signup.jsx
```diff
@@ -0,0 +1,300 @@
+// React Imports
+import React, { useState, useEffect } from 'react';
+import { useSearchParams } from 'react-router-dom';
+import { TextField, Button, CardHeader } from '@mui/material';
+
```
andycwilliams: We should probably replace empty lines with relevant labels such as `// Material UI Imports` and the same for hooks and contexts. Also arrange them the same way that we have with other files (e.g. group all MUI imports together and place everything in the same descending order)

Although now I'm noticing a similar case in `Profile.jsx` and perhaps elsewhere

src/pages/Signup.jsx
```diff
@@ -0,0 +1,300 @@
...
+import Container from '@mui/material/Container';
+import Paper from '@mui/material/Paper';
+import Typography from '@mui/material/Typography';
+import useContactsList from '@hooks/useContactsList';
+import useSession from '@hooks/useSession';
+
+const formRowStyle = {
+  margin: '20px 0'
+};
+
```
andycwilliams, line 41: I assume we'll focus on fully applying MUI styling in another PR, so I don't think it's essential for this one

src/pages/Signup.jsx
```diff
@@ -0,0 +1,300 @@
...
+        <form onSubmit={handleSubmit} style={formRowStyle} autoComplete="off">
+          <TextField
+            style={textFieldStyle}
+            id="email-form"
+            aria-label="Email"
+            label="Email"
+            variant="outlined"
+            value={email}
+            onChange={(e) => setEmail(e.target.value)}
+          />
```
andycwilliams, line 168: This field should have `required` added

src/pages/Signup.jsx
```diff
@@ -0,0 +1,300 @@
...
+
+        <form onSubmit={handleSubmit} style={formRowStyle} autoComplete="off">
+          <TextField
+            style={textFieldStyle}
+            id="email-form"
+            aria-label="Email"
+            label="Email"
+            variant="outlined"
+            value={email}
+            onChange={(e) => setEmail(e.target.value)}
```
andycwilliams, line 177: This field should have required added

src/pages/Signup.jsx
```diff
@@ -0,0 +1,300 @@
...
+          />
+          <br />
+          <TextField
+            style={textFieldStyle}
+            id="password-form"
+            aria-label="Password"
+            label="Password"
+            variant="outlined"
+            value={password}
+            onChange={(e) => setPassword(e.target.value)}
```
andycwilliams, line 189: Same as above

src/pages/Signup.jsx
```diff
@@ -0,0 +1,300 @@
...
+          />
+          <br />
+          <TextField
+            style={textFieldStyle}
+            id="confirm-password-form"
+            aria-label="Confirm Password"
+            label="Confirm Password"
+            variant="outlined"
+            value={confirmPassword}
+            onChange={(e) => setConfirmPassword(e.target.value)}
```
andycwilliams, line 214: Same as above

src/pages/Signup.jsx
timbot1789: Yeah I don't have any design from UX for this, so I just wanted to get something clean and readable

src/pages/Signup.jsx
timbot1789: fixed

src/pages/Signup.jsx
timbot1789: fixed

src/pages/Signup.jsx
timbot1789: fixed

src/pages/Signup.jsx
timbot1789: fixed

src/pages/Signup.jsx
```diff
@@ -0,0 +1,304 @@
...
+    password,
+    confirmPassword,
+    podName,
+    createWebId: true,
+    createPod: true,
+    rootPod: false,
+    register: true
+  };
+
+  const response = await fetch(oidcRegistrationPath, {
```
leekahung, line 68: There seems to be an error related to the fetch function in `Signup.jsx`. The URL for pod registration seems to be incorrect when testing it out on localhost `http://localhost:3000idp/register/` it seems to be missing the `/` between the server URL and the path for registration.

https://github.com/codeforpdx/PASS/assets/14917816/41f93798-1c4f-4d83-98f4-75ece44cf506

src/pages/Profile.jsx
```diff
@@ -9,14 +9,12 @@ import Box from '@mui/material/Box';
 import Button from '@mui/material/Button';
 import CircularProgress from '@mui/material/CircularProgress';
 import Typography from '@mui/material/Typography';
-// Model Imports
```
leekahung, line 12: I'm a bit confused over the removal of these comments one line comments for separating imports (Line 12 and Line 14). Thought they were useful in organizing them within component jsx files?

src/pages/Signup.jsx
leekahung: Nevermind. Think this was more of an .env thing. Got this fixed.

leekahung:
Been trying to replicate this registration flow with another browser to mimic case manager sending a client the signup link. Have managed to get a new pod registered (up to screenshot 4), but couldn't manage to get it to register with the original case manager.

https://github.com/codeforpdx/PASS/assets/14917816/8e4b34ed-7ed1-4a74-86e3-17535341fa76

When trying to see if the new pod exists, I've checked local_temp_server_files, but it's not there.

Screen Shot 2023-08-21 at 16 53 37

leekahung:
A bit confused. I've managed to get a new pod registered on the local solid server, and inside that pod there's now the contact information of the case manager (bug with profile information not showing correctly should be fixed in PR #379, please disregard viewing the client profile). However, it would seem that the case manager would still have to manually link up the client's pod via add contact. Was that intended?

https://github.com/codeforpdx/PASS/assets/14917816/0b9c6a37-069a-4905-904b-5a80493c6bbe

The account does exist now on localhost and the .acl in PASS seems to be setup for the case manager. Just not on the case manager's Contacts list.

Screen Shot 2023-08-21 at 17 14 41

Screen Shot 2023-08-21 at 17 16 13

src/pages/Profile.jsx
andycwilliams: Yes. And we should be sure to spell it "Modal", not "Model" like above

src/pages/Signup.jsx
```diff
@@ -0,0 +1,304 @@
...
+            aria-label="Email"
+            label="Email"
+            variant="outlined"
+            value={email}
+            onChange={(e) => setEmail(e.target.value)}
+            required
+          />
+          <br />
+          <TextField
+            style={textFieldStyle}
```
andycwilliams, line 183: I think it would help security if we automatically hide the password as the user enters it. That seems pretty standard practice. And then it could include a button to reveal it. Like this:



This also determines how strong the password is. We could add that as well. Probably not needed for MVP but might be nice to include down the line.

Another aspect to consider is whether we want to set minimum requirements for passwords, such as no fewer than 8 characters, at least one number, etc.

src/pages/Profile.jsx
leekahung: No model is correct, i.e. model for the data structure and schema. It's separate from the components that are modals that are used for the forms.

timbot1789:
> A bit confused. I've managed to get a new pod registered on the local solid server, and inside that pod there's now the contact information of the case manager (bug with profile information not showing correctly should be fixed in PR #379, please disregard viewing the client profile). However, it would seem that the case manager would still have to manually link up the client's pod via add contact. Was that intended?

For this PR, yes. Since it was unclear to me how we want to handle a new user registration on the case manager end (should it go to the case manager themselves, or to an organization signup list, or something else) I removed it from this PR. Didn't want to write up something that would just get deleted.

Of course, we can't leave clients fully orphaned after signup with no case manager to contact, so I added the agent the client signed up with to the client's contact list.

src/pages/Signup.jsx
timbot1789: I'll add this in. Is there a particular component you had in mind for handling these passwords?

src/pages/Profile.jsx
timbot1789: I'll add them back in, but generally I think that comments like these just add noise to the file, and the info can be better communicated with good naming.

e.g. is this harder to understand:
```js
import { ClientModel } from '@models'
import { ContactsModel } from '@models'
```
than this:
```js
// Model imports
import { ClientModel } from '@models'
import { ContactsModel } from '@models'
```

timbot1789: Fixed. The comments aren't identical because the profile component was refactored and given new imports.

src/pages/Signup.jsx
andycwilliams: This should cover everything we need:
https://mui.com/material-ui/react-text-field/#input-adornments

timbot1789: Sounds good. That won't handle the password evaluation at the bottom, but that can be a separate PR.

timbot1789: Fixed

src/pages/Signup.jsx
```diff
@@ -0,0 +1,341 @@
...
+  }, []);
+
+  useEffect(() => {
+    if (session.info?.isLoggedIn)
+      initializePod(session.info.webId, podUrl, caseManagerWebId, session.fetch);
+  }, [session.info.isLoggedIn]);
+
+  useEffect(() => {
+    if (isSuccess && caseManagerWebId) {
+      const [cmFirstName, cmLastName] = caseManagerName?.split(' ') ?? [null, null];
```
leekahung, line 300: Think we could address this in another issue, but for splitting up the name between given and family name, think we could be difficult to simply use split since there could be names like mine with `Ka Hung` being the given name and `Lee` being the family name.

I think a strategy people use to use for my mails would be to concat the given name so that it's `KaHung` and leave the final space alone so it'll be `KaHung Lee` in the mail or `Kahung Lee`. An alternative to this was to add new fields so we could explicitly add family and given names to the profile card.

leekahung (approved):
Outside of the comment for the name, I think this should be good to go.

src/pages/Profile.jsx
```diff
@@ -80,6 +84,13 @@ const Profile = () => {
     >
       <Box sx={{ display: 'flex', flexDirection: 'column', alignItems: 'center', gap: '20px' }}>
         <Typography sx={{ fontWeight: 'bold', fontSize: '18px' }}>Profile Information</Typography>
+        <Typography>
+          {client ?? (
```
andycwilliams, line 88: So I just tested out the latest version of Development and am getting an app-breaking error from this component, saying that `client is not defined`.

It seems like `client` in this line is supposed to be `contact` instead.

I see a couple of other times `client` is used in this component but those don't break functionality.

leekahung: Good catch. Think we could open a bug report for it.

PR #379, opened 2023-08-11
Issue 377/private profile info and Issue 387 fixing profile bug

leekahung:
## Create Private Profile Information on User's pod

---
## This PR:
**1.**    Generates a new `Profile` container (in /PASS) with a `privateProfile.ttl` (this container is separate from the public /profile container)
**2.**   A new ACL file will be set for Profile; a follow-up PR will be created to account for permission adjustments
**3.**   Including a single new field for the profile card using information from `privateProfile.ttl` as proof-of-concept
**4.**   Utilizing schema for date of birth (DOB) from `hmis-interop` for private profile (https://github.com/hmis-interop/logical-model/blob/FY2022-v1.0.0/src/logical-model.n3.owl#DOB)
**5.**   Included fixes for #387 where `client`, `clientProfile`, and `setClient` from `Profile`, `ProfileComponent`, `DocumentListContext`, etc. needs to be updated to `contact`, `contactProfile`, and `setContact`, respectively

Clip of Profile update with private profile:

https://github.com/codeforpdx/PASS/assets/14917816/60cfa735-b45c-40bc-9372-211b9ec4f0ed

Clip of bug fix for contacts:

https://github.com/codeforpdx/PASS/assets/14917816/e4a1e19f-cb91-402c-808e-9d6fb3b135d7

The format used for the HMIS schema follows what is instructed in the model overview documentation (https://github.com/hmis-interop/logical-model/blob/FY2022-latest/doc/HMIS-logical-model-overview.pdf, page 4)

---
### Components
- `src/components/Profile/ProfileComponents.jsx`
- `src/components/Profile/ProfileImageField.jsx`
- `src/components/Contacts/ContactListTableRow.jsx`
- `src/components/NavBar/NavMenu.jsx`

### Tests
- `test/model-helpers/Profile.test.jsx`
- `test/components/Profile/ProfileComponent.test.jsx`
- `test/components/Profile/ProfileImageField.jsx`

### Other Files
- `src/constants/rdf_predicates.js`
- `src/contexts/SignedInUserContext.jsx`
- `src/contexts/DocumentListContext.jsx`
- `src/model-helpers/Profile.js`
- `src/pages/Profile.jsx`
- `src/utils/pod-management/pod-helpers.js`

---
## Future Steps/PRs Needed to Finish This Work (optional):
- Set up permissions for private profile information
- Expand information stored in `privateProfile.ttl`
- Incorporate more HMIS definitions as part of profile
- Consolidate certain create container functions in `pod-helpers.js`

PR #388, opened 2023-08-18
Issue 358 - Putting SetAclPermissions into Modal

### LINKED ISSUE #358: Render permissions forms on profile page as modals after an onClick event only

brancwill:
---
name: Pull Request
about: Create a pull request to merge your code into the PASS codebase
title: ''issue-358/render-permission-forms"
labels: ''Enhancement"
assignees: ''brancwill"

---

**Same PR as #370 (See reason for closing in original PR)** I changed the `SetAclPermissionForm` and `SetAclPermsDocContainerForm` into modals. The former is now opened by a button that I've attached to each file in the Document Table to open the modal with the `File Name` set to said file's name. The latter is opened by a button next to the `Add Document` button.

---
## This PR:
**1.**    

---
### Components
`src/components/Form/SetAclPermissionsForm.jsx` (now `SetAclPermissionsModal.jsx`)
`src/components/Form/SetAclPermsDocContainerForm.jsx` (now `SetAclPermissionsModal.jsx`)
`src/components/Form/FormSection.jsx`
`src/components/Documents/DocumentTable.jsx`
`src/components/Documents/DocumentTableRow.jsx`

### Tests
List file names here

### Other Files
`src/pages/Profile.jsx`
`src/components/Modals/index.js`
`src/components/Form/index.js`
`src/typedefs.js`

---
Screenshot1
Screenshot2

---

---
@xscottxbrownx and I spoke about potentially condensing the table through either a menu containing the document actions (setting permissions, deletion, etc.) or through an accordion (@leekahung 's suggestion). If this is something we'd like to do, I can either add to this PR, or start another one.

---
**1.**  I'm pretty unfamiliar with MUI and JSDoc, so I copy and pasted a lot. With JSDoc in particular, I wasn't sure if I should add Modals to `SetAclPermissionsModal` and `SetAclPermsDocContainerModal`'s `@memberof` or not. Feedback would be appreciated, I'm happy to tweak anything.


src/typedefs.js
```diff
@@ -184,6 +184,33 @@ const React = require('react');
...
+ * @exports clientListTableRowProps
+ * @typedef {object} clientListTableRowProps
+ * @property {object} client - Object containing client information
+ * @property {React.Dispatch<React.SetStateAction<boolean>>} setShowDeleteClientModal
+ * - React set function for DeleteClientModal
+ * @property {React.Dispatch<React.SetStateAction<null>>} setSelectedClientToDelete
+ * - React set function for setting client to be deleted
+ * @memberof typedefs
+ */
+
```
leekahung: Recent changes in Development moved typedefs for Contacts within ContactListTable.jsx and ContactListTableRow.jsx, so these changes could actually be removed.

src/typedefs.js
```diff
@@ -212,7 +252,7 @@ const React = require('react');
  *
  * @exports profileComponentProps
  * @typedef {object} profileComponentProps
- * @property {object} [clientProfile] - Contact object with data from profile
+ * @property {object} [clientProfile] - Client object with data from profile
```
leekahung: I'm currently fixing these typedefs in PR #379. For now, think we could leave Line 255 alone and change this back to Contact.

The same could be said about Line 268 with `clientProfile`

src/components/Documents/DocumentTableRow.jsx
```diff
@@ -6,6 +6,7 @@ import { useSession } from '@hooks';
 import DeleteOutlineOutlinedIcon from '@mui/icons-material/DeleteOutlineOutlined';
 import FileOpenIcon from '@mui/icons-material/FileOpen';
 import IconButton from '@mui/material/IconButton';
+import ShieldIcon from '@mui/icons-material/Shield';
```
leekahung: I think we could leave the permission icons with the settings icon/gear.

Personally, I think that's more intuitive than a shield.

src/components/Modals/SetAclPermissionsModal.jsx
```diff
@@ -0,0 +1,213 @@
...
+      runNotification('Failed to set permissions. Reason: File not found.', 5, state, dispatch);
+    } finally {
+      setTimeout(() => {
+        clearInputFields();
+      }, 3000);
+    }
+  };
+
+  return (
+    <div>
```
xscottxbrownx: is there a need for this div wrapper?

src/components/Documents/DocumentTableRow.jsx
xscottxbrownx: Personally, I'm also not a big fan of the shield - BUT if we do use it, we've been using the outlined versions of the icons.  IF we use shield, I think my pick would be the `AdminPanelSettingsOutlined` icon.

src/components/Modals/SetAclPermissionsModal.jsx
```diff
@@ -0,0 +1,213 @@
...
+
+      runNotification(
+        `${permissions.read ? 'Give' : 'Revoke'} permission to ${
+          permissionState.podUrlToSetPermissionsTo
+        } for ${dataset.modalType === 'container' ? 'Documents Container' : docType}.`,
+        5,
+        state,
+        dispatch
+      );
+    } catch (error) {
```
xscottxbrownx, line 92: error message isn't quite right... it will always show as FILE NOT FOUND. That should be something else when giving permission to a single file.  Like if a failure in that case I assume it's because the other pod url isn't found or some other reason.

Maybe it could just be: `Reason: ${error.message}`      ?

brancwill:
@leekahung The changes have been made 👍

src/components/Modals/SetAclPermissionsModal.jsx
```diff
@@ -0,0 +1,213 @@
...
+ * SetPermissionsModal Component - Modal component that generates the form for
+ * setting ACL permissions to another user's documents, or document container
+ * in their Solid Pod via Solid Session.
+ *
+ * @memberof Modals
+ * @name SetAclPermissionsModal
+ * @param {setAclPermissionsModalProps} Props - Props for SetAclPermsDocContainerModal component
+ * @returns {React.JSX.Element} The SetAclPermsDocContainerModal Component
+ */
+
```
andycwilliams: If my understanding is correct, we want to remove the space in between JSDocs and their related functions

src/components/Modals/SetAclPermissionsModal.jsx
```diff
@@ -0,0 +1,213 @@
...
+              <DocumentSelection
+                htmlForAndIdProp="set-acl-doctype"
+                handleDocType={handleDocType}
+                docType={docType}
+              />
+            ) : null}
+
+            <br />
+            <FormControl fullWidth sx={{ display: 'flex', gap: 2, flexDirection: 'column' }}>
+              <Button
```
xscottxbrownx: something to add a GitHub Issue for later - these buttons are one on top of other. Other modals have the yes/no, confirm/cancel buttons side by side... 

Some modals have squared buttons, some have rounded pill shaped buttons.

We'll need to make these consistent.

src/components/Modals/SetAclPermissionsModal.jsx
```diff
@@ -0,0 +1,213 @@
...
+            </FormControl>
+
+            {dataset.modalType === 'document' ? (
+              <DocumentSelection
+                htmlForAndIdProp="set-acl-doctype"
+                handleDocType={handleDocType}
+                docType={docType}
+              />
+            ) : null}
+
```
andycwilliams: I'm curious how much this `DocumentSelection` will be used here. Isn't the document type already set and this modal is to just change permissions?

src/pages/Profile.jsx
```diff
@@ -89,24 +108,37 @@ const Profile = () => {
...
+          {!client && (
+            <Button
+              variant="contained"
+              color="primary"
+              size="small"
+              aria-label="Set Container Permissions Button"
+              startIcon={<SettingsIcon />}
+              onClick={() => handleAclPermissionsModal('container')}
+            >
+              Set Permission to All Documents
```
andycwilliams: I wonder if we could come up with more succinct phrasing for this text. But it could be difficult.

src/components/Modals/SetAclPermissionsModal.jsx
```diff
@@ -0,0 +1,213 @@
...
+  };
+
+  return (
+    <div>
+      <Dialog open={showModal} onClose={clearInputFields}>
+        <FormSection
+          title={
+            dataset.modalType === 'container'
+              ? 'Permission for all Documents'
+              : 'Permission for Document'
```
xscottxbrownx: Just a thought, not sure if it would look good - or be too susceptible of making the modal too wide...

but thought of possibly putting the file name in the title (like `: 'Permission for ${documentName}'` if it's setting permissions for a single document, instead of the `<Typography />` with `File Name: {documentName}`

If that worked/looked good/etc, we could get rid of line 117-131.

xscottxbrownx (commented):
Some thoughts on this before approving

src/components/Modals/SetAclPermissionsModal.jsx
leekahung: Yes, that is correct. It's so that the JSDoc comment can associate itself with the thing directly below it. The space between them in Line 37 could be removed.

src/components/Modals/SetAclPermissionsModal.jsx
leekahung: Yeah, there should be different error messages based on which "modal" is being used. The container version or the file version.

Both can't use the same `File not found.` as an error message.

src/components/Modals/SetAclPermissionsModal.jsx
leekahung: It looks like this component combines both permission modals together, so this would still be needed here.

But since that's the case, think we could use `&&` instead of a ternary with `?` and `null`.

src/components/Documents/DocumentTableRow.jsx
```diff
@@ -64,6 +65,11 @@ const DocumentTableRow = ({ document }) => {
           <FileOpenIcon />
         </IconButton>
       </StyledTableCell>
+      <StyledTableCell align="center">
+        <IconButton type="button" onClick={() => handleAclPermissionsModal('document', name)}>
+          <SettingsIcon />
```
andycwilliams: Just to test, I would like to try out other icons and see how we feel about them. For example, `LockOutlinedIcon` could be used for documents that have no permissions given, then it would switch to `LockOpenOutlinedIcon` whenever any sort of permission is granted. Or perhaps `KeyOffOutlinedIcon` and `KeyOutlinedIcon` instead.

My thinking is it might provide the user with a bit more information at a glance.

Haven't tested these myself yet but I'm curious.

src/components/Modals/SetAclPermissionsModal.jsx
xscottxbrownx: I BELIEVE what Andy is saying is that the modal that “needs” this, shouldn’t.  It’s the modal for setting permissions on a single file, correct?

That modal renders when the icon on the specific file is clicked, so we should be able to reference that specific file and pass it to the modal.  

This modal, the way it is now, should really only be necessary if we somehow got to it from an unspecified way - like just having a button at the top of the document list.

leekahung: Ah, I get it now. Given that the document is already selected per row, the document selection in this case is no longer necessary. I'm fine with that. We could actually remove this from this modal altogether.

xscottxbrownx:
Hey @brancwill,

just checking in with you and seeing how things are going?

Any questions, or clarification needed?

brancwill:
Hey @xscottxbrownx , sorry for my absence on this. I made a lot of the changes y'all suggested.

Screenshot (14)

I tried including the document name into the title. I don't think it's too wide, but I would like the text to be centered. Not sure how you do that with MUI though.

timbot1789:
> I tried including the document name into the title. I don't think it's too wide, but I would like the text to be centered. Not sure how you do that with MUI though.

@brancwill I think you can use the `align` property on the `Typography` element for that.

https://mui.com/material-ui/api/typography/#props

src/components/Documents/DocumentTableRow.jsx
timbot1789: I like the idea of a lock or key icon for permissions rather than a gear icon. Gear icon is a bit vague.

src/components/Documents/DocumentTable.jsx
```diff
@@ -32,6 +37,7 @@ const DocumentTable = () => {
     'Upload Date',
     'Expiration Date',
     'Preview File',
+    'Permissions',
```
timbot1789: Just a thought, and something we can revisit in a separate PR:

'Permissions' to me sounds a bit stiff and technical. What about naming this feature "sharing?" So instead of "setting permissions," you "share" or "unshare" a document.

src/components/Documents/DocumentTableRow.jsx
timbot1789: Or maybe even a sharing icon:
https://mui.com/material-ui/material-icons/?query=share&selected=Share

src/pages/Profile.jsx
timbot1789: "Share documents folder"

src/components/Documents/DocumentTable.jsx
brancwill: Align worked, thanks. Sharing sounds good to me, I can edit stuff to say that if you'd like. Or we can do it in another PR like you said.

src/components/Documents/DocumentTableRow.jsx
brancwill: I don't know how to get the info of whether a document has been shared or not. Maybe we could try that in another PR? I changed the gear to a Share Icon though.

src/components/Modals/SetAclPermissionsModal.jsx
brancwill: Got it, I changed it to `error.message`

src/components/Modals/SetAclPermissionsModal.jsx
brancwill: Nope, got rid of it

src/pages/Profile.jsx
brancwill: Sounds good, I changed it to that

src/components/Modals/SetAclPermissionsModal.jsx
brancwill: Ok, I removed it

src/components/Modals/SetAclPermissionsModal.jsx
brancwill: Cool, it's been removed

src/components/Modals/SetAclPermissionsModal.jsx
```diff
@@ -3,66 +3,91 @@ import React, { useContext, useState } from 'react';
...
+ * SetPermissionsModal Component - Modal component that generates the form for
+ * setting ACL permissions to another user's documents, or document container
+ * in their Solid Pod via Solid Session.
  *
- * @memberof Forms
- * @name SetAclPermissionForm
- * @returns {React.JSX.Element} The SetAclPermissionForm Component
+ * @memberof Modals
+ * @name SetAclPermissionsModal
+ * @param {setAclPermissionsModalProps} Props - Props for SetAclPermsDocContainerModal component
```
xscottxbrownx: no longer DocContainer

src/components/Modals/SetAclPermissionsModal.jsx
```diff
@@ -3,66 +3,91 @@ import React, { useContext, useState } from 'react';
...
+ * @returns {React.JSX.Element} The SetAclPermsDocContainerModal Component
```
xscottxbrownx: no longer DocContainer

src/components/Modals/SetAclPermissionsModal.jsx
```diff
@@ -3,66 +3,91 @@ import React, { useContext, useState } from 'react';
...
     try {
-      await setDocAclPermission(session, docType, permissions, podUrl, webIdToSetPermsTo);
+      switch (dataset.modalType) {
+        case 'container':
+          await setDocContainerAclPermission(session, permissions, podUrl, webIdToSetPermsTo);
+          break;
+        case 'document':
+          await setDocAclPermission(
+            session,
+            dataset.docType,
```
xscottxbrownx, line 76: according to `Profile.jsx` dataset has a `documentType`, not a `docType`.

src/components/Modals/SetAclPermissionsModal.jsx
```diff
@@ -84,29 +109,14 @@ const SetAclPermissionForm = () => {
...
-      statusType="Status"
-      defaultMessage="No action yet..."
-    >
-      <Box display="flex" justifyContent="center">
+    <Dialog open={showModal} onClose={clearInputFields}>
+      <FormSection state={state} statusType="Status" defaultMessage="No action yet...">
+        <Typography variant="h5" align="center" mb={4}>
+          {dataset.modalType === 'container'
+            ? 'Permission for all Documents'
+            : `Permission for ${dataset.docName}`}
```
xscottxbrownx: according to `Profile.jsx` `dataset` has a `documentName`, not a `docName`.

src/pages/Profile.jsx
```diff
@@ -104,24 +125,37 @@ const Profile = () => {
...
+              onClick={() => handleAclPermissionsModal('container')}
+            >
+              Share Documents Folder
+            </Button>
+          )}
+          <Button
+            variant="contained"
+            color="secondary"
+            size="small"
+            aria-label="Add Client Button"
```
xscottxbrownx: inaccurate `aria-label`

xscottxbrownx (changes requested):

src/components/Documents/DocumentTable.jsx
xscottxbrownx: I feel that's part of this PR. 

- The column title in the document table is added by this PR
- The modal to "share" the file or folder is added by this PR

So changing either of those should be part of this PR, in my opinion.

src/typedefs.js
```diff
@@ -196,13 +196,26 @@ const React = require('react');
  * @memberof typedefs
  */
 
+/**
+ * documentTableProps is an object that stores the props for the DocumentTable
+ * component.
+ *
+ * @exports documentTableProps
+ * @typedef {object} documentTableProps
+ * @property {(documentName: string, modalType: string) => void} handleAclPermissionsModal
```
leekahung: I've just checked the function `handleAclPermissionsModal`. It seems like it takes in 3 parameters now not 2 with documentName and modalType. The new parameters seems to be `modalType`, `docName`, and `docType`. The JSDoc would need to be updated accordingly. I believe they're all still just strings? So we just need to change this to:

```@property {(modalType: string, docName: string, docType: string) => void}```

This should be true for Line 217 as well for `documentTableRowProps`.

File: src/components/Documents/DocumentTable.jsx (hunk 12 above)
Comment (leekahung) [reply src/components/Documents/DocumentTable.jsx]: I like this idea. Think we could use "share/unshare" for permission setting.

**Review by leekahung — COMMENTED** [review]
I think that should be it on my end. I see that there are still things others pointed out that still needs to be addressed, I'm ready to approve if they're good as well.

File: src/components/Documents/DocumentTableRow.jsx (hunk 11 above)
Comment (andycwilliams) [reply src/components/Documents/DocumentTableRow.jsx]: I think determining whether or not it's been shared can be left for a future PR. We could even add a simple `useState` that toggles the icon when clicked, like with the pin in `ContactListTableRow.jsx`. Assuming we want to go that route.

**Comment by andycwilliams** [comment]
Quick request: Could we add a character limit to displaying the file name? Currently it stretches out the modal and adds a scroll bar at the bottom. I think we can cut it off after a point to simplify usage. Should be easy enough to add.



Here's an example from `UploadDocumentModal.jsx` starting on line 163:
```
<FormHelperText
  sx={{
    width: '200px',
    whiteSpace: 'nowrap',
    overflow: 'hidden',
    textOverflow: 'ellipsis'
  }}
>
  File to upload: {file ? file.name : 'No file selected'}
</FormHelperText>
```

xscottxbrownx (approved):
Looks to me like all concerns have been addressed.   Based on my testing, I approve code as is now.

leekahung (approved):
Aside from resolving merge conflicts, think this is good to go as well.

leekahung:
Hey @brancwill was wondering about the progress of this PR. If there's any problem like needing some help with the merge conflict, let us know. Thanks!

leekahung:
The merge came out pretty good, good work @xscottxbrownx! Still a bit iffy about switching the shared icon with an outlined version, but that's just cosmetic, the PR is still good to go.

How about @timbot1789 and @andycwilliams?

xscottxbrownx:
> The merge came out pretty good, good work @xscottxbrownx! Still a bit iffy about switching the shared icon with an outlined version, but that's just cosmetic, the PR is still good to go.


We can change it, the others were all outlined versions - so just made it consistent. But I'm not married to it, nor really care either way.

andycwilliams (approved):
I also prefer `ShareIcon` over `ShareOutlinedIcon` but am fine with either. I would like to get this merged sooner rather than later, so I'm good with approving this as is.

xscottxbrownx:
Switched back to `share` icon, instead of outlined version.

leekahung:
Think this is mostly ready. Going to merge this into Development. Any further issues related to this PR could be addressed in future PRs.

PR #392, opened 2023-08-21
Delete old notification system

russfraze:
---
name: Delete Old Notification System
about: This PR is to delete the useStatusNotification calls, the context, hook, and utility helpers that were a part of the older message notification system. There were also parts of the FormSection component and its props are deleted as well. 

---
I'm seeing that there are now setTimouts that are not calling anything and should be deleted as well. 

---
Please take a look.


russfraze:
I don't know why its saying this branch is out of date, I pulled from Development right before I pushed and its also telling me its up to date when I do a pull now.

xscottxbrownx:
> I don't know why its saying this branch is out of date, I pulled from Development right before I pushed and its also telling me its up to date when I do a pull now.

It's because the PR is set to merge into Master branch... change it to Development branch.

Screenshot 2023-08-21 at 12 44 51 PM

Click "edit" to the right of this

src/components/Form/SetAclPermissionForm.jsx
```diff
@@ -59,37 +53,21 @@ const SetAclPermissionForm = () => {
...
-        `${permissions.read ? 'Give' : 'Revoke'} permission to ${
-          permissionState.podUrlToSetPermissionsTo
-        } for ${docType}.`,
-        5,
-        state,
-        dispatch
-      );
       addNotification(
         'success',
         `${permissions.read ? 'Gave' : 'Revoked'} permission to ${
```
xscottxbrownx: I think the "to" should also be a ternary based on `permissions.read`.  I'd prefer it to be `${permissions.read ? 'to' : 'from'}` as just "to" doesn't really work for `Revoked permissions to {person} for {doctype}`.

src/components/Form/SetAclPermsDocContainerForm.jsx
```diff
@@ -56,37 +50,21 @@ const SetAclPermsDocContainerForm = () => {
...
-        `${permissions.read ? 'Give' : 'Revoke'} permission to ${
-          permissionState.podUrlToSetPermissionsTo
-        } for Documents Container.`,
-        5,
-        state,
-        dispatch
-      );
       addNotification(
         'success',
         `${permissions.read ? 'Gave' : 'Revoked'} permission to ${
```
xscottxbrownx: I think the "to" should also be a ternary based on `permissions.read`.  I'd prefer it to be `${permissions.read ? 'to' : 'from'}` as just "to" doesn't really work for `Revoked permissions to {person} for {doctype}`.

src/components/Form/SetAclPermissionForm.jsx
```diff
@@ -59,37 +53,21 @@ const SetAclPermissionForm = () => {
...
         } for ${docType}.`
       );
     } catch (error) {
-      runNotification('Failed to set permissions. Reason: File not found.', 5, state, dispatch);
       addNotification('error', 'Failed to set permissions. Reason: File not found.');
     } finally {
-      setTimeout(() => {
-        clearInputFields();
-      }, 3000);
+      setTimeout(() => {}, 3000);
```
xscottxbrownx: don't think this does anything or is needed.

src/components/Form/FormSection.jsx
```diff
@@ -5,7 +5,6 @@ import Box from '@mui/material/Box';
 import Typography from '@mui/material/Typography';
```
xscottxbrownx, line 5: `FormSection` will also be deemed completely unnecessary and deleted once PR #388 gets merged in, and this one gets merged in.  It's pretty much unnecessary without the old notification system. Doesn't really add anything.  We can copy the `LogoutModal` pretty much where the title is in `<DialogTitle />` of `<Dialog />`.

src/components/Form/SetAclPermsDocContainerForm.jsx
```diff
@@ -56,37 +50,21 @@ const SetAclPermsDocContainerForm = () => {
...
         } for Documents Container.`
       );
     } catch (error) {
-      runNotification('Failed to set permissions. Reason: File not found.', 5, state, dispatch);
       addNotification('error', 'Failed to set permissions. Reason: File not found.');
     } finally {
-      setTimeout(() => {
-        clearInputFields();
-      }, 3000);
+      setTimeout(() => {}, 3000);
```
xscottxbrownx: don't think this does anything

src/components/Form/SetAclPermsDocContainerForm.jsx
```diff
@@ -56,37 +50,21 @@ const SetAclPermsDocContainerForm = () => {
...
-      );
       addNotification(
         'success',
         `${permissions.read ? 'Gave' : 'Revoked'} permission to ${
           permissionState.podUrlToSetPermissionsTo
         } for Documents Container.`
       );
     } catch (error) {
-      runNotification('Failed to set permissions. Reason: File not found.', 5, state, dispatch);
       addNotification('error', 'Failed to set permissions. Reason: File not found.');
```
xscottxbrownx: Not sure if this "Reason" is correct.

src/components/Modals/AddContactModal.jsx
```diff
@@ -12,7 +11,6 @@ import FormControl from '@mui/material/FormControl';
 import IconButton from '@mui/material/IconButton';
 import InputAdornment from '@mui/material/InputAdornment';
 import TextField from '@mui/material/TextField';
-import { runNotification } from '@utils';
 import { FormSection } from '../Form';
```
xscottxbrownx, line 14: Can we split these imports with comments, just to keep all files consistent?  Got component and hook in Material UI

andycwilliams:
So issue #301 was tied to another PR but was never closed. Since this is related, maybe we could connect that issue, and then finally close it once this is merged.

src/components/Modals/AddContactModal.jsx
```diff
@@ -103,16 +58,9 @@ const AddContactModal = ({ addContact, showAddContactModal, setShowAddContactMod
...
     try {
       await addContact(userObject);
     } finally {
-      runNotification(
-        `"${userObject.givenName} ${userObject.familyName}" added to client list`,
-        5,
-        state,
-        dispatch
-      );
       addNotification(
```
xscottxbrownx, line 65: I don't think this whole try/finally block is correct. 

- I believe the successful `addNotification()` needs to be in the `try` block after the `await`.
- We need a `catch` block.
- This `setTimeout()` clears the inputs after 2 seconds and closes modal. <-- I'm not sure the others have this. Or is necessary. I don't know, but we should figure it out and make it consistent with rest, whichever way it needs to be.

My personal thoughts are clearing the inputs is great (if it's necessary), and maybe modal should only close on a `success`? So, if `error` the user can try again easier?

src/components/Modals/AddContactModal.jsx
```diff
@@ -134,12 +81,7 @@ const AddContactModal = ({ addContact, showAddContactModal, setShowAddContactMod
       aria-labelledby="dialog-title"
```
xscottxbrownx, line 84: this `aria` isn't linked to anything. Nothing with this id.   It will be necessary when we use `<DialogTitle />` and `<DialogContent />` after removing `FormSection`

src/components/Modals/DeleteContactModal.jsx
```diff
@@ -77,12 +59,7 @@ const DeleteContactModal = ({
       aria-describedby="dialog-description"
```
xscottxbrownx, line 60: The `aria-labelledby` is not linked to anything again. Will be when use `<DialogTitle />` instead of `<FormSection />`         Can probably switch it now if not using the spinner from `FormSection`.

src/components/Modals/UploadDocumentModal.jsx
```diff
@@ -69,11 +64,8 @@ const UploadDocumentModal = ({ showModal, setShowModal }) => {
     e.preventDefault();
 
     if (!docType) {
-      runNotification('Upload failed. Reason: No document type selected.', 5, state, dispatch);
       addNotification('error', 'Upload failed. Reason: No document type selected.');
-      setTimeout(() => {
-        dispatch({ type: 'CLEAR_PROCESSING' });
-      }, 3000);
+      setTimeout(() => {}, 3000);
```
xscottxbrownx: don't think this `setTimeout()` does anything

src/components/Modals/UploadDocumentModal.jsx
```diff
@@ -69,11 +64,8 @@ const UploadDocumentModal = ({ showModal, setShowModal }) => {
     e.preventDefault();
 
     if (!docType) {
```
xscottxbrownx: I haven't tested, but I don't think this is needed either...  I believe the `DocumentSelection` is `required` - so never be no `docType`...

xscottxbrownx (changes requested):
After this PR, the other one for turning the permission forms into modals, and my PR for a single ConfirmationModal to replace a few - I think we need to make another issue to make the modals consistent.

Button placement, pill over rounded boxes, using DialogTitle instead of FormSection, etc.

xscottxbrownx:
> So issue #301 was tied to another PR but was never closed. Since this is related, maybe we could connect that issue, and then finally close it once this is merged.

interesting... I guess it depends on how people feel about the intermediary notifications (something's happening). 

I do feel the user should be alerted that something is happening while the async function is running (previous to receiving a success or error snackbar notification.)

Could be done with loading button, an app-wide listening spinner animation component, etc...

---

We could also delete that part from the issue and make a separate issue for this specific topic.

src/components/Form/FormSection.jsx
```diff
@@ -48,12 +39,6 @@ const FormSection = ({
     </Typography>
     {children}
     {showSpinner && <CircularProgress />}
```
leekahung: Still a bit confused over the spinner here. I'm not sure where it's being used when going over the notifications.

src/components/Form/SetAclPermissionForm.jsx
```diff
@@ -43,14 +42,9 @@ const SetAclPermissionForm = () => {
...
   };
 
-  const clearInputFields = () => {
-    dispatch({ type: 'CLEAR_PROCESSING' });
-  };
-
   // Event handler for setting ACL permissions to file container on Solid
   const handleAclPermission = async (event) => {
     event.preventDefault();
-    dispatch({ type: 'SET_PROCESSING' });
```
leekahung, line 53: The original intent for `dispatch({ type: 'SET_PROCESSING' });` was to set a state for the button in use for these forms/modals to be disabled so that it'll prevent additional clicks for that button when the function is running.

Could you set up a new useState in place of this (i.e. `processing` remains true until it reaches the finally block where the setTimeout is located, so something like `setProcessing(true)`?)

The same would be applied for all the other dispatch calls being removed.

src/components/Form/SetAclPermissionForm.jsx
leekahung: Think we could keep this for a new `setProcessing(false)` for the button?

src/components/Form/SetAclPermissionForm.jsx
```diff
@@ -43,14 +42,9 @@ const SetAclPermissionForm = () => {
     setDocType(event.target.value);
   };
 
-  const clearInputFields = () => {
-    dispatch({ type: 'CLEAR_PROCESSING' });
```
leekahung, line 47: Since we're removing the old dispatch call, we can replace it with the set function from `useState`, something like `setProcessing(false)`.

The same would be applied for all the other dispatch calls being removed.

leekahung (changes requested):
See feedback.

src/components/Form/SetAclPermissionForm.jsx
russfraze: Screen Shot 2023-08-22 at 7 57 34 AM
Like this?

src/components/Form/FormSection.jsx
russfraze: This means I don't need to delete it now then? Trying to determine if this is an actionable comment.

src/components/Form/SetAclPermsDocContainerForm.jsx
russfraze: This currently does nothing. I can delete it unless we are going with Ka Hung's comment of keeping them for a future setProcessing(false), which I assume will be part of a different PR?

src/components/Modals/AddContactModal.jsx
russfraze: I'm wondering if fixing this is a part of this PR or has to do with the modals? My aim in this PR was to get rid of the old notification system and I'm wondering if this has to do with revamping this modal.

src/components/Form/SetAclPermsDocContainerForm.jsx
timbot1789: If it doesnt do anything, we can get rid of it. It's in the git history, so we can restore it later if we need

timbot1789: Although looking at Ka Hung's comment, he's describing an issue that would need to be fixed in this PR.

src/components/Modals/UploadDocumentModal.jsx
russfraze: Yeah this is a required field so it wont run. Deleting it.

src/components/Form/FormSection.jsx
russfraze: The spinner was taken out when I was trying to separate the implementation of the notifications from the PR where the system was created.

Either way it appeared in the modal while the async function was running. I feel the spinner is a better solution than a message because of how quickly it happens. Flashing a message too quickly for the user to read it is not needed if a loading spinner can tell them that something is happening. 

I do think the spinner could be deleted in this issue and I could set up a separate one which might be a better Idea if Form Section is going away.

src/components/Form/SetAclPermissionForm.jsx
russfraze: I made this change.

src/components/Form/SetAclPermsDocContainerForm.jsx
russfraze: I've addressed Ka Hung's comment and now the timeouts are being used to switch the processing state.

src/components/Form/SetAclPermissionForm.jsx
```diff
@@ -59,37 +55,23 @@ const SetAclPermissionForm = () => {
...
-        `${permissions.read ? 'Gave' : 'Revoked'} permission to ${
+        `${permissions.read ? 'Gave permission to' : 'Revoked permission from'} ${
           permissionState.podUrlToSetPermissionsTo
         } for ${docType}.`
       );
     } catch (error) {
-      runNotification('Failed to set permissions. Reason: File not found.', 5, state, dispatch);
       addNotification('error', 'Failed to set permissions. Reason: File not found.');
     } finally {
       setTimeout(() => {
```
timbot1789: We can still remove the `setTimeout` though. No reason to wait 3 seconds to clear processing if processing is finished

src/components/Form/SetAclPermsDocContainerForm.jsx
xscottxbrownx: Just like the other one you did

src/components/Form/SetAclPermissionForm.jsx
russfraze: I can remove the setTimouts in these modals, and the modals will go away immediately instead of the two second delay it currently has. If we do that there will be no need for the setProcessing state because the button will be instantly gone as well.  

Let me know if this is the way we want to go.

leekahung: No, we'll still need the `setProcessing` state to disable the button while it's running to prevent duplicate event calls. For example, we don't want a user to spam upload documents while the file is already processing under the hood after the first click.

It's separate from the `setTimeout`, which was originally meant to just delay it from the notification time from the notification status. But since we're moving it to a snack bar, I can see the point of not needing `setTimeout` anymore.

xscottxbrownx: Maybe we can do something like this:

Screenshot 2023-08-23 at 8 45 25 AM

That way if the async function takes a while to process for whatever reason - the button to run it again is disabled. Then modal disappears after receiving a Resolve or Reject.

?

Keep in mind some of these aren't modals yet in this PR because Brandon's hasn't been merged.

src/components/Form/FormSection.jsx
xscottxbrownx: I would still make the changes you did for now... Just kind of depends on whose PR gets merged first I guess. If Brandon's, then I think this whole file can be deleted.

src/components/Form/SetAclPermissionForm.jsx
xscottxbrownx: Maybe this whole discussion would be faster as a quick call.   I'm still not sure exactly when we want the modals to disappear.

If we are going to use some sort of animation in the modal itself (loading button was what I proposed) to show user that something is happening, then we'd want it to disappear after the task is complete.

If we are going to use a separate component to show user that something is happening, then we'd want it to disappear immediately after button press.


So, the answer to this effects my new PR work, and the next one to make all the modals more consistent in styling/look.

src/components/Form/FormSection.jsx
russfraze: Taking the spinner out. Let me know if we want to open a new issue to add another iteration of the spinner in a different PR.

src/components/Modals/AddContactModal.jsx
xscottxbrownx: Personally, I still see this as "implementing the new messaging system" - I should've caught it on last PR

src/components/Form/SetAclPermissionForm.jsx
```diff
@@ -151,12 +131,7 @@ const SetAclPermissionForm = () => {
           />
 
           <FormControl fullWidth sx={{ marginTop: '2rem' }}>
-            <Button
-              variant="contained"
-              disabled={permissionState.podUrlToSetPermissionsTo === podUrl || state.processing}
-              type="submit"
-              color="primary"
-            >
+            <Button variant="contained" disabled={processing} type="submit" color="primary">
```
xscottxbrownx: The previous disabling had the button also disabled when the user put in their own podUrl - so they couldn't try to change permissions of their own pod.   Think we need to add this back in, unless you found another way to address that possibility.

src/components/Form/SetAclPermsDocContainerForm.jsx
```diff
@@ -127,12 +107,7 @@ const SetAclPermsDocContainerForm = () => {
           </FormControl>
           <br />
           <FormControl fullWidth>
-            <Button
-              variant="contained"
-              disabled={permissionState.podUrlToSetPermissionsTo === podUrl || state.processing}
```
xscottxbrownx, line 132: The previous disabling had the button also disabled when the user put in their own podUrl - so they couldn't try to change permissions of their own pod.   Think we need to add this back in, unless you found another way to address that possibility.

src/components/Modals/UploadDocumentModal.jsx
```diff
@@ -183,7 +159,6 @@ const UploadDocumentModal = ({ showModal, setShowModal }) => {
               </Button>
               <Button
                 variant="contained"
-                disabled={state.processing || !file}
```
xscottxbrownx, line 186: Still need this to work. Without it, you can click upload as soon as you've selected a document type and it errors in console.  We want it so user can't try until we've got the bare minimum we need (doctype and an uploaded file.)

src/components/Form/SetAclPermsDocContainerForm.jsx
russfraze: Added this back in.

src/components/Modals/UploadDocumentModal.jsx
russfraze: Fixed this.

src/components/Modals/AddContactModal.jsx
russfraze: Made these changes in the try catch block.

leekahung:
It's related to the `handleDeleteDocument` function in `DocumentTableRow.jsx` not having the notification hooked up to the new snackbars. There are no notifications triggered whenever a document is deleted.

https://github.com/codeforpdx/PASS/assets/14917816/add091ed-c504-4373-ae6e-38c90150d657

Think we could do a quick inclusion of the new hook for that component.

src/components/Form/FormSection.jsx
```diff
@@ -21,15 +19,7 @@ import { StatusNotification } from '../Notification';
...
-const FormSection = ({
-  title,
-  state,
-  showSpinner,
-  statusType,
-  defaultMessage,
-  file = null,
-  children
-}) => (
+const FormSection = ({ title, children }) => (
```
leekahung, line 22: Given the removal of the older notification props, think you could update the typedef of `formSectionProps` in `typedef.js`. See screenshot that they're still defining the old props.

Screen Shot 2023-08-28 at 14 41 41

With the exception of title and children, think the rest of the definitions can be removed.

src/components/Form/SetAclPermsDocContainerForm.jsx
leekahung: Yeah, the reason for this error is not correct. This is the error statement for failing to set permissions for the Documents container. For now, I think we could use `error.message`.

src/pages/Profile.jsx
```diff
@@ -89,7 +89,7 @@ const Profile = () => {
       <Box sx={{ display: 'flex', flexDirection: 'column', alignItems: 'center', gap: '20px' }}>
         <Typography sx={{ fontWeight: 'bold', fontSize: '18px' }}>Profile Information</Typography>
         <Typography>
-          {client ?? (
+          {contact ?? (
```
leekahung: The bug where it breaks the webpage when clicking on the contact URL is not really a problem with this PR. A full fix for this should be ready when PR #394 gets merged into to Development and this branch getting updated.

leekahung (changes requested):
I think we're close. Outside of those things I've just commented on, everything else looks good.

src/components/Form/FormSection.jsx
russfraze: Updated.

leekahung (approved):
Think what we have here should be good from my end.

xscottxbrownx:
Need to fix conflicts in `Profile.jsx` and address the error message in `SetAclPermsDocContainerForm.jsx` still - one of the few unresolved conversations above.

Using the `resolve conversation` feature of github is great, so we know when everything has been addressed properly.

xscottxbrownx (approved):
Seems to me like everything was addressed!

russfraze:
With this merge, new snackbar style notifications should now be implemented app wide and the old notification system deleted.

PR #394, opened 2023-08-27
Bug fix for contacts page

leekahung:
## Bug Fix for Contacts Page

---
## This PR:
**1.**    Fixes the bug where the website fails to render after login due to undefined variable `client` by changing it to `contact` and modifying the area which are causing the issue
**2.**   Also removes the unused `link` element from index.html for the original css file

---
### Components/Pages
- `src/pages/Profile.jsx`

### Other Files
- `index.html`

---

Screenshot of fix:

https://github.com/codeforpdx/PASS/assets/14917816/1de0d2d0-ed42-4b28-9064-07f69719033d

Screenshot of signup link rendering for signups page:

https://github.com/codeforpdx/PASS/assets/14917816/1c51add1-2782-4773-94ea-bdc8d7cc2a20

### DIFF
--- a/src/pages/Profile.jsx
+++ b/src/pages/Profile.jsx
@@ -42,7 +42,7 @@ const Profile = () => {
   const [loadingProfile, setLoadingProfile] = useState(true);
 
   useEffect(() => {
-    const fetchClientProfile = async () => {
+    const fetchContactProfile = async () => {
       const profileData = await fetchProfileInfo(session, webIdUrl);
       setContactProfile({
         ...contact,
@@ -52,7 +52,7 @@ const Profile = () => {
     };
 
     if (contact) {
-      fetchClientProfile();
+      fetchContactProfile();
     } else {
       setContactProfile(null);
     }
@@ -108,7 +108,7 @@ const Profile = () => {
           variant="contained"
           color="secondary"
           size="small"
-          aria-label="Add Client Button"
+          aria-label="Add Contact Button"
           startIcon={<AddIcon />}
           onClick={() => setShowModal(true)}
         >
--- a/src/pages/Profile.jsx
+++ b/src/pages/Profile.jsx
@@ -108,7 +108,7 @@ const Profile = () => {
           variant="contained"
           color="secondary"
           size="small"
-          aria-label="Add Contact Button"
+          aria-label="Add Document Button"
           startIcon={<AddIcon />}
           onClick={() => setShowModal(true)}
         >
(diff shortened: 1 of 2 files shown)


andycwilliams (approved):
Good quick fix

Only question is whether `client` should also be changed to `contact` on lines 45 and 111

leekahung:
> Good quick fix
> 
> Only question is whether `client` should also be changed to `contact` on lines 45 and 111

Yeah, I think I could do that before merging in the fixes. I'll check any of the test files, too, in case it fails one of them.

src/pages/Profile.jsx
```diff
@@ -88,13 +88,13 @@ const Profile = () => {
     >
       <Box sx={{ display: 'flex', flexDirection: 'column', alignItems: 'center', gap: '20px' }}>
         <Typography sx={{ fontWeight: 'bold', fontSize: '18px' }}>Profile Information</Typography>
-        <Typography>
-          {client ?? (
+        {!contact ? (
```
leekahung, line 91: Initially wanted to keep `??` but I couldn't get it to work. Had to use ternary instead.

xscottxbrownx: Screenshot 2023-08-27 at 7 21 22 PM

This didn't work? This is what Russell has in his PR for deleting notification system.

src/pages/Profile.jsx
```diff
@@ -108,7 +108,7 @@ const Profile = () => {
           variant="contained"
           color="secondary"
           size="small"
-          aria-label="Add Client Button"
+          aria-label="Add Contact Button"
```
xscottxbrownx: Not adding a client/contact here - I commented on this in Brandon's PR... This is adding a document button.

src/pages/Profile.jsx
leekahung: It didn't.

I'm getting a blank page with a rendering error on the console log for some reason. I'm still not sure what's going on since this condition is also being used for the `set permission` forms (that would be converted soon), but it's seems to be related to contact being passed in as the condition. (see clip)

https://github.com/codeforpdx/PASS/assets/14917816/cf7b1035-ee84-462c-856c-bd8833d846e1

The ternary seems to work out though as mentioned and shown (see clip below).

https://github.com/codeforpdx/PASS/assets/14917816/c4e750f5-370c-4a50-8439-7a38adf1809f

src/pages/Profile.jsx
leekahung: Ah, good catch. I see it now. I'll have this fixed.

leekahung:
Hey @xscottxbrownx and @timbot1789, I've just updated PR based on the new comments. Let me know if there's still anything else that would need to be corrected. Would like to merge this in as soon as possible to fix the existing rendering bug.

timbot1789 (approved):

timbot1789 (commented):
this looks good. merging

PR #395, opened 2023-08-30
Only parse valid contacts

timbot1789:
Simple fix to filter out entries in the contacts document with invalid data. Should make things a little cleaner for people working on older data, and should reduce weird crashes in the UI


leekahung (approved):
I'm good with this.

PR #400, opened 2023-09-04
Issue 399/create message read status

### LINKED ISSUE #399: Enhancement: Unread/Read status of messages
With the message system in place, we don't have a system set for if a message was read or unread. By including a status for `read` in the message object we could keep track on read/unread messages using that. This feature would really just be needed for the inbox and nothing else.

---
By including a `read` status property for the message object, we could the value toggled from `false` to `true` after viewing the message for the first time by clicking on for a preview of the message.

---
### Rationale:
As stated in the description for this feature, by including a `read` status as part of the message object, PASS could keep track of read/unread messages of the user.

---
We could include this `read` status as part of the message object. This status could be toggled from `false` to `true` after viewing the message for the first time by clicking on for a preview of the message.

To make this backwards compatible with older message that doesn't have the read status, we could simply have them behave as if they're unread and add the unread status to the old message.

leekahung:
## This PR:
**1.**    Includes a new read status as part of message object
**2.**   Utilizes the new read status to indicate messages that has been read by altering the opacity of the header section of the message preview
**3.**   Made changes compatible with older message objects that doesn't have read status by making them false by default then including a new read status as part of the message
**4.**   Made changes compatible with older message objects that doesn't have read status by generating a new read status property for the message object when message preview is clicked and is set to true

---
### Components
- `src/components/Messages/MessagePreview.jsx`

### Other Files
- `src/constants/rdf_predicates.js`
- `src/contexts/MessageContext.jsx`
- `src/typedefs.js`
- `src/utils/network/session-core.js`
- `src/utils/network/session-helper.js`

---

When read status is updated:

https://github.com/codeforpdx/PASS/assets/14917816/aa109f0f-4bed-46ad-9094-aa0543a0cab6

When read status is updated on older message objects from Solid:

https://github.com/codeforpdx/PASS/assets/14917816/04125daf-6d49-4b60-9427-f275adbc4896

---
- Utilize the read status as an indicator for the message icon on whether or not there are unread messages or a number of unread messages

---
**1.**    PR related to Issue #399


xscottxbrownx:
### I really like that you went with just changing opacity over making a `NEW` icon or word or something!  

---

1) Do you have any thoughts as to why the first video seems to change opacity immediately, and second video takes a while?

2) May be just out of scope for this PR, but thoughts on using a `.map()` in `MessagePreview` on the `Grid` items - so if we make future changes it doesn't need to be written 3 times?  Something like:
```js

const messageInfo = [
    {
        title: 'Sender: ',
        text: {message.sender}
    },
    {
        title: 'Subject: ',
        text: {message.title}
    },
    {
        title: 'Date: ',
        text: {message.uploadDate.toLocaleDateString()}
    }
];

{messageInfo.map( message => (
    <Grid item xs={3} sx={{ opacity: message.readStatus ? '0.5' : '1' }}>
        <Typography> {message.title} <strong> {message.text} </strong> </Typography>
     </Grid>   
))
```

It's just a thought. We'd be adding more lines now (due to setting up the array) - but MAY be easier to maintain in future.

leekahung:
> 1. Do you have any thoughts as to why the first video seems to change opacity immediately, and second video takes a while?

Yeah, I think it's because the first one was part of the new updates where the read status property already exist, so it's just a matter of switching the value from `false` to `true`.

However in the older message object in the second clip, the read status doesn't exist at all. So the function has to generate a new Thing for the RDF, which then gets reflected as read status `true`.

> 2. May be just out of scope for this PR, but thoughts on using a `.map()` in `MessagePreview` on the `Grid` items - so if we make future changes it doesn't need to be written 3 times?
>
> It's just a thought. We'd be adding more lines now (due to setting up the array) - but MAY be easier to maintain in future.

No problem with using `.map`, think it's a good idea. I'll refactor it here with this PR.

xscottxbrownx:
> Yeah, I think it's because the first one was part of the new updates where the read status property already exist, so it's just a matter of switching the value from `false` to `true`.
> 
> However in the older message object in the second clip, the read status doesn't exist at all. So the function has to generate a new Thing for the RDF, which then gets reflected as read status `true`.
> 

ahh yes. Makes sense. Thank you.

xscottxbrownx (approved):

PR #403, opened 2023-09-06
Generate pdf demo

timbot1789:
This is a POC of a system that lets you fill out a pdf template using the data in your pod. The "pdfdesigner" contains a UI with 2 parts. The left column lets you query documents in your pod using SPARQL. The right side lets you build PDF templates that take in the resulting data of the query. None of the UI is final. Or good. Or really thought out for more than 5 minutes.

This adds two main dependencies:

- The SPARQL/GraphQL query engine, [Comunica](https://comunica.dev/) This engine was built by the same guy who built community solid server, and is being maintained by one of his grad students, as well as other members of the solid project
- The PDF rendering engine and UI, [PDFME](https://github.com/pdfme/pdfme). This creates PDF templates in simple JSON, making them pretty easy to integrate into other parts of our system. It also provides a simple editing UI so we don't have to build one from scratch (although we probably need to fork it).


andycwilliams:
This is amazing. That's my comment for now.

PR #405, opened 2023-09-10
389 Revamped Home Page

### LINKED ISSUE #389: Revamp Home Page
To date, the Home Page has been minimal and lacking in information.

---
Update the current Home Page (a box with that simply states "HOME PAGE") with a more detailed one.

---
### Rationale:
Home pages benefit from informing users, particularly first-time ones, about the background and features of the app. They also provide a welcoming and appealing initial experience.

---
The UX team is in the process of designing a home page. The assignee should consult with them before beginning work.

Raspber:
# Revamped Home Page

---
## This PR:
**1.** Added some basic images along with filler text so the home page is not empty.
**2.** Request demo button call to action.  
**3.** Moved assets to public folder

---
### Components

- Home.jsx
-HomeSection.jsx
-KeyFeatures.jsx

### Tests

No test files are affected

### Other Files

- app.png
- key-features-image.png
- web_security.png

---
Current green theme with revamped home page
revamped-home-page

---

---


src/pages/Home.jsx
```diff
@@ -13,26 +19,247 @@ import Paper from '@mui/material/Paper';
...
+            justifyContent="center"
+          >
+            <Grid item xs={12}>
+              <Box
+                component="img"
+                src="./src/assets/web_security.png"
+                alt="web-security-image"
+                sx={{
+                  width: '90%', // Default to 90%
+                  '@media (min-width:600px) and (max-width:959px)': {
```
leekahung: With MUI components, we're using the default breakpoints for responsive components, see example with the Typography component below this one with `xs`, `sm`, and `md`. Why not utilize those? Or if we need, why not just create custom breakpoints for MUI in `theme.js` to include these if we want to keep these breakpoints?

src/pages/Home.jsx
```diff
@@ -13,26 +19,247 @@ import Paper from '@mui/material/Paper';
...
-      <Paper elevation={2} sx={{ display: 'inline-block', mx: '2px', padding: '20px' }}>
+const Home = () => {
+  const theme = useTheme();
+  const isReallySmallScreen = useMediaQuery(theme.breakpoints.down('sm'));
+  return (
+    <Container component="main" maxWidth="s">
+      <Box
+        sx={{
+          marginTop: 18,
+          marginBottom: 18,
```
leekahung: Think we could replace this with margin: "18px 0"

src/pages/Home.jsx
```diff
@@ -13,26 +19,247 @@ import Paper from '@mui/material/Paper';
...
-        textAlign: 'center',
-        padding: '20px'
-      }}
-    >
-      <Paper elevation={2} sx={{ display: 'inline-block', mx: '2px', padding: '20px' }}>
+const Home = () => {
+  const theme = useTheme();
+  const isReallySmallScreen = useMediaQuery(theme.breakpoints.down('sm'));
+  return (
+    <Container component="main" maxWidth="s">
```
leekahung: I don't think `maxWidth="s"` is a valid attribute for `Container`?

src/pages/Home.jsx
```diff
@@ -13,26 +19,247 @@ import Paper from '@mui/material/Paper';
...
+                    width: '45%'
+                  }
+                }}
+              />
+              <Typography
+                sx={{
+                  color: 'temporaryPurple.dark',
+                  fontSize: { xs: '2rem', sm: '2rem', md: '3rem' },
+                  textAlign: 'center',
+                  mb: '24px'
```
leekahung: I've noticed you're using `mb` for some `sx` attributes and `margin-bottom` for others. Think we should stick with one version to avoid confusion.

src/pages/Home.jsx
leekahung, line 1: Think we could split the contents within the Home page into three smaller components or one reusable component where you could place an image at the top, a title in the middle, and everything else below:

```
<Box>
  <img />
  <strong>title</strong>
  <other content>
</Box>
```

And use a `.map` to generate the three sections. Either approach would do.

src/theme.js
```diff
@@ -28,6 +29,13 @@ const theme = createTheme({
     neutral: {
       main: '#64748B',
       contrastText: '#fff'
+    },
+    temporaryPurple: {
```
leekahung: I think you don't need to state that it's temporary. If we wish to change the color we can simply change it to purple.

leekahung (changes requested):
See feedback.

src/pages/Home.jsx
Raspber: Will add the break points to theme.js

src/pages/Home.jsx
Raspber: This sounds like a great idea. Ill do this.

Raspber:
> See feedback.

Will make the changes required👍🏽

src/pages/Home.jsx
```diff
@@ -1,38 +1,264 @@
...
+import { useTheme } from "@mui/material/styles";
+import SecurityIcon from "@mui/icons-material/Security";
+import Diversity1Icon from "@mui/icons-material/Diversity1";
+import SupportIcon from "@mui/icons-material/Support";
 /**
  * Home - First Page you encounter in PASS before login.
  * Should not display if you are already logged in
  *
- * @memberof Pages
+ * @memarginBottomerof Pages
```
xscottxbrownx: should be `@memberof`

src/pages/Home.jsx
```diff
@@ -1,38 +1,264 @@
...
+          flexDirection: "column",
+          alignItems: "center",
+          textAlign: "center",
+          padding: "20px"
+        }}
+      >
         <section id="home">
-          <h1>HOME PAGE</h1>
+          <Stack
+            width={isReallySmallScreen ? 1 : 3 / 3}
```
xscottxbrownx: is this not the same?

src/pages/Home.jsx
```diff
@@ -1,38 +1,264 @@
...
+                  },
+                  "@media (min-width:1440px)": {
+                    width: "45%"
+                  }
+                }}
+              />
+              <Typography
+                sx={{
+                  color: "temporaryPurple.dark",
+                  fontSize: { xs: "2rem", sm: "2rem", md: "3rem" },
```
xscottxbrownx: not sure this is optimized... I'd have to look at our other code, but not sure if you need the `sm`.   Happens a lot in this file.

xscottxbrownx (changes requested):
I back KaHung's comment on making the homepage reusable components. It'll be much easier to change styling/maintain in future.

---

Another issue is the images. We are going to run into the same issue we did with navbar logo.  

Maybe we need to merge KaHung's fix for that before this PR, then move these assets to that folder?

src/pages/Home.jsx
xscottxbrownx: What I mean by this is that I believe MUI to be "mobile-first" like Tailwind and Bootstrap. It may stay at that size until you tell it otherwise.  But again, I'm not positive.    I may be mixing it up with MUI media queries.

src/pages/Home.jsx
Raspber: I believe i have fixed this.

src/assets/app.png
leekahung, line 1: The image assets should be moved to the new `public` directory so that they can be rendered in the production build. Previously, the image files did not render if we were to source from `src/assets` and the image links all broke in production even though it was fine on the development build.

This is more of a build thing by Vite, but it should be a quick fix. Check out the recent PR #407 bug fix for guidance on what to do.

src/components/Home/ReusableHomeSection.jsx
```diff
@@ -0,0 +1,55 @@
+// React Imports
+import React from 'react';
+// Material UI Imports
+import { Box, Typography, Button } from '@mui/material';
+
+const ReusableHomeSection = ({ src, alt, title, description, button, href, label }) => {
```
leekahung: Create a new JSDoc for this new component. I think you could use the `DeleteClientModal.jsx` as a guide to how to write it up. (see https://github.com/codeforpdx/PASS/blob/Development/src/components/Modals/DeleteContactModal.jsx)

src/theme.js
```diff
@@ -28,6 +29,13 @@ const theme = createTheme({
     neutral: {
       main: '#64748B',
       contrastText: '#fff'
+    },
+    Purple: {
```
leekahung: Just `purple` should be fine

leekahung (dismissed):
Think we're getting close to where we want. See feedback.

src/pages/Home.jsx
leekahung: I don't think this has been addressed yet?

src/pages/Home.jsx
leekahung: If `xs` and `sm` are the same, I think we only need to set one of them? Don't think we need 2.

I think the same could be said for the other media queries in this file.

src/pages/Home.jsx
leekahung: This has not been addressed.

src/components/Home/ReusableHomeSection.jsx
```diff
@@ -0,0 +1,55 @@
...
+      component="img"
+      src={src}
+      alt={alt}
+      sx={{
+        width: { xs: '100%', sm: '75%', md: '65%' }
+      }}
+    />
+    <Typography
+      sx={{
+        color: 'Purple.dark',
```
leekahung: Think we could just adjust this to just `purple`?

src/pages/Home.jsx
```diff
@@ -13,26 +20,85 @@ import Paper from '@mui/material/Paper';
...
+              title="Key Features"
+            />
+            <ReusableKeyFeatures
+              icon={
+                <SecurityIcon
+                  sx={{
+                    color: 'Purple.main'
+                  }}
+                />
+              }
```
leekahung: Think this entire block could be made into one line: `icon={<SecurityIcon sx={{ color: 'purple.main' }}/>}`

src/pages/Home.jsx
```diff
@@ -13,26 +20,85 @@ import Paper from '@mui/material/Paper';
...
+              description="Store vital documents like IDs, social security information, birth certificates, medical records, and bank statements in a digital, valid format."
+            />
+            <ReusableKeyFeatures
+              icon={
+                <Diversity1Icon
+                  sx={{
+                    color: 'Purple.main'
+                  }}
+                />
+              }
```
leekahung: This block could also be made into one line

src/pages/Home.jsx
```diff
@@ -13,26 +20,85 @@ import Paper from '@mui/material/Paper';
...
+              description="The platform facilitates smooth communication between non-profit organizations, caseworkers, and the individuals they sesrve. It allows non-profit organizations to maintain a contact list, and casewrokes are assigned contacts whose pods they can access securely."
+            />
+            <ReusableKeyFeatures
+              icon={
+                <SupportIcon
+                  sx={{
+                    color: 'Purple.main'
+                  }}
+                />
+              }
```
leekahung: This block could also be made into one line

src/components/Home/ReusableHomeSection.jsx
leekahung: Just noticed this now, but this should be a round bracket/parenthesis `(` since it's doing an implicit return for the new component.

src/components/Home/ReusableHomeSection.jsx
```diff
@@ -0,0 +1,55 @@
...
+          backgroundColor: 'Purple.light',
+          width: { xs: 1, sm: 1 / 3, md: 1 / 5 },
+          borderRadius: '25px'
+        }}
+      >
+        {button}
+      </Button>
+    )}
+  </>;
+};
```
leekahung: This should be the closing bracket/parenthesis `)` for the new component.

src/components/Home/ReusableHomeSection.jsx
```diff
@@ -0,0 +1,55 @@
...
+          color: 'Purple.main',
+          backgroundColor: 'Purple.light',
+          width: { xs: 1, sm: 1 / 3, md: 1 / 5 },
+          borderRadius: '25px'
+        }}
+      >
+        {button}
+      </Button>
+    )}
+  </>;
```
leekahung: the semi-colon is not needed

src/components/Home/ReusableHomeSection.jsx
leekahung: This should fix the rendering problem that plagues this PR.

src/components/Home/ReusableKeyFeatures.jsx
```diff
@@ -0,0 +1,43 @@
+// React Imports
+import React from 'react';
+// Material UI Imports
+import { Box } from '@mui/system';
+import { Typography } from '@mui/material';
+
+const ReusableKeyFeatures = ({ icon, title, description }) => {
```
leekahung: This shares the same rendering issue as above, but could easily be fixed with the brackets/parentheses mentioned above for `ReusableHomeSection.jsx`

src/components/Home/ReusableHomeSection.jsx
leekahung, line 1: Also, think we could simply call this `HomeSection.jsx`

src/components/Home/ReusableKeyFeatures.jsx
leekahung, line 1: We could simply name this `KeyFeatures.jsx`

src/pages/Home.jsx
leekahung: Think this was what I mentioned with regards to using both `xs` and `sm`. Un-resolving this. I don't think this has been addressed?

src/components/Home/ReusableHomeSection.jsx
xscottxbrownx: Agreed. All React components are inherently reusable.

src/components/Home/ReusableHomeSection.jsx
```diff
@@ -0,0 +1,55 @@
+// React Imports
+import React from 'react';
+// Material UI Imports
+import { Box, Typography, Button } from '@mui/material';
```
xscottxbrownx: Import these individually (like our other components do) to keep the package size down.

src/components/Home/ReusableHomeSection.jsx
```diff
@@ -0,0 +1,55 @@
...
+        textAlign: 'center',
+        marginBottom: '24px'
+      }}
+    >
+      <strong>{title}</strong>
+    </Typography>
+    <Typography
+      sx={{
+        color: 'Purple.main',
+        fontSize: { sm: '1.5rem', md: '1.5rem' },
```
xscottxbrownx: don't think the `md` is necessary. This should render at 1.5rem from `sm` up.

---

Side note:  maybe it's necessary here due to the amount of text, but we haven't had to do font sizes like this yet elsewhere.   The `Typography` component has different variations you can set (like h1-h6, body, etc.)  See [here](https://mui.com/material-ui/react-typography/#component) and [here](https://mui.com/material-ui/react-typography/#changing-the-semantic-element) for more info - if you haven't already.

src/components/Home/ReusableKeyFeatures.jsx
```diff
@@ -0,0 +1,43 @@
+// React Imports
+import React from 'react';
+// Material UI Imports
+import { Box } from '@mui/system';
```
xscottxbrownx: for MUI use default imports vs named imports... their documentation talked about performance issues/package size/etc... 

so these should be more like: `import Box from '@mui/material/Box';` 

see `Home.jsx` for more examples.  It's a little annoying as a developer, instead of just being able to use the named import unfortunately.

src/pages/Home.jsx
```diff
@@ -13,26 +20,85 @@ import Paper from '@mui/material/Paper';
...
+              alt="Web App"
+              title="An App Built for Case Workers"
+              description="Pass allows users to quickly and securely share documents of their clients within the app. The app helps case workes verify and share documents such as ID and proof of income while having total control of the documents."
+            />
+            <ReusableHomeSection
+              src="./src/assets/key-features.png"
+              alt="Key Features"
+              title="Key Features"
+            />
+            <ReusableKeyFeatures
```
xscottxbrownx: since all these icons that are passed to `ReusableKeyFeatures` are `color: 'Purple.main'`
**&**
in `ReusableKeyFeatures` the `Typography`, located in the same `Box` as this icon, is also the same color

Then, we should just set the `Box` in `ReusableKeyFeatures` with `color: 'Purple.main'` and so we don't have to do this each time here (or on the `Typography` within the box either.)

src/components/Home/ReusableKeyFeatures.jsx
```diff
@@ -0,0 +1,43 @@
...
+import { Typography } from '@mui/material';
+
+const ReusableKeyFeatures = ({ icon, title, description }) => {
+  <>
+    <Box
```
xscottxbrownx: See my comment in the `Home.jsx` file of setting color for `Box` here, so we don't have to set it for each icon and this `Typography`.

xscottxbrownx (changes requested):

src/components/Home/ReusableKeyFeatures.jsx
xscottxbrownx: agreed, same reason as above.

src/pages/Home.jsx
xscottxbrownx: See my comment on `Home.jsx` to make this even shorter here.

andycwilliams:
I personally like making purple the main color. However, in my meetings with the UX team, it's become clear that choosing a theme/color palette is actually much more complicated than we might have assumed. As part of our overall plan, finding a new theme is a lower priority. So in my opinion we can hold off on changing that for the time being.

src/components/Home/ReusableHomeSection.jsx
andycwilliams: @Raspber Are you running the prettier script before pushing commits? If not, that should clear up a lot of these minor issues. If available, you could also add prettier as an extension in your code editor. That's what I do.

Raspber:
> I personally like making purple the main color. However, in my meetings with the UX team, it's become clear that choosing a theme/color palette is actually much more complicated than we might have assumed. As part of our overall plan, finding a new theme is a lower priority. So in my opinion we can hold off on changing that for the time being.



> I personally like making purple the main color. However, in my meetings with the UX team, it's become clear that choosing a theme/color palette is actually much more complicated than we might have assumed. As part of our overall plan, finding a new theme is a lower priority. So in my opinion we can hold off on changing that for the time being.

Deleted the purple per request.

src/components/Home/ReusableKeyFeatures.jsx
Raspber: sent a private pm in regards to this. @xscottxbrownx @leekahung

xscottxbrownx:
@Raspber - will you still have access to these graphics after merging into codebase?  

Debating if we should add the purple images in the codebase or not, since we aren't using yet - and not sure if that's what team will land on.

Raspber:
> @Raspber - will you still have access to these graphics after merging into codebase?  
> 
> Debating if we should add the purple images in the codebase or not, since we aren't using yet - and not sure if that's what team will land on. 

Yes, i do have access to these images. I think I can get rid of the purple images because ux team is most likely going to come up with a brand new home page. Ka Hung just wanted something in there so it wasnt so plain. In less the ux team may want to use some of these images.

andycwilliams:
> @Raspber - will you still have access to these graphics after merging into codebase?
> 
> Debating if we should add the purple images in the codebase or not, since we aren't using yet - and not sure if that's what team will land on.

We can talk more about it at the meeting but I think we shouldn't include anything until we know for sure that we need it. And yeah it all depends on the UX team anyhow so I'd wait at least until we talk to them.

xscottxbrownx:
@Raspber at the meetup we agreed, if you can delete the purple image files from this - we can approve and merge.

xscottxbrownx:
We don't need the images in both `public` and `src` folders... 

I did ask for help at last night's meeting to research what the deal is for vite and proper way to use images. So, hopefully we can get a "correct" answer and procedure for us.

ogorman89:
> I did ask for help at last night's meeting to research what the deal is for vite and proper way to use images. So, hopefully we can get a "correct" answer and procedure for us.

@xscottxbrownx The images are loading fine from the `public` folder when I build locally. Was there something else I needed to do to reproduce the issue? 

My understanding of your comment is that we just want to figure out the best practice for storing these assets going forward. 

Edit: referenced incorrect folder.

xscottxbrownx:
> > I did ask for help at last night's meeting to research what the deal is for vite and proper way to use images. So, hopefully we can get a "correct" answer and procedure for us.
> 
> @xscottxbrownx The images are loading fine from the `public` folder when I build locally. Was there something else I needed to do to reproduce the issue?
> 
> My understanding of your comment is that we just want to figure out the best practice for storing these assets going forward.
> 
> Edit: referenced incorrect folder.

Yes photos and actually the `HomeSection` components are loading fine now because we fixed the actual issue which was improper destructuring of props being passed to the component.

But now 2 things...

For this PR
1) photos have been put in both `src` and `public` folders, just needs to be one.  Right now they are being loaded from `src/assets` as you can see in `Home.jsx` - but there was a past PR that Ka Hung ended up adding the `public` folder instead because although it loaded on Development, there were issues on Master.

In General
2) need to research what is proper way to use/store/access images and paths in Vite.  So, we can move forward in future with no issues.   My 5 second research was that `public` folder was one way they could be handled.   Just need to figure it out, and then be consistent from there forward.

leekahung:
Yeah, it seems like Vite doesn't pick up the source files for images under src/assets in the production build (see issue #406). Images end up breaking there even though they show up locally in the development build.

Having the image files stored in the public directory is one way to have static assets built with the site. (See link in issue #406 for Vite documentation and PR #407 on how I implement it here on PASS)

Since that's the case, I agree with Scott. It'll be good to  have them all inside public. Though I wonder if we make folders inside public mainly for images like public/assets or public/images and place it in them instead of the root of public (have not tried it yet in the project).

andycwilliams:
For what it's worth, the professional projects I've seen have had images in the `public` folder, so that seems to be best practice anyhow.

Also, having folders inside of `public` ought to work fine. Though I haven't experimented very much with it.

Otherwise, we're very close to merging this, no?

xscottxbrownx:
> Otherwise, we're very close to merging this, no?

Yes, we just need to:
 1. get rid of the assets from the `src` folder
 2. put in the public path for the image src's in the `Home` component instead of the source path.

xscottxbrownx (changes requested):
See my last comment responding to Andy... two things to address.

xscottxbrownx:
@Raspber why doing this here on line 43, but not on line 53 and 60?

Screenshot 2023-09-23 at 6 04 42 AM

The vite documentation [here](https://vitejs.dev/guide/assets.html#the-public-directory) says it should be `/assets/filename.fileextenstion` (eventhough it seems to work either way.)

--------

@leekahung do you know of anybody still currently working on the driver's license barcode data capture idea?  This PR currently deletes the `barcode demo` from `src/assets` (probably because I wasn't clear enough in my number 2 above in my last comment), and not sure if we should delete that yet.

leekahung:
> @leekahung do you know of anybody still currently working on the driver's license barcode data capture idea?  This PR currently deletes the `barcode demo` from `src/assets` (probably because I wasn't clear enough in my number 2 above in my last comment), and not sure if we should delete that yet.

At the moment, no. Think Brian and Jared were the people working on them before. With regards to the barcode demo file, it's simply there to help developers test out the feature initially since it's saved with the repo. But it's not really necessary, I think.

Raspber:
> @Raspber why doing this here on line 43, but not on line 53 and 60?
> 
> Screenshot 2023-09-23 at 6 04 42 AM
> 
> The vite documentation [here](https://vitejs.dev/guide/assets.html#the-public-directory) says it should be `/assets/filename.fileextenstion` (eventhough it seems to work either way.)
> 
> --------
> 
> @leekahung do you know of anybody still currently working on the driver's license barcode data capture idea?  This PR currently deletes the `barcode demo` from `src/assets` (probably because I wasn't clear enough in my number 2 above in my last comment), and not sure if we should delete that yet.
> 

Did not see that. Thank you. Will get it removed. I can also add the barcode image if youd like.

xscottxbrownx:
@Raspber - all good.  Fixed. Merging now.

Great work on all this!

xscottxbrownx (approved):
All issues addressed

Raspber:
> @Raspber - all good.  Fixed. Merging now.
> 
> Great work on all this!

Thanks for looking out! Its appreciated!

leekahung:
Could you update the PR details with the changes included @Raspber? Thanks!

Issue #420, opened 2023-09-14
Bug: Signup page errors when signing up new pod


leekahung:
### Bug Title: [Bug] Signup page errors when signing up new account

---
### Bug Description:

At the moment, there seems to be an issue when attempting to sign up a new pod. Was not able to reproduce what was shown in PR #378 before it was merged into Development.

---
### Reproduction Steps:    

See clip below.

---
### Expected Behavior:

Allows people to sign-up for new account

---
### Actual Behavior:
What actually happened, and how it differed from the expected behavior.

---
### Screenshots (If applicable):

https://github.com/codeforpdx/PASS/assets/14917816/0d0bd183-74a1-40ed-91e9-49482f02d231

leekahung:
@bingeboy

bingeboy:
I'm getting a similar error after successfully signing up with the form.
Seems to be an issue with `useContactList.js`  

Steps to create the error:
Browser FF Node v16.2

1. Start dev pod and dev server on localhost
2. From homepage http://localhost:5173 click LOGIN button sign into the pod
3. Once the user landing page loads click the link Your Signup Link
4. Create a sign in
5. When the page loads with You have successfully registered for a pod the error should be seen in the console.
6.  XHR requests that are throwing an error 409

Aspireve:
Hey Hi @leekahung 
Could I work on this issue?

Aspireve:
Hey Hi!!
I just cloned your repo and tried solving this bug, however I guess another new bug has developed, I am actually now not being able to open the Login Screen after pressing the Login button on the navbar
Could you suggest what should I do?
This is the error I'm getting

![image](https://github.com/codeforpdx/PASS/assets/93852415/dfae94c6-8d9b-4dc8-810f-794c11bea84a)

leekahung:
Hey @Aspireve. With regards to that error, I think it's due to the local environment not set up correctly. Please check your .env file. If you manage to get the environment set up, you should get the following (assuming you've entered into a valid Solid server).

https://github.com/codeforpdx/PASS/assets/14917816/8007afe4-d992-487e-b7cd-768975de92e3

Aspireve:
Ohh cool!! I was now able to login with a Solid Id.
Could you tell me how to sign up for a new account on the website so I could find the page where the bug resides?

Aspireve:
Ohh I just found it lol.

Aspireve:
Heyy @leekahung 
[...]

PR #431, opened 2023-09-19
Issue 408 refactor signup page into smaller chunks

### LINKED ISSUE #408: Enhancement: Refactor Signup Page into smaller chunks

The existing signup page at `src/pages/Signup.jsx` works well for what we need, but think could be refactored the different components outside of the page and have those imported in instead.

---
See above.

---
### Rationale:
It'll be easier to check/review/optimize/maintain the smaller chunks/components if we break them down.

---
Simply to break off the components built inside `src/pages/Signup.jsx` into their own separate files under `src/components/Signup` or something similar.

bingeboy:
#Issue 408 refactor signup page into smaller chunks

---
## This PR:
Resolves #408 

---
## The files this PR effects:
See [Files Changed](https://github.com/codeforpdx/PASS/pull/431/files)

### Components
Signup.jsx
Signup/Components/PodRegistrationForm.jsx
Signup/Components/ShowNewPod.jsx

### Tests
Added test for the PodRegistrationForm.test.jsx
ShowNewPod.test,jsx seemed to be failing due to #420  

---
Should act the same as what is currently in Development. The goal was to refactor the code.
Please let me know any preferred naming conventions for the components or files.


src/components/Signup/Components/PodRegistrationForm.jsx
```diff
@@ -0,0 +1,106 @@
...
+ * @function
+ *
+ * @params
+ *   @obj
+ *     @param {string} email - value of email input on signup form
+ *     @param {string} password - value of password input on signup form
+ *     @param {string} confrimPassword - value of confirmPassWord on signup form
+ *   @param {string} oidcProvider - TODO add description here 
+ *
+ *   @returns {object} json response
```
bingeboy, line 22: Need to update this to `@returns {promise}`

src/components/Signup/pod/Signup.jsx
```diff
@@ -0,0 +1,152 @@
+// React Imports
```
bingeboy: I need to moved this file to the parent.

src/components/Signup/pod/Signup.jsx
```diff
@@ -0,0 +1,152 @@
...
+  const [showConfirmPassword, setShowConfirmPassword] = useState(false);
+
+  const [searchParams] = useSearchParams();
+  const handleSubmit = async (event) => {
+    event.preventDefault();
+    register(email, password, confirmPassword);
+  };
+
+  return (
+    <>
```
bingeboy: This is a little strange to me `</>`

leekahung: `</>` is shorthand for React Fragment.

xscottxbrownx: Just in case...  

In React/JSX you can only return one component/node/element, so the options is wrapping everything in a `<div>` but this just creates something else in the DOM that actually has no use/need except that React requires that we only return one element.  So, instead of using a `div`, we can use a React Fragment - `<React.Fragment>` or `<>` + `</>` for short.

test/components/Notification/NotificationContainer.test.jsx
```diff
@@ -18,7 +18,6 @@ it('renders all the notifications from notification context', () => {
   ];
 
   render(<MockNotificationContainer notifications={notifications} />);
-  screen.debug();
```
bingeboy, line 21: this was being loud in `stdout` so I removed

src/components/Signup/PodRegistrationForm.jsx
```diff
@@ -0,0 +1,152 @@
...
+import { Visibility, VisibilityOff } from '@mui/icons-material';
+import FilledInput from '@mui/material/FilledInput';
+import IconButton from '@mui/material/IconButton';
+import InputAdornment from '@mui/material/InputAdornment';
+import InputLabel from '@mui/material/InputLabel';
+import Button from '@mui/material/Button';
+import Card from '@mui/material/Card';
+import CardHeader from '@mui/material/CardHeader';
+
+/* Styles for MUI components */
```
andycwilliams, line 17: These can probably be done in line with the MUI components. We don't really have separate styling used elsewhere. Right now much of it is the default. Margin can be included using `sx={{ margin: 8 }}`, for example.

src/components/Signup/PodRegistrationForm.jsx
```diff
@@ -0,0 +1,152 @@
...
+  padding: '8px',
+  margin: '8px',
+  boxShadow: '0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19)'
+};
+
+const formRowStyle = {
+  margin: '20px 0'
+};
+
+const textFieldStyle = {
```
andycwilliams, line 28: `TextField` has the option to use `margin="normal"`, which adds the standard amount of margin. This is what our other text fields currently use.

src/components/Signup/PodRegistrationForm.jsx
```diff
@@ -0,0 +1,152 @@
...
+import InputLabel from '@mui/material/InputLabel';
+import Button from '@mui/material/Button';
+import Card from '@mui/material/Card';
+import CardHeader from '@mui/material/CardHeader';
+
+/* Styles for MUI components */
+const cardStyle = {
+  padding: '8px',
+  margin: '8px',
+  boxShadow: '0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19)'
```
andycwilliams, line 21: I don't think a final design for signup has been decided yet. But right now it looks like two components layered on top of each other, which to me is a bit strange. `Set Up a New Pod` right on top of `Register for PASS` seems like too much. Maybe others will disagree, but I think they can be combined to look more like a single component. This would mean removing the border and box shadow for `Set Up a New Pod`.

bingeboy: @andycwilliams I briefly spoke to @leekahung about this and we sort of thought restyling would be out of scope of this ticket. I simply took what was being used on the Signup page and broke into small digestible code.

src/components/Signup/PodRegistrationForm.jsx
bingeboy: Got it,  I'll make those inline.

src/components/Signup/ShowNewPod.jsx
```diff
@@ -0,0 +1,39 @@
+// React Imports
```
bingeboy, line 1: I created a test file for this component but there is a filed bug #420  which was throwing an error on the test. I'm decided to not address it now.

src/components/Signup/PodRegistrationForm.jsx
leekahung: Yeah, think we could keep this PR dedicated to refactoring. We could get the styling updated in a subsequent/follow-up PR.

bingeboy: @andycwilliams @leekahung I started making a few changes based on feedback and things definitely started to shift around. IMHO we should maybe loop back on the style changes in a new task. Is that cool with you guys?

leekahung: Since I'm away at the moment (didn't bring my computer with me on trip) I could mainly just do comments or replies. Think if the others are okay with it, I'm good.

In this case, think what you suggested will be good.

xscottxbrownx: I'm fine with changing look/styling in another PR. Maybe go ahead and make that issue, but say that this issue needs to be resolved before starting work on that one - just so nobody picks it up and creates merge conflicts?

bingeboy:
@timbot1789 I think this is the test I mentioned on the call last night:
```
//React imports
import React from 'react';
import { render, screen } from '@testing-library/react';
import { BrowserRouter as Router } from 'react-router-dom';

// Testing imports
import { expect, it, describe } from 'vitest';

// Component being tested
import PodRegistrationForm from '../../../src/components/Signup/PodRegistrationForm';

describe('PodRegistrationForm', () => {
  it('Renders the PodRegistrationForm Component', () => {
    render(
        <Router>
          <PodRegistrationForm />
        </Router>
      );

    const textField = screen.getByLabelText('Email');
    const inputElementPassword = screen.getByLabelText('Password');
    const inputElementConfirmPassord = screen.getByLabelText('Confirm Password');

    expect(textField).to.equal('Email');
    expect(inputElementPassword).to.equal('Password');
    expect(inputElementConfirmPassord).to.equal('Confirm Password');
  });
});
```
I think i'm going to be in PDX tonight so maybe we can discuss in person.

andycwilliams:
Were there any updates on this at last night's meeting?

timbot1789:
> Were there any updates on this at last night's meeting?

@andycwilliams I walked JP through the testing framework we're using. He wants to add a few more tests to it, and add one more commit.

bingeboy:
@timbot1789 I added another test for the ShowNewPod component.

test/components/Signup/PodRegistrationForm.test.jsx
```diff
@@ -0,0 +1,28 @@
+// React imports
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { BrowserRouter as Router } from 'react-router-dom';
+
+// Testing imports
+import { expect, it, describe } from 'vitest';
+
+// Component being tested
+import PodRegistrationForm from '../../../src/components/Signup/PodRegistrationForm';
```
timbot1789: Non-issue: you can use `@components/Signup/PodRegistrationForm`. You don't need to use relative pathing.

timbot1789 (approved):
LGTM. Nice Work!

bingeboy:
Squashing and merging once all tests are 🟢

PR #445, opened 2023-10-06
Issue-444 / fix alt typo for images

### LINKED ISSUE #444: [BUG] - typo in alt of HomeSection.jsx
### Bug Description:
The images used in `HomeSection.jsx` should have alt tags for accessibility and seo, but the are misspelled to `lt` so are not working as intended.

---
### Expected Behavior:
Replace typo with `alt` so it will work as intended.

xscottxbrownx:
## This PR:
Resolves #444 
 
**1.**   Replaces `lt` typo with `alt`


bingeboy:
Once `test/pages/__snapshots__/Home.test.jsx.snap` is updated for the test to pass this LGTM 👍

andycwilliams (approved):
The test still needs to pass, but it's such a quick fix I'm okay with approving it now so it can be merged once that's done.

ogorman89 (approved):
Looks good. See @andycwilliams  & @bingeboy notes on the Home.test snapshot.

xscottxbrownx:
Thanks all, I'll need to dig into this - I was under the impression it wasn't good practice to use snapshots when unit testing front-end react code.  I missed this entirely.

timbot1789 (changes requested):
Looks like this is causing a test failure. You'll need to fix that before we can merge this.

timbot1789:
> Thanks all, I'll need to dig into this - I was under the impression it wasn't good practice to use snapshots when unit testing front-end react code. I missed this entirely.

I added some snapshots in older tests when we were first writing the tests, since we used them at my last job a lot. I've come to agree though that snapshots aren't particularly useful. You can remove it if you want, but I'd ask you add some other tests in that show the intended behavior of the component (RTL stuff)

xscottxbrownx:
> Looks like this is causing a test failure. You'll need to fix that before we can merge this.

Got snapshot updated and passing.

timbot1789 (approved):

PR #446, opened 2023-10-08
Make Docker Config usable in development

timbot1789:
# Description
The docker config merged earlier works for the opencommons deploys, but it has specific data that keeps it from being used in development. This PR reconfigures the config to be usable on any computer.

# Testing
1. Pull branch
2. Make sure docker is installed on your computer. Instructions [here](https://docs.docker.com/engine/install/)
3. Run `docker compose up -d`
4. Once that is finished running, navigate to `localhost:3000` and set up your pod.
5. Navigate to `localhost:5173` and check that PASS is working properly.


nginx.conf
```diff
@@ -0,0 +1,31 @@
+events {
+	use epoll;
+	worker_connections 128;
+}
+
+http{
+
+server {
```
bingeboy: Don't full understand why this is here but I don't know much about nginx.conf files.

bingeboy (approved):
Left a minor question on the `nginx.conf` file but I've tested and verified that Solid is running from my docker. Pretty cool 👍

nginx.conf
timbot1789: Whoops I was not supposed to add this file. This is the reverse proxy for my personal server. I'll delete it.

PR #448, opened 2023-10-09
Improve accessibility of labels

milofultz:
## Improve accessibility of labels

Resolves no existing Github issue, only minor fixes. First PR, so let me know if I should be making issues before I make a small PR like this to follow protocol.
 
**1.** Removes unnecessary `aria-label`s.
**2.** Improves existing `aria-label`s.
**3.** Use standard capitalization instead of all caps and rely on styling for capitalization.

### Components

* src/pages/Home
* Modals
* Footer/RenderCallToActionSection.jsx
* NavBar/OidcLoginComponent.jsx

### Tests

* Navbar
* Home snapshot

### Other Files

* src/pages/Contacts.jsx
* src/pages/Profile.jsx

No visual changes.

### DIFF
--- a/src/components/Modals/ConfirmationButton.jsx
+++ b/src/components/Modals/ConfirmationButton.jsx
@@ -28,7 +28,6 @@ const ConfirmationButton = ({ title, confirmFunction, processing }) => (
   <Button
     variant="contained"
     color="primary"
-    aria-label="Confirm Button"
     endIcon={<CheckIcon />}
     onClick={confirmFunction}
     disabled={processing}
--- a/src/components/Modals/ConfirmationButton.jsx
+++ b/src/components/Modals/ConfirmationButton.jsx
@@ -33,7 +33,7 @@ const ConfirmationButton = ({ title, confirmFunction, processing }) => (
     disabled={processing}
     sx={{ marginLeft: '1rem' }}
   >
-    {title.toUpperCase()}
+    {title}
   </Button>
 );
 
--- a/src/components/NavBar/OidcLoginComponent.jsx
+++ b/src/components/NavBar/OidcLoginComponent.jsx
@@ -33,8 +33,7 @@ const OidcLoginComponent = () => {
           }
         }}
         InputProps={{
-          disableUnderline: true,
-          'aria-label': 'OIDC Input Field'
+          disableUnderline: true
         }}
         sx={{
           backgroundColor: 'white',
--- a/src/components/NavBar/OidcLoginComponent.jsx
+++ b/src/components/NavBar/OidcLoginComponent.jsx
@@ -47,7 +47,6 @@ const OidcLoginComponent = () => {
         type="submit"
         color="secondary"
         size="large"
-        aria-label="Login Button"
         onClick={() => {
           loginHandler();
           localStorage.setItem('oidcIssuer', oidcIssuer);
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,54 +1,25 @@
----
 name: Bug report
 about: Create a report to help us improve
 title: ''
 labels: ''
 assignees: ''
 
----
-
-### Bug Title:
-Please provide a CONCISE description of the bug.
-
----
 ### Bug Description:
 Please provide a DETAILED description of the bug.
 
----
 ### Reproduction Steps:    
 **1.**    
 **2.** If needed, delete if not    
 **3.** If needed, delete if not
 
----
 ### Expected Behavior:
 What should have happened if the bug didn't occur.
 
----
-### Actual Behavior:
-What actually happened, and how it differed from the expected behavior.
-
----
 ### Screenshots (If applicable):
 If possible, add screenshots to help explain the problem.
 
----
-### Desktop (please complete the following information):
- - OS: [e.g. iOS]
- - Browser [e.g. chrome, safari]
- - Version [e.g. 22]
-
----
-### Smartphone (please complete the following information):
- - Device: [e.g. iPhone6]
- - OS: [e.g. iOS8.1]
- - Browser [e.g. stock browser, safari]
- - Version [e.g. 22]
-
----
 ### Additional Context (optional):
 Add any other context about the problem here.
 
----
 ### Possible Solution (optional):
 If you have a suggestion on how to fix this bug.
--- a/.github/ISSUE_TEMPLATE/enhancement_request.md
+++ b/.github/ISSUE_TEMPLATE/enhancement_request.md
@@ -1,27 +1,21 @@
----
 name: Enhancement request
 about: Suggest an enhancement for this project
 title: 'Enhancement: [Your Title Here]'
 labels: enhancement
 assignees: ''
 
----
 
 ### Describe the Current Behavior/Feature:
 Provide a clear and concise description of the current behavior or feature.
 
----
 ### Proposed Behavior/Feature:
 A clear and concise description of the enhancement you want. 
 
----
 ### Rationale:
 Please outline why this enhancement is needed and how it will benefit the project.
 
----
 ### Proposed Implementation (if applicable):
 Provide a detailed low level description of how you propose the enhancement should be implemented.
 
----
 ### Additional context:
 Add any other context or screenshots about the enhancement request here.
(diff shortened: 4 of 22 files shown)


src/components/Modals/ConfirmationButton.jsx
```diff
@@ -28,7 +28,6 @@ const ConfirmationButton = ({ title, confirmFunction, processing }) => (
   <Button
     variant="contained"
     color="primary"
-    aria-label="Confirm Button"
```
milofultz, line 31: Removing this allows assistive technology to rely on the button text instead

src/components/NavBar/OidcLoginComponent.jsx
```diff
@@ -33,8 +33,7 @@ const OidcLoginComponent = () => {
           }
         }}
         InputProps={{
-          disableUnderline: true,
-          'aria-label': 'OIDC Input Field'
```
milofultz, line 37: Differs from `label` prop, which should be used instead

xscottxbrownx:
@milofultz I'm still learning here, and wanted to ask you about when a component has multiple buttons how do we test now without aria-labels?  Seems you've been using `getByRole('button')` and curious how do we do this when multiple buttons.

milofultz:
> @milofultz I'm still learning here, and wanted to ask you about when a component has multiple buttons how do we test now without aria-labels? Seems you've been using `getByRole('button')` and curious how do we do this when multiple buttons.

Hey @xscottxbrownx , what I have done in the past has been use other parts of the element that are less prone to change. For instance, maybe the overall structure of the tested component remains the same (e.g. there is always one button), or the accessible name associated with it is fairly static (e.g. a button contains text "submit").

Let's say we have a login field with username and password fields. To get the username field, I could use one of these methods:

* `getByRole('textbox', { name: 'Username' })` (see https://testing-library.com/docs/queries/byrole/)
* `getAllByRole('textbox')[0]`

The former is good because you can use the accessible name, which aids in accessibility and doesn't rely solely on an `aria-label`. YOu can see how it's calculated below the "Terminology" section in this doc: https://www.w3.org/TR/accname-1.2/#mapping_additional_nd_te

I sometimes use the latter if the structure is almost assuredly unchanging, but I should probably use the former in those cases anyway.

Let me know if I can add more clarity here. Thank you!

xscottxbrownx:
> Hey @xscottxbrownx , what I have done in the past has been use other parts of the element that are less prone to change. For instance, maybe the overall structure of the tested component remains the same (e.g. there is always one button), or the accessible name associated with it is fairly static (e.g. a button contains text "submit").
> 
> Let's say we have a login field with username and password fields. To get the username field, I could use one of these methods:
> 
> * `getByRole('textbox', { name: 'Username' })` (see https://testing-library.com/docs/queries/byrole/)

So in your example of a button with the text `submit`, is that what you put as `name`? Like:
`getByRole('button', {name: 'submit'})`

---

And in your `3` in the PR description...   are you saying that the screen readers/assistive technology will read what we have in our codebase vs the styling?  
Example:
you changed `CANCEL` button text to `Cancel` and said if we wanted to all-caps it, to do so through CSS basically. 
Is this just a standard/best-practice, or having all-caps in the codebase causes issues?

xscottxbrownx:
This all looks good. I'm just asking questions to better my understanding and so when we write more tests (tons of components are missing test), I don't need to rely on checking for `aria-label` of elements.

bingeboy:
Is there an accessibility plugin you recommend to verify these types of changes? I've look at the code and the changes seem good to me but I'm not an accessibility expert. I did a little research and read a few posts about why these `aria-labels` shouldn't be needed. Very informative. 👍 

I use to use Wave but it's been a few years :)

andycwilliams:
> Is there an accessibility plugin you recommend to verify these types of changes? I've look at the code and the changes seem good to me but I'm not an accessibility expert. I did a little research and read a few posts about why these `aria-labels` shouldn't be needed. Very informative. 👍

I'm not familiar with any but I'd be surprised if there weren't any.

Could you link to these posts please?

bingeboy:
@andycwilliams https://wave.webaim.org/ and https://www.freecodecamp.org/news/web-accessibility-common-aria-mistakes-to-avoid/

milofultz:
> So in your example of a button with the text `submit`, is that what you put as `name`? Like: `getByRole('button', {name: 'submit'})`
> 
> And in your `3` in the PR description... are you saying that the screen readers/assistive technology will read what we have in our codebase vs the styling? Example: you changed `CANCEL` button text to `Cancel` and said if we wanted to all-caps it, to do so through CSS basically. Is this just a standard/best-practice, or having all-caps in the codebase causes issues?

Yeah, the example you would use `submit` or any [text match object](https://testing-library.com/docs/queries/about#textmatch), so regex would work, too, like `getByRole('button', {name: /submit/i})`, if you wanted to make sure it matched without case sensitivity.

For the capitalization, there isn't any WCAG guidance on it, but generally, screen readers can interpret all capital series of letters as acronyms and CSS is essentially always ignored. So "SUBMIT" in the HTML could be read as "ESS EWE BEE EM EYE TEE" instead of "submit". Using `<...>Submit</...>` with `text-transform: uppercase` (which I think MUI is already doing by default) will result in a better user experience.

milofultz:
> Is there an accessibility plugin you recommend to verify these types of changes? I've look at the code and the changes seem good to me but I'm not an accessibility expert. I did a little research and read a few posts about why these `aria-labels` shouldn't be needed. Very informative. 👍
> 
> I use to use Wave but it's been a few years :)

The easiest tool to set up and use for the most basic stuff (certainly not all accessibility issues) is [Microsoft Accessibility Insights](https://accessibilityinsights.io/). It's Chrome/Edge only, unfortunately, but it is very good at quick passes for the super low-hanging fruit.

For _verification_ and other more nuanced accessibility stuff, there are essentially no tools that can fully verify anything due to a variety of factors. If any tools say they can, they are lying (see [accessibility overlays](https://adrianroselli.com/2022/03/overlays-underwhelm-a11y-nyc.html)). Things like Deque's axe (which can be integrated into Jest and [Vitest](https://github.com/chaance/vitest-axe), but I haven't tested it yet) are a great way to stop common regressions and find slightly more complex issues. But at the end of the day, it's just becoming familiar with the dialect of assistive technologies and their quirks.

milofultz:
In general, re: when `aria-label` is/isn't needed, check out [how accessible names are resolved](https://w3c.github.io/accname/#mapping_additional_nd_te) and use the [Chrome](https://developer.chrome.com/docs/devtools/accessibility/reference/#pane)/[Firefox](https://www.a11yproject.com/posts/using-firefox-for-accessibility-testing/) Devtools's Accessibility pane to see how it's calculated.

Firefox example of the "Update branch" button's accessible name:

Screen Shot 2023-10-12 at 8 53 48 AM

Chrome example of the "Sign up" button (note that it's actually a _link_:

Screen Shot 2023-10-12 at 8 57 08 AM

Chrome's gives a little more info on how it calculates, which is great.

bingeboy (approved):
LGTM 👍 thanks for all the details

xscottxbrownx (changes requested):
If taking out the label in `Home.jsx`, should adjust the component it was being passed to `HomeSection.jsx`.

---

And maybe switch the `{title.toUppercase()}` to `{title}` in `ConfirmationButton.jsx`, as it does seem you are correct - MUI seems to have default uppercase styling on buttons within modals.

milofultz:
> If taking out the label in `Home.jsx`, should adjust the component it was being passed to `HomeSection.jsx`.
> 
> And maybe switch the `{title.toUppercase()}` to `{title}` in `ConfirmationButton.jsx`, as it does seem you are correct - MUI seems to have default uppercase styling on buttons within modals.

Got them in there. Thanks!

src/components/NavBar/OidcLoginComponent.jsx
xscottxbrownx: you said `label` prop should be used?

xscottxbrownx (commented):
Just question on `OidcLoginComponent`, then can approve.

src/components/NavBar/OidcLoginComponent.jsx
milofultz: Yeah, the [`label` that is a top-level prop](https://mui.com/material-ui/api/text-field/#TextField-prop-label) (in this case, "Pod Server URL"), not an object property passed into [`InputProps`](https://mui.com/material-ui/api/text-field/#TextField-prop-InputProps).

xscottxbrownx (approved):

PR #450, opened 2023-10-09
Add `lang` attribute to html root

milofultz:
https://www.w3.org/WAI/standards-guidelines/act/rules/b5c3f8/
 
Adding a `lang `attribute to our root HTML page helps assistive technologies better understand the language that the document is in.


leekahung (approved):
That's fine. Can approve this. 👍

PR #451, opened 2023-10-10
#413 Convert the "To" field in New Message Modal to select contacts

### LINKED ISSUE #413: Enhancement: Convert the "To" field in New Message Modal to select contacts
Currently, if you want to send a message, you need to fill the "To" field with the whole web ID of whoever you want to message. This is a multipart URL (like `https://timstanden.solidcommunity.net`). It needs to be typed exactly, and is very difficult for users to remember.

image

Instead, we could make the "To" field select from the user's contacts list. If a person is in your contacts list, you simply type in their name, and we autopopulate the field with the proper URL, like gmail does with a person's name and email:
image

Implementing this will make it significantly easier to send messages in PASS

cnease10:
## Change message recipient input to dropdown

---
## This PR:
Resolves #413
 
**1.**    The message recipient input was difficult for users to type the full webId, this simplifies messaging a contact by using your contacts' names 

---
### Components
src/components/Modals/NewMessageModal.jsx

### Tests
test/components/Modals/NewMessageModal.test.jsx

---
image

---
Using react autocomplete to present a dropdown of user contacts. Selecting a contact should display the name and set the recipient URL to the appropriate pod URL.


leekahung:
If you want to test out the feature with two separate Pods, think you could either make another pod in your local Solid server at localhost or make another Pod on one of the test servers like solidcommunity.net.

Either would be a good choice. To mimic two separate users, you could login to one pod in your working window in the browser and a second pod in a private window.

DionSat:
Okay, so I tried 2 different users by using 2 pods in different browsers and it seems to work. I was able to receive all the messages that I sent. There was an issue where they weren't appearing when using the refresh button for the inbox. But when I refreshed the entire page it seemed to show up.

leekahung:
> Okay, so I tried 2 different users by using 2 pods in different browsers and it seems to work. I was able to receive all the messages that I sent. There was an issue where they weren't appearing when using the refresh button for the inbox. But when I refreshed the entire page it seemed to show up. 

It might be related to one of the states not updating when rendering. We can try checking the side effects with a console log.

If it ends up being that case, we could try to utilize the useEffect hook with a dependency array with the state.

leekahung:
~~Another option would be to utilize the list from the custom useContactList hook itself and just filter it to get the specific contact and have that show up in the rendering~~

Edit: Never mind, I see that you're already trying that in the new commits.

DionSat:
It was on the Firefox browser that it happened, so at first I thought it might be a browser issue. Chrome was able to use the refresh button and show the new messages. But now after refreshing the Firefox page it seems to be working. At least its not doing it anymore. Not sure if I can recreate it.

src/components/Modals/NewMessageModal.jsx
```diff
@@ -136,18 +143,26 @@ const NewMessageModal = ({ showModal, setShowModal, oldMessage = '', toField = '
...
+                  recipientPodUrl: newValue.id ?? newValue
+                });
+              }}
+              fullWidth
               required
               autoFocus
-              label="To"
-              fullWidth
               disabled={toField !== ''}
+              renderInput={(params) => <TextField {...params} label="To*" />}
```
leekahung: Quick question, what's with the additional `*` for the `label`? Wouldn't `To` suffice?

DionSat: I'm actually not sure, what * is for. It shouldn't really be necessary. I'll mess around later to see and update it if it isn't.

DionSat:
Okay, it works without the asterisk so its not necessary. I also changed the tests to find the input from the autocomplete Id provided by Cierra, because there might be more than one autocomplete combobox role for the message in the future. Also Merry Christmas to you all.

leekahung (approved):

Issue #414, opened 2023-09-13
Enhancement: Add "Message" action to contacts list


timbot1789:
Currently, the contacts list only has a few actions: go to pod, pin contact, and delete

![image](https://github.com/codeforpdx/PASS/assets/37914436/e8471315-ae28-4bd8-aeb9-97c36cd2960c)

Another useful action to have on this table would be to message the client. Add a button to each contact in this list that launches  the "send message" modal when clicked, that prepopulates the "to" field with the webId of the contact.

andycwilliams:
Since I'm already working on #343 I can just pick this one up too. I think it'd be easier that way.

andycwilliams:
So I added the modal and now am setting it up to populate the `To:` field. What is the most efficient way to do this? Would it be to utilize the `oldMessage` parameter (I'm guessing via `message.recipientPodUrl` in the new message modal)?

timbot1789:
You should be able to pull the value of the `to` field from the webId of the contact, in the contactsList. Take a look at `src/hooks/useContactsList.js`.

PR #457, opened 2023-10-14
[Enhancement] - Replace existing Contact List with MUI Data Grid

### LINKED ISSUE #343: Enhancement: Update Contacts View to display Contact Cards instead of Table
Recent updates have compressed the amount of information displayed in the client's table. It now looks like this:
image

Not very much useful information displayed there.

Our figma mockups also display something else here. Instead of a table, we have a list of cards for each client. We can then click on the client to go to their profile.

We already have a similar card displayed on the profile page:
image

We could use these cards in place of the table rows to display client information

leekahung:
## This PR:

Updates existing table used for our Contact List with MUI Data Grid. The module for MUI Data Grid has been included to package.json. The column with originally with the link to the contact profile has been split up into the original text and a button icon in order to allow for filtering via MUI Data Grid. Other than that, the grid should function similarly to our previous table.

The unused `ContactListTableRow.jsx` is removed and a new component `ContactProfileIcon.jsx` is created for the React Router Link to the contact profile page.

The PR also includes a minor correction for `contactProfile.profileImage` in `src/components/Profile/ProfileImageField.jsx`. Before the profileImage would be incorrectly rendered because of the incorrect variable name, `profileImg`, which doesn't exist as part of `contactProfile`.

The column with pinning has been removed (UX team believes if there's already a filter + sorting capability with MUI Data Grid, this might not be necessary; it'll be better to have this moved as a type of status instead). The name from Person has been split up into "First Name" and "Last Name" for a better sorting experience.

---
### Components
- `src/components/Contacts/ContactListTable.jsx`
- `src/components/Contacts/ContactListTableRow.jsx` (file removed)
- `src/components/Contacts/ContactProfileIcon.jsx` (file added)
- `src/components/Profile/ProfileImageField.jsx`

### Test Files
- `test/components/Contacts/ContactsListTable.test.jsx`
- `test/components/Profile/ProfileImageField.test.jsx`
- `test/pages/Contacts.test.jsx`

### Other Files
- `src/components/Contacts/index.js`
- `package.json`
- `package-lock.json`

---

https://github.com/codeforpdx/PASS/assets/14917816/76a2e574-7e2f-488c-8308-866ffff8d7ae

---
## Future Considerations:
- Implement accessibility improvements with MUI Data Grid after merger


test/components/Contacts/ContactsListTable.test.jsx
```diff
@@ -25,11 +25,11 @@ it('renders all clients from client context', () => {
...
 
-  const row1 = allRows[1];
-  const row2 = allRows[2];
+  const row1 = allRows[1].querySelector("[title='Aaron Abby']");
+  const row2 = allRows[2].querySelector("[title='Bob Builder']");
 
-  expect(row1.cells.item(0).innerHTML).toContain('Aaron Abby');
-  expect(row2.cells.item(0).innerHTML).toContain('Bob Builder');
+  expect(row1.innerHTML).toBe('Aaron Abby');
+  expect(row2.innerHTML).toBe('Bob Builder');
```
milofultz: For this query, you could use `queryByRole` to avoid the `querySelector` call. (see [this and following section in Kent C. Dodds page on avoiding `querySelector`](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-container-to-query-for-elements))

```tsx
  const row1 = queryByRole('cell', { name: 'Aaron Abby' });
  const row2 = queryByRole('cell', { name: 'Bob Builder' });

  expect(row1).not.toBeNull();
  expect(row2).not.toBeNull();
```

src/components/Contacts/ContactProfileIcon.jsx
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <Link
+      to={`/profile/${encodeURIComponent(contact.id)}`}
+      state={{ contact: contact.value }}
+      style={{ textDecoration: 'none', color: theme.palette.primary.dark }}
+    >
+      <IconButton onClick={() => handleSelectProfile(contact.value)}>
+        <ContactPageIcon />
+      </IconButton>
+    </Link>
```
milofultz: This will render a `button` inside of an `a` tag. Interactive buttons shouldn't be nested because usually one of the two will be ignored. 

From an accessibility perspective, generally [`a` tags](https://www.w3.org/TR/wai-aria-1.2/#link) should be used to navigate and [buttons](https://www.w3.org/TR/wai-aria-1.2/#button) should be used for an action. In this case, I think you could add the `onClick` to the `Link` element without any adverse effects and improve the user experience.

src/components/Contacts/ContactListTable.jsx
```diff
@@ -50,31 +54,68 @@ const ContactListTable = ({ contacts, deleteContact }) => {
...
-                contact={contact}
-                deleteContact={deleteContact}
-              />
-            ))}
-          </TableBody>
-        </Table>
-      </TableContainer>
-    </ThemeProvider>
+    <Box sx={{ margin: '20px 0' }}>
+      <DataGrid
```
milofultz, line 82: [There are a lot of accessibility issues with this component](https://github.com/mui/mui-x/issues?q=is%3Aopen+label%3A%22component%3A+data+grid%22+label%3Aaccessibility+), but I do think that a `grid` role is the right one so don't want to stand in the way of progress on this point. I tried it with VoiceOver and it wasn't really doing what I was expecting, but don't know if that's our or their side.

milofultz (changes requested):
If we could resolve the button nested in an `a` tag, that would be great. Thank you!

src/components/Contacts/ContactProfileIcon.jsx
leekahung: This is a React Router Link that helps with sending a specific state (i.e. the `contact` object) along with the redirect and browser URL. A simple anchor tag in this case would not do.

src/components/Contacts/ContactListTable.jsx
leekahung: Think having a fix for those would be great. We could open up an issue to fix the accessibility problems here in a follow-up. 👍

test/components/Contacts/ContactsListTable.test.jsx
leekahung: I'll give this a try.

leekahung:
> If we could resolve the button nested in an `a` tag, that would be great. Thank you!

Thanks for the review @milofultz! I'll have the changes made in a bit.

src/components/Contacts/ContactProfileIcon.jsx
leekahung: Or do you mean to just move the onClick to Link and just leave the Icon wrapped without the IconButton? I could try that.

andycwilliams:
Is there an associated issue for this?

@milofultz This is valuable feedback, at least for my own understanding. More insight into accessibility is really helpful for the project.

leekahung:
> Is there an associated issue for this? 

I was looking around the issues. There doesn't seem to be one for this or the ones that were there was closed.

test/components/Contacts/ContactsListTable.test.jsx
leekahung: Got this updated in fix.

src/components/Contacts/ContactProfileIcon.jsx
leekahung: Got this changed to:

```jsx
    <Link
      to={`/profile/${encodeURIComponent(contact.id)}`}
      state={{ contact: contact.value }}
      onClick={() => handleSelectProfile(contact.value)}
    >
      <ContactPageIcon sx={{ color: 'gray' }} titleAccess="contact profile link" />
    </Link>
```

timbot1789:
I think the data grid is a good enhancement over the current view. I especially like how we can easily add sorting to it (which may eliminate the need for pinning / favoriting). 

How would you say it relates to [this ticket](https://github.com/codeforpdx/PASS/issues/343), where I talk about converting the contacts list into a series of cards?
I based that off of [one of the figma designs](https://www.figma.com/file/BfsGjGbFUVvtqNq18sykoY/Pass-Project?type=design&node-id=1216-2986&mode=design) UX submitted. 
image

leekahung:
> How would you say it relates to [this ticket](https://github.com/codeforpdx/PASS/issues/343), where I talk about converting the contacts list into a series of cards? I based that off of [one of the figma designs](https://www.figma.com/file/BfsGjGbFUVvtqNq18sykoY/Pass-Project?type=design&node-id=1216-2986&mode=design) UX submitted.

I think we could certainly have this link this to that issue since it does update the UI.

If we want to, think we could also separate out different fields from `contact` (like first name, last name) in a follow-up PR to make better use of Data Grid's innate sorting capabilities instead of our existing sorting scheme.

andycwilliams:
> > How would you say it relates to [this ticket](https://github.com/codeforpdx/PASS/issues/343), where I talk about converting the contacts list into a series of cards? I based that off of [one of the figma designs](https://www.figma.com/file/BfsGjGbFUVvtqNq18sykoY/Pass-Project?type=design&node-id=1216-2986&mode=design) UX submitted.
> 
> I think we could certainly have this link this to that issue since it does update the UI.
> 
> If we want to, think we could also separate out different fields from `contact` (like first name, last name) in a follow-up PR to make better use of Data Grid's innate sorting capabilities instead of our existing sorting scheme.

When I last talked to UX about this, they said they were still working on the desktop version of these cards. So that combined with temporarily focusing on documentation recently, I haven't done a ton with #343 lately (and should really have moved that to the backlog).

I'm still willing to complete that, it would just mean waiting until this is merged before continuing.

Also, I would likely focus on #414 before moving onto #343.

timbot1789:
> When I last talked to UX about this, they said they were still working on the desktop version of these cards. So that combined with temporarily focusing on documentation recently, I haven't done a ton with #343 lately (and should really have moved that to the backlog).
> 
> I'm still willing to complete that, it would just mean waiting until this is merged before continuing.
> 
> Also, I would likely focus on #414 before moving onto #343.

That sounds good to me. I personally like the card look more than the table. IMO it's more humanizing, and shows the data in a more familiar way. And I think it will work well in a desktop environment. The table view is good when you have several rows worth of data you need to look at all at once, but I don't know how often that will be the case here.

src/components/Contacts/ContactListTable.jsx
```diff
@@ -50,31 +54,68 @@ const ContactListTable = ({ contacts, deleteContact }) => {
   const contactsCopy = [...contacts];
   const sortedContacts = contactsCopy.sort(comparePerson);
 
+  const columnTitlesArray = [
+    {
+      field: 'Contact',
+      width: 160,
+      headerAlign: 'center',
+      align: 'center',
+      sortable: false
```
xscottxbrownx: should the contact name not be sortable?

xscottxbrownx: I assume this is the arrow that shows if you hover and makes sorted by ascending or descending values...

xscottxbrownx: If so, personally I would think this to be one of the most used features of the grid honestly.

package.json
```diff
@@ -34,6 +34,7 @@
     "@mui/material": "^5.12.2",
     "@mui/styled-engine-sc": "^5.12.0",
     "@mui/system": "^5.12.1",
+    "@mui/x-data-grid": "^6.16.2",
```
andycwilliams, line 37: Isn't this data grid what we used months back when creating the original table and implementing MUI? I don't recall why we decided not to go with it at that time.

src/components/Contacts/ContactProfileIcon.jsx
milofultz: Yeah, that's what I was thinking. Thank you!

milofultz:
> > Is there an associated issue for this?
> 
> I was looking around the issues. There doesn't seem to be one for this or the ones that were there was closed.

@andycwilliams @leekahung The issues I had when using VoiceOver on the data grid (which I'll admit isn't technically super proficient) were just basic navigation. In comparison to [a data grid like this one from W3](https://www.w3.org/WAI/ARIA/apg/patterns/grid/examples/data-grids/#aboutthisexample) which was super easy to navigate and interact with, the MUI one on their site was almost unusable. 

I don't quite have to technical know how with grid yet to understand why this is the case, but I saw similar complaints in their backlog (https://github.com/mui/mui-x/issues/4749, https://github.com/mui/mui-x/issues/4658, https://github.com/mui/mui-x/issues/4282).

Again, don't know how much of this is them, us, or me, but I saw a little bit of signal confirming my suspicions in their backlog so thought I'd bring it up.

edit: oop misread the original post, but leaving this here for clarity anyway, sorry for the erroneous tag now :)

src/components/Contacts/ContactListTable.jsx
leekahung: I was thinking about sorting with what we have now. Was thinking we could perhaps split up `person` back into just `givenName` and `familyName`. If we do that, think having the sorting then would be better. See clip:

https://github.com/codeforpdx/PASS/assets/14917816/c749ec9e-218e-4895-9d95-eea30d59d8a3

It'll open up more fields for filtering as well.

https://github.com/codeforpdx/PASS/assets/14917816/e1df46fd-f968-4d14-942f-a180a19925e6

The changes done in those samples are relatively simple.

package.json
leekahung: Yeah, it was the same one as before. I think at the time we didn't went with it was because there were still a few thing we wish to discuss, although I don't remember which things.

src/components/Contacts/ContactProfileIcon.jsx
```diff
@@ -0,0 +1,59 @@
...
+      setContact(contactInfo);
+    } catch {
+      setContact(null);
+      navigate('/contacts');
+      addNotification('error', 'WebId does not exist');
+    }
+  };
+
+  return (
+    <Link
```
milofultz, line 49: thanks again! 🙏 🙇

test/components/Contacts/ContactsListTable.test.jsx
```diff
@@ -14,34 +14,59 @@ const MockTableComponent = ({ contacts }) => (
...
+      person: 'Bob Builder',
+      webId: 'https://example.com/Builder'
+    }
   ];
 
   const { getByText } = render(<MockTableComponent contacts={[...originalArray]} />);
 
-  const client1 = getByText('Aaron Zeigler');
-  const client2 = getByText('Bob Builder');
+  const client1 = getByText('Zeigler');
```
milofultz: could use the same `getByRole` schema here as above

milofultz (approved):
Just one small thing, but not a blocker

test/components/Contacts/ContactsListTable.test.jsx
leekahung: No problem! I'll have this switched up before merging.

leekahung:
Got everything updated from the suggestions. Let me know if there's still anything that needs to be reviewed for this PR.

test/components/Contacts/ContactsListTable.test.jsx
leekahung: Got this updated to use `getByRole` 👍

src/components/Contacts/ContactListTable.jsx
leekahung: Got Person split up into First Name and Last Name and allow sorting of those fields. The clip for the update is now shown in the PR description. Let me know if you think this is good.

leekahung:
Hey @xscottxbrownx @andycwilliams. Think I've manage to touch on the things you guys mentioned. Let me know if there's still other things you guys need to review or it's good to go?

src/components/Contacts/ContactListTable.jsx
```diff
@@ -44,31 +50,74 @@ const ContactListTable = ({ contacts, deleteContact }) => {
   const contactsCopy = [...contacts];
   const sortedContacts = contactsCopy.sort(comparePerson);
```
andycwilliams: With the new data grid, is a custom sorting function still necessary? Unless I'm mistaken, it should be possible to use `initialState` in `DataGrid` for the same results.

src/components/Contacts/ContactListTable.jsx
```diff
@@ -44,31 +50,74 @@ const ContactListTable = ({ contacts, deleteContact }) => {
...
 
+  const columnTitlesArray = [
+    {
+      field: 'First Name',
+      width: 120,
```
andycwilliams: Could we possibly expand the table a bit when in desktop views? The quickest way would be by widening the columns.

I know there isn't a ton of data to display at the moment, but it'd help to use up some of all the empty space. Plus, in case anyone has a longer name, it'd prevent that from being cut off.

src/components/Contacts/ContactListTable.jsx
```diff
@@ -44,31 +50,74 @@ const ContactListTable = ({ contacts, deleteContact }) => {
...
+      filterable: false,
+      width: 70,
+      headerAlign: 'center',
+      align: 'center'
+    },
+    {
+      field: 'actions',
+      type: 'actions',
+      width: 70,
+      getActions: (contactData) => [
```
andycwilliams, line 70: Although it's not essential, I think it'd be nicer to have a title for the delete column. If only because it's a bit odd to have one nameless column. The same kind of column in the documents is named, so this'd make it more consistent.

Adding `headerName: 'Delete',` here would accomplish this.

andycwilliams (commented):
Overall, it looks really nice. It's strange seeing the data grid again after all these months.

I think it would have been good to first create an issue before opening a PR as we normally do.

Curious, is the grid also intended to be implemented with the documents table?

leekahung:
> Curious, is the grid also intended to be implemented with the documents table?

Yeah, I think the grid would work well for the document list as well. Think we could have an issue created for that.

src/components/Contacts/ContactListTable.jsx
leekahung: We could certainly drop the custom sort and leave it as an initial state.

src/components/Contacts/ContactListTable.jsx
leekahung: I'll add one. I'll expand the width of this too.

src/components/Contacts/ContactListTable.jsx
leekahung: Would like to keep the minimum width of the overall grid to 300px for mobile, but I'll play around with the column widths.

src/components/Contacts/ContactListTable.jsx
leekahung: Got this included.

src/components/Contacts/ContactListTable.jsx
leekahung: Made text columns flexible. If we get more columns included in between them in the future, we could make them flexible as well.

https://github.com/codeforpdx/PASS/assets/14917816/1ab60850-c076-4f2b-95d3-7877a2309e0b

src/components/Contacts/ContactListTable.jsx
leekahung: Got the custom sort dropped for sorting from `initialState`

xscottxbrownx:
Issue with rendering size on skinny mobile devices (like Galaxy Fold and similar), but not too worried about for MVP.

leekahung:
> Issue with rendering size on skinny mobile devices (like Galaxy Fold and similar), but not too worried about for MVP.

Yeah, I believe those screen sizes are at around 280px width? At an earlier discussion, I believe we've talked about limiting the lower screen limit to about 300px. Think I could adjust it to make it fit the smaller size using `vw` for the width size on the box surrounding the grid.

https://github.com/codeforpdx/PASS/assets/14917816/7ccaa6fb-1468-4e81-b17b-7301bc07353b

andycwilliams (commented):
Looking better. There's a plan to add at least one more column (an icon button to allow for direct messaging), so that will help fill out the table.

I'm seeing some large gaps between the navbar and table (see screenshot) when using larger viewports. Looks to be caused by `justifyContent: 'center'` on line 58 in `Contacts.jsx`

Capture

Is this the desired effect? It might be nicer to have it consistent on all viewports and make the table always close to the navbar. Just want to be sure.

I'd really like to hear from case workers about how small the screens they actually use are. It may not even be an issue to set a 300px limit. But it may take time to get feedback.

leekahung:
> I'm seeing some large gaps between the navbar and table (see screenshot) when using larger viewports. Looks to be caused by `justifyContent: 'center'` on line 58 in `Contacts.jsx`

I could certainly pull the Data Grid back up closer the the Navbar without the `justifyContent`. I'll also make the grid use a set height to fill out the the spacing in the "Contacts" page. It would not affect the mobile view.

https://github.com/codeforpdx/PASS/assets/14917816/82e3a4d7-ebd7-4be2-ab09-c50660de02ca

> I'd really like to hear from case workers about how small the screens they actually use are. It may not even be an issue to set a 300px limit. But it may take time to get feedback.

With regards to the mobile width limit, what we're making here should be quite flexible down to 280px and lower, so I don't think we need to worry about that point now. (As you can see below, the grid content stretches out to 380px, but the grid itself can remain small (around ~210px in the example clip) thanks to the horizontal scrolling implemented with the grid.)

https://github.com/codeforpdx/PASS/assets/14917816/f6f04cfc-fb9d-46b5-bb66-75efae38edea

leekahung:
I believe the changes here are now adequate?

PR #459, opened 2023-10-16
improve react query behavior

timbot1789:
A small improvement to react query behavior. React query automatically refetches any data marked as "stale" when it's requested. By default `staleTime` is set to zero, everything is automatically stale and always refetched. This change sets `staleTime` to one minute. Documents won't be refetched until they're a minute old. This should reduce some redundant fetches.

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -22,8 +22,14 @@ import Layout from './layouts/Layout';
 /**
  * @typedef {import("./typedefs").messageListObject} messageListObject
  */
-
-const queryClient = new QueryClient();
+const queryClientConfig = {
+  defaultOptions: {
+    queries: {
+      staleTime: 1000 * 60 // one minute
+    }
+  }
+};
+const queryClient = new QueryClient(queryClientConfig);
 
 const App = () => (
   <QueryClientProvider client={queryClient}>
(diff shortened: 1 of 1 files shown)


xscottxbrownx:
sounds good to me.

not sure if leaving a space between the jsdoc typedef and the additions is needed or not - so as to not link the jsdoc with the additions to codebase?

leekahung:
> sounds good to me.
> 
> not sure if leaving a space between the jsdoc typedef and the additions is needed or not - so as to not link the jsdoc with the additions to codebase?

If they're imported typedefs, they don't to be directly above the lines of code it's being used for like what we do for components which JSDoc needs to be directly above them to be associated.

leekahung (changes requested):
I've just noticed that the two imported typedefs aren't being used in `App.jsx` at all. They're likely remnants from an earlier time in the project, but I think we could actually remove them here since they're not longer in use.

timbot1789:
Fixed JSDoc typedefs. @leekahung can you add a work item for yourself to add jsdoc validation into the project as part of the linter? Checking this stuff manually is pretty inefficient.

leekahung:
> Fixed JSDoc typedefs. @leekahung can you add a work item for yourself to add jsdoc validation into the project as part of the linter? Checking this stuff manually is pretty inefficient.

The plugin for the linter for JSDoc is already in the project. I'll see if I could get the specific validation call turned on for it for imports like this.

leekahung (approved):
Looks good for merging.

PR #469, opened 2023-10-19
Limit size of file upload for profile photo

### LINKED ISSUE #289: Limiting size of file uploaded onto PASS for profile images
At the moment, users can simply upload a file of any size for their profile picture, given that it's an accept file for upload (i.e. ".png", ".jpeg", ".gif", etc.).

Perhaps we could introduce a size limit to file upload.

ogorman89:
## This PR:
Resolves #289 
 
**1.**  Limits profile photo file size to 1MB
**2.** Shows an error in bottom left if the photo size is greater than 1MB

https://github.com/codeforpdx/PASS/assets/3335277/69b8a420-5658-4e60-b257-8c8889acbdcf

Just a limit for the small avatar image, so that it is no longer unconstrained. 

We may want to revisit and adjust the profile photo file size requirement during user testing.

None for this PR, but there are a couple of items mentioned in issue #289 worth future discussions:

**1.** See @timbot1789 notes on limiting document uploads with an environment variable configurable by the hosting org.
**2.** Similarly, we might consider a configurable total pod size constraint so that the total storage size of a pod is limited so that dimension is not unconstrained. This is more of a concern for the pod hosting organization whoever that ends up being.

### DIFF
--- a/src/components/Profile/ProfileImageField.jsx
+++ b/src/components/Profile/ProfileImageField.jsx
@@ -11,6 +11,7 @@ import ImageIcon from '@mui/icons-material/Image';
 import Typography from '@mui/material/Typography';
 // Contexts Imports
 import { SignedInUserContext } from '@contexts';
+import useNotification from '../../hooks/useNotification';
 
 /**
  * @typedef {import("../../typedefs").profileImageFieldProps} profileImageFieldProps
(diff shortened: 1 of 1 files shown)


leekahung (approved):
The changes here looks good to me.

andycwilliams:
I agree with the future steps needed but we need more user testing and that may not be for a while. So this is good to go and gives us a good basis to work with.

PR #470, opened 2023-10-20
412 enhancement add a login modal to select pod provider

### LINKED ISSUE #412: Enhancement: Add a login modal to select pod provider
Currently, when you log into PASS, you are required to add the pod provider url to provider field. This field is prepopulated from the `.env` file, however, if you want to change to a different provider, you need to put the full URL in manually, which can be annoying. It's also confusing for newcomers, who don't know what URL to put in that form.

image

Instead, we can still prepopulate the field with the url from the `.env` file, but then provide a list of common options if they want to pick something else. 

### Possible designs
A dropdown like podpro.dev would be good, clean way to do this
image

We could also use a large modal like solidcommunity.net, but that's a bit heavier
image

rioredwards:
## This PR:
Resolves #412
 
**1.**  Replaces basic text field for one with autocomplete (a dropdown appears with recommended options) in the OidcLoginComponent.
**2.** Adds SUGGESTED_OIDC_OPTIONS constant; a set of recommended pod servers. 

## Screenshots:
PASS_Autocorrect_Feature_Preview

## Future Steps/PRs Needed to Finish This Work:
- The SUGGESTED_OIDC_OPTIONS has placeholder options in it, for now, that should be replaced.

- The UI for the component was based on the design of [Pod Pro's design](https://podpro.dev/login). The one I made ended up being more heavy-duty than theirs because I was using MUI's [Auto Complete](https://mui.com/material-ui/react-autocomplete/) component, which is more bulky by default. We could slim it down if needed. I started on a version that looked more like theirs, here was the result:

Pod Pro's Design:
Pod Pro's Component

My Attempt at a similar design:
More Similar Design to Pod Pro

The code I used to achieve that (**not** part of this PR):

```javascript
<Autocomplete
        // ...
        PaperComponent={({ children }) => (
          <Box
            sx={{
              bgcolor: 'black',
              opacity: '0.8',
              fontSize: '12px',
              width: 'max-content',
              transform: 'translateX(15%) translateY(8px)'
            }}
          >
            {children}
          </Box>
        )}
        // ...
      />
```

- One last idea I had was to make the component expand to be wider when selected, similar to [This](https://mdbootstrap.com/docs/standard/extended/search-expanding/). This could keep the UI clean, while also accommodating potentially long URLs in the dropdown. I tried incorporating this myself but ended up giving up... I don't have much experience using MUI, but if someone else does, I imagine it wouldn't be too difficult.


timbot1789 (commented):
I like this design a lot. It gets the user to the info they want quickly, but unintrusively. And it seems fairly easy to maintain and extend. Nice work!

Three suggestions:

1. Move the `SUGGESTED_OIDC_OPTIONS` into an environment variable (put it in `env.template`). That way we can have separate recommendations based on the deployment environment (we'll want to recommend localhost in dev, but not in production)
2. Change the suggestion list to be this list:
  - `http://localhost:3000/`
  - `https://opencommons.net/`
  - `https://solidcommunity.net/`
  - `https://login.inrupt.com/`
  - `https://inrupt.net/`
3. Add another unit test to `test/components/NavBar/OidcLoginComponent.test.jsx` that shows that when the login component has focus, the autocomplete box appears. I see you updated an existing test to select the new combobox, but we should add a new test that shows the combo box is doing what we expect. We use [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/) for that.

src/components/NavBar/NewOINK.jsx
rioredwards, line 1: This was a throwaway file which ended up in the commit by accident 😅

src/components/NavBar/OidcLoginComponent.jsx
rioredwards, line 1: I decided to only save the oidcIssuer to local storage when the login button is pressed... I had previously set it to save onBlur, onChange, etc... which seems unnecessary in hindsight. This fixes that.

src/components/NavBar/OidcLoginComponent.jsx
```diff
@@ -20,7 +20,9 @@ import { ENV } from '../../constants';
  */
 const OidcLoginComponent = () => {
   const { login } = useSession();
-  const SUGGESTED_OIDC_OPTIONS = ENV.VITE_SUGGESTED_OIDC_OPTIONS.split(', ');
+  const SUGGESTED_OIDC_OPTIONS = ENV.VITE_SUGGESTED_OIDC_OPTIONS?.split(', ') || [
+    'http://localhost:3000/'
+  ];
```
rioredwards, line 30: I set this variable to a default value of ['http://localhost:3000/'] for when the .env is not set up. This may or may not be necessary... 

I mostly added it because various tests that were not directly related to this component were failing. I couldn't figure out a good way to mock the .env throughout all the tests, so this line allows the component to render without needing the environment variable. Not sure if this is a good solution or not.

rioredwards:
Hey, @timbot1789! I made some changes based on your suggestions, let me know what you think...

I also added some comments to explain my thinking. I'm still learning and I wanna make sure that I'm following best practices and not breaking things! Don't hesitate to let me know if I need to make more changes.

src/components/NavBar/OidcLoginComponent.jsx
timbot1789: I think this is fine for now. We can always come back to it.

Generally, if you need just some URL in your code, `http://www.example.com` is a good place to point to. It's a special website reserved specifically for that purpose.

For PASS `opencommons.net` or `solidcommunity.net` are good spots as well.

src/components/NavBar/OidcLoginComponent.jsx
timbot1789: That sounds reasonable.

timbot1789 (approved):
Code itself looks good. However, it seems that we've left this linger too long and we have MERGE CONFLICTS. We'll need to resolve those, then we're good to merge.

leekahung:
> Code itself looks good. However, it seems that we've left this linger too long and we have MERGE CONFLICTS. We'll need to resolve those, then we're good to merge.

I could help with the clean-up if needed, let me know here or on Discord @rioredwards

rioredwards:
> I could help with the clean-up if needed, let me know here or on Discord @rioredwards

Sounds good, I'll send you a message on Discord

timbot1789:
@rioredwards you're good to merge this yourself once all the git checks have passed

PR #473, opened 2023-10-23
438/hmis editing framework

### LINKED ISSUE #438: Create HMIS Profile editing page + form framework
## Details:
When we sign up users for PASS, we want to get certain information about them. The more information we can collect about a person, the better able case managers will be to provide services. For this purpose, we want to create a registration workflow where a user can create a profile. This profile will need to contain a large amount of information (we will try to collect most of the data allowed in [this data ontology](https://hmis-interop.github.io/)), and so we will need a fairly sophisticated registration flow. For this issue, I'd like to do the following things:

1. Create a new tab in PASS called "HMIS Profile" that navigates you to this page
2. This page should be under the route `/hmis_profile` (use react router)
3. The page should display:   - On the right a form card that will serve as a placeholder for a basic info form. 
  - On the left there should be a sidebar allowing you to select different forms (housing status, employment history, etc). 
  - On the bottom should be next and previous buttons allowing you to navigate to the next or previous forms in the list.
image

## Additional considerations

- The driving goal of this UI is to make it very easy for a user to put in a large amount of optional, highly personal, and sometimes emotionally damaging information (e.g. history of domestic abuse). We want to make it easy for a user to skip over or ignore a section that is not relevant to them, or that they simply don't want to fill out right now.

## Additional information
- You may be able to use the [Material UI stepper component](https://mui.com/material-ui/react-stepper/) to implement the form progression
- I am not strongly attached to any particular UI design. If you have an idea to clean this up, please feel free to suggest it.
- I am generally basing this design off of tax software like [tax slayer](https://www.taxslayer.com/)

timbot1789:
## This PR:
Resolves #438 
 
**1. Adds a basic UI framework for adding HMIS forms. Right now it only contains placeholder forms to demonstrate how forms should function and how they can be tested. The forms themselves will come in later PRs. 

This framework also contains very little styling. I wanted to keep things basic so UI/UX could style things as they see fit.
  

[Screencast from 10-22-2023 05:09:08 PM.webm](https://github.com/codeforpdx/PASS/assets/37914436/137757b1-85fb-44db-b600-1447f80a4684)

I chose not to use the MUI steps component here because I imagine this to be a display of a large amount of information a person can fill out over several sessions (or never fill out at all). So it didn't make much sense to me to say "housing info must be filled out before financial info." I wanted a layout that could give a suggested progression, while still letting the user jump to any form they want.

In a follow up PR, we will need to tie these forms to the pod itself. That will involve creating a new document on the pod, and creating a hook on the front end to access it.


src/components/HmisForms/FormLayout.jsx
```diff
@@ -0,0 +1,36 @@
...
+  const path = location.pathname.split('/').pop();
+
+  localStorage.setItem('restorePath', location.pathname);
+  const pageIdx = HMIS_FORM_LIST.findIndex((form) => form.path === path);
+  return (
+    <Container sx={{ margin: '8px' }}>
+      <Card>{children}</Card>
+      <Container sx={{ display: 'flex', justifyContent: 'space-between' }}>
+        {pageIdx > 0 ? (
+          <Link component={RouterLink} to={`../${HMIS_FORM_LIST[pageIdx - 1].path}`}>
```
timbot1789: The `..` at the start of these routes means "go to a sibling route"

leekahung (commented):
Think this is looking great so far. Code is pretty good as well. Good stuff!

With regards to the wording of things, I think we could change "HMIS Profile" to something like "Personal Profile". While under the hood, everything involved in this profile will be HMIS-related, I think don't think we need to expose normal users to it, similar to how we're limiting exposure to Solid.

andycwilliams (commented):
Looks like a really good foundation.

I agree with @leekahung about making the names and phrasing more general. 

And I take it we're not planning on using modals for these forms? Not an issue at all, just wanting to be certain. Since there are multiple consecutive forms, it probably makes more sense to not have to deal with a series of modals anyway.

timbot1789:
> With regards to the wording of things, I think we could change "HMIS Profile" to something like "Personal Profile". While under the hood, everything involved in this profile will be HMIS-related, I think don't think we need to expose normal users to it, similar to how we're limiting exposure to Solid.

Sure that sounds fine to me. Renamed everything to "personal profile." We will need to make sure that users understand the difference between the "personal profile" here and the "public profile" that's stored on the web ID though. Maybe "civic profile?" We'll need to futz around with the wording, but that doesn't need to be done for a while.

image

timbot1789:
> I take it we're not planning on using modals for these forms? Not an issue at all, just wanting to be certain. Since there are multiple consecutive forms, it probably makes more sense to not have to deal with a series of modals anyway.

Correct, I don't think a series of modals makes sense for the workflow I'm imagining here. Same reason why I didn't use the stepper component Scott suggested in the issue description. I'm imagining 2 main use cases:

1. A user is filling out their profile for the first time, and needs to be guided through it
2.  A user has filled out their profile before, and needs to reference or update a specific part of it

Modals work for use case 1, but don't really work for use case 2. This UI mostly targets use case 2, while still allowing for use case 1. I wanted to focus on case 2 to also make development/testing of the profile easier. A dev won't want to go through a progression of modals every time they need to update a single value if they don't have to.

If we find we need a better flow for first time users, we can definitely add a stepper or series of modals later.

timbot1789:
Actually, I think I like the name "Civic Profile." I updated the PR to that name. Take a gander and let me know what you think.

leekahung:
> Sure that sounds fine to me. Renamed everything to "personal profile." We will need to make sure that users understand the difference between the "personal profile" here and the "public profile" that's stored on the web ID though. Maybe "civic profile?" We'll need to futz around with the wording, but that doesn't need to be done for a while.
> 
> image
> 

Yeah, the naming stuff could get pretty messy.

Think we could have something like "Account Settings" for the public profile. That way, it's like a user setting up things in their gmail account related only to gmail stuff.

Then we could move "Date of Birth" back out and move it into the hmis profile, so it's like setting up the personal information within the system.

leekahung:
Actually, thinking about it again. Why not have this under the existing `/profile` route. I think we could make that work.

We could use `/profile` as a base route for the tabs and include what we have now for profile as one of those tabs (maybe something like `/profile/account` or `/profile/user`). So we can get 4 links under `/profile`.

And we can reroute the contact profile links to `/profile/account/<other_user_webId>` or `/profile/user/<other_user_webId>`

andycwilliams:
> Actually, thinking about it again. Why not have this under the existing `/profile` route. I think we could make that work.
> 
> We could use `/profile` as a base route for the tabs and include what we have now for profile as one of those tabs (maybe something like `/profile/account` or `/profile/user`). So we can get 4 links under `/profile`.
> 
> And we can reroute the contact profile links to `/profile/account/<other_user_webId>` or `/profile/user/<other_user_webId>`

I like this. It makes sense to me to consolidate routes when we can.

timbot1789:
> Actually, thinking about it again. Why not have this under the existing `/profile` route. I think we could make that work.
> 
> We could use `/profile` as a base route for the tabs and include what we have now for profile as one of those tabs (maybe something like `/profile/account` or `/profile/user`). So we can get 4 links under `/profile`.
> 
> And we can reroute the contact profile links to `/profile/account/<other_user_webId>` or `/profile/user/<other_user_webId>`

I want to separate the private profile and public profile as much as possible. It will be confusing enough as is for users to have two separate profiles. I don't want to risk a situation where a person accidentally puts confidential information in their public profile. Having them as forms right next to each other would definitely do that.

I think having one flow to edit the public webId and separate, distinct flow to edit confidential information is a good way to communicate their difference.

leekahung:
> I think having one flow to edit the public webId and separate, distinct flow to edit confidential information is a good way to communicate their difference.

Think that should be good. We could keep them as separate pages (one with the tabs like what is shown here and one with our existing profile page) and have the routes consolidated under `/profile`.

Jared-Krajewski:
This looks good to me. I think caseworkers will be happy to have this place to collect information from.

Jared-Krajewski (approved):
looks good to me. I think caseworkers will be happy to have a place to collect this information from. I know its not of any immediate necessity but  any thoughts on possibly auto populating some of this information off of the license barcode scanner or having it populate into a section within these forms?

andycwilliams (approved):
Any suggestions I am aware of can be left for a future PR/discussion. It's good to get this framework in place now.

PR #482, opened 2023-10-29
440/create civic profile framework

timbot1789:
Adds a hook to allow working with civic profile in react components. Also adds more tests to `useContactsList` and fixes a couple bugs I found there.


leekahung (approved):
The additions looks good to me.

src/hooks/useCivicProfile.js
```diff
@@ -0,0 +1,99 @@
...
+    .addDate(RDF_PREDICATES.legalDOB, dateOfBirth)
+    .addInteger(RDF_PREDICATES.legalGender, gender)
+    .build();
+
+let storedDataset;
+const useCivicProfile = () => {
+  const queryClient = useQueryClient();
+  const { session, podUrl } = useSession();
+  const { fetch } = session;
+  const fileUrl = podUrl && new URL('PASS/AdditionalProfiles/civic_profile.ttl', podUrl).toString();
```
leekahung: One small thing is, we've already created a path previously for additional profile files at `PASS/Profile/` (it was where we stored the previous information about data of birth under `privateProfile.ttl`) in their pods. So was thinking we could do the same thing here and repurpose the route to have `civic_profile.ttl` instead.

andycwilliams (approved):

src/hooks/useCivicProfile.js
timbot1789: Sure

PR #483, opened 2023-10-30
Create RDF Collection Hook

timbot1789:
Creates a shared hook for use with data models and parsing, and refactors useContactsList and useCivicProfile to use them.

The hook is designed to act as a base for shared logic for other hooks, which can then directly expose simple interfaces to components.

It takes the following props:

- parse: (data: solidDataset) => any 
     a function to parse the solidDataset. The function receives a solidDataset, and its return value will be placed in the hook's `data` parameter
- serialize: (data: any) => Thing
  A function used to convert a data object into a `Thing` that can be used to interact with an RDF graph. This is used primarily by the `add` mutation
- fileUrl: string | URL
  the URL to load the collection from
- fetchData: async (url: URL, params)
  A specialized `fetch`

It returns all the properties from React query along with a few additional ones:

- `add` (object: any) => unknown 
  A mutation to add an object to the graph. You can also use this to update a Thing if the object maps to the same ID
- `delete` (id: string) => unknown
  A mutation to remove a Thing from the graph. Pass in the ID/URL of the Thing you wish to remove, and it will delete the thing


leekahung:
Hey @timbot1789, just wanted to make sure about this PR. Is this ready for review or it's still in progress?

timbot1789:
> Hey @timbot1789, just wanted to make sure about this PR. Is this ready for review or it's still in progress?

This is ready for review

src/hooks/useRdfCollection.js
```diff
@@ -0,0 +1,84 @@
...
+import {
+  getThing,
+  removeThing,
+  setThing,
+  createSolidDataset,
+  getSolidDataset,
+  saveSolidDatasetAt
+} from '@inrupt/solid-client';
+
+const useRdfCollection = (parse, serialize, fileUrl, fetchData) => {
```
leekahung, line 35: Think we could include JSDoc to describe what the new parameters are for this shared hook. The description in the PR should suffice.

leekahung (approved):
I like this. The shared hook definitely helps streamline how we process RDFs to and from Solid. I'm good with approving this.

Also, I think we could add what you used for the description for the new parameters as JSDoc comments for the hooks for documentation.

src/hooks/useRdfCollection.js
timbot1789: Sounds good. Added.

PR #460, opened 2023-10-17
Bug 454: when clicking on message in inbox the state of the unread me…

bingeboy:
# Message Badge Not Updating After Message Read

## This PR:
Resolves #454 
Added the React update to the `<Messages />` page so the state will stay in sync with the podserver message read status for the inbox. 

## Video:
[screen-recorder-mon-oct-23-2023-12-58-19.webm](https://github.com/codeforpdx/PASS/assets/96048/5407c9e0-9234-435c-8831-38f64fb5834c)

### How to test

1. Login to podserver and navigate to the messages section
2. send more than one message to a pod you have access to or yourself
3. check that the count is on the message icon on the top right
4. state should update once the podserver is updated

## Issues needing discussion/feedback (optional):
**1.**  Need to add test coverage... could be an additional issue

PR #485, opened 2023-10-31
Issue 414/add message to contacts

### LINKED ISSUE #414: Enhancement: Add "Message" action to contacts list

### LINKED ISSUE #422: Incorrect jsDoc comments in `AddContactModal.jsx`
### Bug Title:
Fix incorrect jsDoc in `AddContactModal.jsx`

---
### Bug Description:
The jsDoc for `renderWebId` function at top of file is inaccurate.
The jsDoc for `AddContactModal` is inaccurate.

andycwilliams:
## This PR:
Resolves #414 

I've been focused elsewhere and am only now getting back into the swing of PASS, so please make sure this is proper.

**1.** Adds a message button to each row in the contacts list
**2.** This button auto populates the `To` field with the corresponding username

Capture
Capture2

**1.** Determine whether this code could be more efficient
**2.** Determine if the message icon is adequate or should be replaced


leekahung:
At the moment, the send message function hard-codes the trailing `profile/card#me` that comes with all solid webIds so you only need to enter the pod URL for the `To` field, sort of like not needing to type `www.` for websites.

Was thinking of modifying the UI for the `To` field to make that more apparent (like with a placeholder or text at the end of the field).

src/components/Contacts/ContactListTable.jsx
```diff
@@ -62,6 +72,15 @@ const ContactListTable = ({ contacts, deleteContact }) => {
       headerAlign: 'center',
       align: 'center'
     },
+    {
+      field: 'Message',
+      renderCell: (contactId) => <SendIcon onClick={() => handleSendMessage(contactId)} />,
+      sortable: false,
+      filterable: false,
+      width: 80,
```
leekahung, line 85: Think we could have all the columns with interactive buttons be the same width. So perhaps we could make the other two columns 80px width as well.

andycwilliams:
> At the moment, the send message function hard-codes the trailing `profile/card#me` that comes with all solid webIds so you only need to enter the pod URL for the `To` field, sort of like not needing to type `www.` for websites.
> 
> Was thinking of modifying the UI for the `To` field to make that more apparent (like with a placeholder or text at the end of the field).

Do you mean you'd like to do that in this PR, or just that it's something planned for the future?

> Think we could have all the columns with interactive buttons be the same width. So perhaps we could make the other two columns 80px width as well.

Done in next push.

leekahung:
> > At the moment, the send message function hard-codes the trailing `profile/card#me` that comes with all solid webIds so you only need to enter the pod URL for the `To` field, sort of like not needing to type `www.` for websites.
> > Was thinking of modifying the UI for the `To` field to make that more apparent (like with a placeholder or text at the end of the field).
> 
> Do you mean you'd like to do that in this PR, or just that it's something planned for the future?

Think we can do it in a later PR.

andycwilliams:
> Think we can do it in a later PR.

Sounds good. Is there anything more needed for this?

leekahung:
> > Think we can do it in a later PR.
> 
> Sounds good. Is there anything more needed for this?

If it's still using webId for the "To" field, it needs to be updated to use Pod URL (unless we update the sendMessage function in utils to just accept straight up webIds as well). Other wise think it should be ready?

Is there also a way to make the send icon gray like the other icons? Makes them consistent to the other icons on the row.

andycwilliams:
> If it's still using webId for the "To" field, it needs to be updated to use Pod URL (unless we update the sendMessage function in utils to just accept straight up webIds as well). Other wise think it should be ready?

So just removing `profile/card#me` from `https://exampleusername.solidcommunity.net/profile/card#me`? That should be simple

> Is there also a way to make the send icon gray like the other icons? Makes them consistent to the other icons on the row.

Yep, that's easy.

leekahung:
> > If it's still using webId for the "To" field, it needs to be updated to use Pod URL (unless we update the sendMessage function in utils to just accept straight up webIds as well). Other wise think it should be ready?
> 
> So just removing `profile/card#me` from `https://exampleusername.solidcommunity.net/profile/card#me`? That should be simple

Yeah, I think you could just pull contact.podUrl directly from the contact object:

Screen Shot 2023-10-31 at 19 47 53

andycwilliams:
> Yeah, I think you could just pull contact.podUrl directly from the contact object:

Should be good to go now?

leekahung (changes requested):
Had previously thought it was good to go, but it looks like there still a few kinks to work out.

The first being that I couldn't send any message and the error not entering a recipient pod URL appeared. Looking at `NewMessageModal.jsx`, it's likely that `sendMessageTTL` function in Line 83 did not trigger. The `handleSubmit` function needs to take into consideration the new `toField` variable as well.

Second, the cursor for the send icon remains `cursor` instead of `pointer`. This one should be easier to handle.

https://github.com/codeforpdx/PASS/assets/14917816/e8e1e69a-93c2-4e53-a4ad-5dbf2a6aee2a

leekahung:
> Had previously thought it was good to go, but it looks like there still a few kinks to work out.
> 
> The first being that I couldn't send any message and the error not entering a recipient pod URL appeared. Looking at `NewMessageModal.jsx`, it's likely that `sendMessageTTL` function in Line 83 did not trigger. The `handleSubmit` function needs to take into consideration the new `toField` variable as well.
> 
> Second, the cursor for the send icon remains `cursor` instead of `pointer`. This one should be easier to handle.
> 
>  Screen.Recording.2023-11-02.at.0.20.36.mov

Thinking about it again, I think we could also disable the `To` field since we don't really need to change that if it's going to be a direct message to that person from the contact list.

src/components/Modals/NewMessageModal.jsx
```diff
@@ -32,11 +32,12 @@ import { MessageContext, SignedInUserContext } from '@contexts';
  * @param {boolean} Props.showModal - Boolean for showing message modal
  * @param {React.Dispatch<React.SetStateAction<boolean>>} Props.setShowModal
  * - React set function for showModal
+ * @param Props.toField
```
leekahung: The JSDoc for Line 35 needs to be corrected. It's missing the type and a description. The following format should do:

```js
/*
 * @param {variable_type} variable_in_question - Description of variable
 */
```

You could check the other `@param` lines in the JSDoc for reference.

andycwilliams: All the changes should be implemented now sans the `sendMessageTTL`. Still working on that.

Also wanted to mention that going to the Messages tab gives this error:
```
MessagePreview.jsx:100 Warning: Each child in a list should have a unique "key" prop.
```

leekahung: > All the changes should be implemented now sans the `sendMessageTTL`. Still working on that.
> 
> Also wanted to mention that going to the Messages tab gives this error:
> 
> ```
> MessagePreview.jsx:100 Warning: Each child in a list should have a unique "key" prop.
> ```

This error is related to the rendering of the message list. @bingeboy got that fixed in PR #460. We can ignore that here.

leekahung (approved):
Just confirmed the changes. Functionality works now.

Also, not really a blocker, since Line 159 in `src/components/Modals/NewMessageModal.jsx` with the double `!` works logically, but a bit confusing to read. Think it'll be better to change it to `toField !== ""`.

PR #490, opened 2023-11-05
update packages

timbot1789:
Updates vite, vitest, and a few other things. Was hoping to improve test run times. Didn't work. But the app builds faster now.


leekahung (approved):
Good with package updates. 👍

PR #497, opened 2023-11-07
[New Feature] - Issue 488/breadcrumb navigation

### LINKED ISSUE #488: [New Feature] - Implement Breadcrumb Navigation
### Feature Description:

While navigation within PASS is quite accessible via the navbar and nav menu, I think we can go one further with breadcrumbs especially if people simply want to jump back to an earlier route within that page. The UX team had a preliminary design for this. Think we should get that implemented into PASS.

### Screenshots (if applicable):

breadcrumbs

leekahung:
## This PR:

Resolves Issue #488. This PR also includes minor updates to routes and routing names for easier route management for the new Breadcrumbs implementation. The new Breadcrumbs component will be placed under `src/Layouts` as `Breadcrumbs.jsx` and imported as part of `Layout.jsx`. It renders only when logged into PASS (See clip below)

For the route which the person navigated to will be turned into a text element rather than a link, but all parent routes remains a link.

Other changes include routing changes from `/profile/:webId` to `/contacts/:webId`. While it's intuitive to place profiles under "/profiles" and any potential children routes, it makes more sense to have it placed under "/contacts/:webId" instead since that's where people are accessing their profile from. The existing "/profile" route, however, remains the same for the user's own profile page.

https://github.com/codeforpdx/PASS/assets/14917816/d689b55e-f364-472c-a960-3527d96e0b31

### DIFF
--- a/test/layouts/Layout.test.jsx
+++ b/test/layouts/Layout.test.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { expect, it, vi } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 import { render } from '@testing-library/react';
 import { BrowserRouter } from 'react-router-dom';
 import * as hooks from '@hooks';
@@ -8,6 +8,7 @@ import { ThemeProvider } from '@mui/material/styles';
 import { NavBar } from '@components/NavBar';
 import Layout from '../../src/layouts/Layout';
 import theme from '../../src/theme';
+import createMatchMedia from '../helpers/createMatchMedia';
 
 const MockLayout = () => (
   <ThemeProvider theme={theme}>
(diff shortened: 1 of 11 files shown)


bingeboy:
Pulled and tested locally. 👍

test/layouts/Layout.test.jsx
```diff
@@ -0,0 +1,117 @@
...
+      <Layout ariaLabel="test">
+        <NavBar />
+      </Layout>
+    </BrowserRouter>
+  </ThemeProvider>
+);
+
+describe('Desktop view', () => {
+  it('Renders breadcrumb when logged in, layout grid-template-rows as 64px 64px 1fr 280px', () => {
+    vi.spyOn(hooks, 'useNotification').mockReturnValue({
```
bingeboy, line 25: Slightly off topic of the PR. I checked the `hook/useNotification.js` file and noticed that it was importing the context file and the call `useContext(NotificationContext)` was in the hooks folder. This is the pattern we should all be using for useContext() hooks?

leekahung: ~~Good catch. No, I think we should either just have `Notification` use straight up `useContext` or migrate everything over to the custom hook. I'll open an issue for it.~~

Edit: See comment below.

bingeboy (approved):
LGTM. Nice job.

test/layouts/Layout.test.jsx
leekahung: Hm...scratch that. It might be reasonable to format it like what we have now to `useNotification` and `NotificationContext` since `Notification` is one of our global states used across multiple files. I think that's what we're doing with `useSession` and `SessionContext` is doing. Also reduces the need to import both `useContext` and the context, we just need to import the custom hook.

Edit: Ah, okay. I see some examples for this: https://www.johno.com/using-react-context-with-a-custom-hook ; It looks like people usually make the context/hook that within the same file for export. But given we have dedicated contexts and hooks directories, think what we have now with the context and hooks split up into their own files is good.

bingeboy: Sounds good to me. Thanks for sharing. I've also come across the use of custom error messages declared in Context files which seemed kinda nice.
 
Example:
```
const FooContext = createContext();

export const FooProvider = ({ children }) => {
  //... 
  return <FooContext.Provider value={value}>{children}</FooContext.Provider>;
};

export const useFooContext = () => {
  const context = useContext(FooContext);
  if (!context) {
    throw new Error('useFooContext must be used within an FooProvider');
  }
  return context;
};
```

PR #498, opened 2023-11-08
Camera button works for both desktop and mobile

### LINKED ISSUE #335: [Open Discussion] Device camera uploads

Zakeriya23:
## This PR:Enhance UploadButtonGroup with Webcam Support and Responsive Behavior
Resolves #335 
This PR updates the UploadButtonGroup component to improve its functionality and user experience. Key enhancements include:

Webcam Integration: Added the ability to capture images with a webcam. This feature is enabled via a new "Use Webcam" button that appears only when a webcam is detected on the user's device.
Responsive Design: The component now dynamically renders a "Capture image" button when the screen width is less than 768px, making the interface more mobile-friendly.
File Handling Improvements: A function dataURItoBlob was added to handle captured images more efficiently and convert them to Blob format for easier manipulation.
image
image


leekahung:
Hey @Zakeriya23! Thanks for contributing to the project, I see you've managed to get the branch working.

Could you update the PR description and possibly include a video clip/screenshot of the feature so that other could get a better feel of what this PR is about and how it would work.

For this feature, think it'll be nice to show that it can be used both from the computer webcam as well as from a phone camera. (See PR #357)

timbot1789 (commented):
This is a good start, but I think there's still a bit of work left here. This code seems to replace the camera upload button with a webcam button, which is a behavior regression (and is partly why this PR is failing unit tests).

We instead should have a button that _gracefully degrades_ based on the capabilities of the hosting device. If we detect that a device has a front-facing camera, we should show the current camera button. If we don't detect a front-facing camera, but do detect a webcam, show this webcam button instead. If we detect neither a front-facing camera, nor a webcam, don't show any button at all (there's no need to cover this case in this PR).

leekahung:
Hey @Zakeriya23! Just pulled your branch to view the changes.

I agree with @timbot1789, checking whether or not a device has a forward facing camera, a webcam, or neither would be a great way to condition between utilizing what we have now with the camera feature and what you've created for this PR.

src/components/Modals/UploadButtonGroup.jsx
```diff
@@ -1,12 +1,14 @@
...
 // Material UI Imports
 import Box from '@mui/material/Box';
 import Button from '@mui/material/Button';
 import PhotoCameraIcon from '@mui/icons-material/PhotoCamera';
 import SearchIcon from '@mui/icons-material/Search';
-import useMediaQuery from '@mui/material/useMediaQuery';
-import { useTheme } from '@mui/material/styles';
+import WebcamModal from './WebcamModal';
+// import useMediaQuery from '@mui/material/useMediaQuery';
+// import { useTheme } from '@mui/material/styles';
```
leekahung: While there are other functions that would be used to detect camera/webcams, these media query functions and theming would be of great help for the styling. Think we could leave them on until we come out with a workable design that encompasses both this PR and previous one.

src/components/Modals/UploadButtonGroup.jsx
```diff
@@ -19,22 +21,48 @@ import { useTheme } from '@mui/material/styles';
...
+  // Write the bytes of the string to a typed array
+  const ia = new Uint8Array(byteString.length);
+  for (let i = 0; i < byteString.length; i += 1) {
+    ia[i] = byteString.charCodeAt(i);
+  }
+
+  return new Blob([ia], { type: mimeString });
+};
+
 const UploadButtonGroup = ({ file, setFile }) => {
```
leekahung, line 51: While it utilizes screen width to assume what devices might have front-facing cameras, the `UploadButtonGroup` component was tailored-made for that purpose. Think you could create a new component in this file or in the Modals directory for the button group for the web cam and render it based on the new condition of whether or not there's a camera, webcam, or neither.

src/components/Modals/WebcamModal.jsx
```diff
@@ -0,0 +1,47 @@
...
+        }}
+      >
+        <h2 id="webcam-modal-title">Webcam</h2>
+        <Webcam
+          audio={false}
+          ref={webcamRef}
+          screenshotFormat="image/jpeg"
+          videoConstraints={{ facingMode: 'user' }}
+        />
+        <Button variant="contained" color="primary" onClick={handleCapture}>
```
leekahung, line 39: Although this is a good start, was wondering if there could be another button that could be created for simply exiting "webcam mode"?

leekahung (changes requested):
While I would still like to see the PR description updated, I think we could wait until this is more finished. There are a few things I would like you to address, but think they'll be great additions to this PR.

See feedback and comments.

xscottxbrownx:
@Zakeriya23 are you still interested in this issue, or should somebody take it over?

Zakeriya23:
@xscottxbrownx  This feature has become my main priority. Thank you @timbot1789 @leekahung  for the feedback, I will make those changes!

src/components/Modals/UploadButtonGroup.jsx
Zakeriya23: I agree

Zakeriya23:
Yes, I agree with the comments and strategy that both @leekahung and @timbot1789 have advised. It makes sense and puts the needs of the user first to modify our button's functionality to match the capabilities of the user's device. Enhancing user experience and compatibility, it makes sure our program stays flexible and works on a variety of devices. I will work on putting these modifications into practice so that our button, depending on the system, shows the webcam choice or the camera upload.

Zakeriya23:
@leekahung I am not sure why my screenshot and PR description didn't fully upload. I will edit it requick to include those again.

package.json
```diff
@@ -45,7 +45,11 @@
     "react-dom": "^18.2.0",
     "react-paginate": "^8.2.0",
     "react-router-dom": "^6.10.0",
+<<<<<<< HEAD
+    "react-webcam": "^7.2.0",
+=======
     "react-router-hash-link": "^2.4.3",
+>>>>>>> e81082147c2a8c3c25c1ea905e80ecf248d5bdad
```
leekahung: These things in package.json appears to be remnants of the merge conflict resolution. The `<<<<<< HEAD`, `=======`, and `>>>>>>> e8108...` can all be removed.

I recommend removing your local `node_modules` and running a fresh `npm install` while keeping your `package.json` and `package-lock.json` intact.

src/components/Modals/UploadButtonGroup.jsx
```diff
@@ -19,22 +22,73 @@ import { useTheme } from '@mui/material/styles';
...
-  const theme = useTheme();
-  const isSmallScreen = useMediaQuery(theme.breakpoints.down('sm'));
-  const isCustomMediumScreen = useMediaQuery('(max-width: 768px)');
+  const [hasCamera, setHasCamera] = useState(false);
+
+
+  useEffect(() => {
```
leekahung, line 54: I see `useEffect` has been used, but was not imported. It's probably the cause of the error.

https://github.com/codeforpdx/PASS/assets/14917816/c870f05c-3cf5-498d-b9e1-f7ac56a6fc7e

Update: Just double-checked. The console.log points to `useEffect` not being imported as the main source of error.

Screenshot 2024-01-11 at 11 00 33 PM

src/components/Modals/WebcamModal.jsx
leekahung: Looks like you can escape using the `esc` key on the keyboard. I'm resolving this.

src/components/Modals/UploadButtonGroup.jsx
```diff
@@ -53,24 +107,28 @@ const UploadButtonGroup = ({ file, setFile }) => {
...
           fullWidth
-          required
           startIcon={<PhotoCameraIcon />}
+          sx={{ borderRadius: '20px', marginLeft: '8px' }}
         >
-          Capture image
-          <input type="file" hidden accept="image/*" capture="environment" />
+          Use Webcam
         </Button>
       )}
```
leekahung: There really should be two buttons, but one only renders depending on condition. So instead of:

```
{hasCamera && <WebCam Button>}
```

It really should be:

```
{hasCamera ? <WebCam Button> : <Front-facing Cam Button>}
```

You want to keep the old version to cover both cases since there could be a time when hasCamera is false assuming no web-cam, but the user still has a front-facing one, which in that case, the old version should still work.

https://github.com/codeforpdx/PASS/assets/14917816/d51fba7d-2499-4d1d-981f-23807c386db6

leekahung (changes requested):
Function-wise (assuming the fixes are made with the package.json and useEffect import error), think it should be good to go.

https://github.com/codeforpdx/PASS/assets/14917816/328b5d24-fbb1-4a39-bdfa-20f67ab76264

Keep note of the feedback for changes though.

Other than that, we want the unit tests to be updated so that it'll handle your new component.

package.json
Zakeriya23: Agree

Zakeriya23 (commented):
Implement those changes.

src/components/Modals/UploadButtonGroup.jsx
```diff
@@ -1,12 +1,13 @@
...
 // Material UI Imports
 import Box from '@mui/material/Box';
 import Button from '@mui/material/Button';
 import PhotoCameraIcon from '@mui/icons-material/PhotoCamera';
 import SearchIcon from '@mui/icons-material/Search';
-import useMediaQuery from '@mui/material/useMediaQuery';
-import { useTheme } from '@mui/material/styles';
+import WebcamModal from './WebcamModal';
+// import useMediaQuery from '@mui/material/useMediaQuery';
+// import { useTheme } from '@mui/material/styles';
```
leekahung: Considering these imports are no longer in use, think we could remove them.

test/components/Modals/UploadButtonGroup.test.jsx
```diff
@@ -2,33 +2,12 @@ import React from 'react';
 import { expect, it } from 'vitest';
 import { render } from '@testing-library/react';
 import UploadButtonGroup from '../../../src/components/Modals/UploadButtonGroup';
-import createMatchMedia from '../../helpers/createMatchMedia';
 
 const MockUploadButtonGroup = () => <UploadButtonGroup />;
 
 it('renders only one button above 768px', () => {
```
leekahung: Think this test description needs to be updated based on the updates.

leekahung (commented):
Minor comments on the code. I'll check more tomorrow, but think we're close.

src/components/Modals/UploadButtonGroup.jsx
Zakeriya23: Removed

src/components/Modals/UploadButtonGroup.jsx
```diff
@@ -19,22 +20,70 @@ import { useTheme } from '@mui/material/styles';
  * @param {Function} Props.setFile - The set function for handling files
  * @returns {React.JSX.Element} - The UploadButtonGroup Component
  */
+
+const checkCameraAvailability = async () => {
```
leekahung: Considering the function is make specifically for checking webcam availability, think we could update the function name to `checkWebCamAvailability`

src/components/Modals/UploadButtonGroup.jsx
```diff
@@ -19,22 +20,70 @@ import { useTheme } from '@mui/material/styles';
...
+  }
+
+  return new Blob([ia], { type: mimeString });
+};
+
 const UploadButtonGroup = ({ file, setFile }) => {
-  const theme = useTheme();
-  const isSmallScreen = useMediaQuery(theme.breakpoints.down('sm'));
-  const isCustomMediumScreen = useMediaQuery('(max-width: 768px)');
+  const [hasCamera, setHasCamera] = useState(false);
```
leekahung: Rather than using `hasCamera`/`setHasCamera` for the state, it might be better to use `hasWebCam`/`setHasWebCam`.

This makes things less ambiguous since `hasCamera` could refer to any camera on the device. Having `hasWebCam` directly refers to a specific camera on the devices which the earlier function targets.

package.json
```diff
@@ -59,7 +60,7 @@
   ],
   "devDependencies": {
     "@rollup/plugin-node-resolve": "^15.1.0",
-    "@solid/community-server": "^6.1.0",
+    "@solid/community-server": "^7.0.3",
```
leekahung: Was not expecting this to be changed in this PR. Leave this at version 6. We should do library updates for features not related to your changes in a separate PR.

leekahung (changes requested):
One minor thing left. See feedback.

package.json
Zakeriya23: Reverted back.

leekahung (approved):
Think this looks good now. Can approve this for merging. 👍

Remember to update the PR description so everyone would know what this PR is for.

PR #505, opened 2023-11-08
351 add relationship info

### LINKED ISSUE #351: Add relationship information to contacts

ofu997:
## This PR:
Resolves #351 
 

image

Add relation and relationship status to contact. We are currently (Nov 14) not sure whether we will set this in the add contact modal or from the table in the contact list. 

Currently this creates two new constants files, and changes the useContactsList hook. 

**1.**  I'm unable to do the pod field which is the second field in the screenshot. ```add this field:   pod: <podUrl>, // optional: the person's preferred pod url. If not present, we fetch the first pod url on the profile document. If that's not present, we derive it from the webid.``` When I tried to do it in previous commits there would be Promise or auth errors. This needs to be done later as a separate issue.


leekahung:
Hey @ofu997! Thanks for your contribution to the project.

I have a general sense of what the changes are, but would like to have the PR description updated to detail what you've done/changed for this PR. It'll help let other users get a feel for what this PR is for.

Also, keep note of unused imports and other errors provided by the linter and update the codebase accordingly (see lint check on GitHub: https://github.com/codeforpdx/PASS/actions/runs/6794649087/job/18471386525?pr=505).

src/components/Modals/AddContactModal.jsx
```diff
@@ -64,7 +71,10 @@ const AddContactModal = ({ addContact, showAddContactModal, setShowAddContactMod
     const userObject = {
       givenName: event.target.addUserGivenName.value,
       familyName: event.target.addUserFamilyName.value,
-      webId: event.target.addWebId.value
+      webId: event.target.addWebId.value,
+      pod: event.target.pod.value,
```
leekahung: `pod` is redundant here. It's already part of the object utilized in contacts.

src/components/Modals/AddContactModal.jsx
```diff
@@ -48,6 +52,9 @@ const AddContactModal = ({ addContact, showAddContactModal, setShowAddContactMod
   const [userFamilyName, setUserFamilyName] = useState('');
   const [username, setUsername] = useState('');
   const [webId, setWebId] = useState('');
+  const [pod, setPod] = useState('');
+  const [relationship, setRelationship] = useState('');
+  const [relationshipStatus, setRelationshipStatus] = useState('');
```
leekahung: Unsure about adding these to the modal at this time.

src/components/Modals/AddContactModal.jsx
```diff
@@ -64,7 +71,10 @@ const AddContactModal = ({ addContact, showAddContactModal, setShowAddContactMod
     const userObject = {
       givenName: event.target.addUserGivenName.value,
       familyName: event.target.addUserFamilyName.value,
-      webId: event.target.addWebId.value
+      webId: event.target.addWebId.value,
+      pod: event.target.pod.value,
+      relationship: event.target.relationship.value,
+      relationshipStatus: event.target.relationshipStatus.value
```
leekahung: Also think, we could leave these as blank fields for the time being?

src/components/Modals/AddContactModal.jsx
```diff
@@ -151,39 +163,65 @@ const AddContactModal = ({ addContact, showAddContactModal, setShowAddContactMod
...
-            <Box
-              sx={{
-                display: 'flex',
-                flexDirection: isSmallScreen ? 'column' : 'row',
-                gap: isSmallScreen ? '10px' : '8px',
-                justifyContent: 'center',
-                alignItems: 'center',
-                width: '100%'
-              }}
+          <TextField
```
leekahung: The new additions to the modal are likely the cause for the failing tests as shown in the the test checks.

src/components/Modals/AddContactModal.jsx
```diff
@@ -151,39 +163,65 @@ const AddContactModal = ({ addContact, showAddContactModal, setShowAddContactMod
...
+          <Select
+            labelId="relationship-label"
+            id="relationship"
+            name="relationship"
+            value={relationship}
+            label="Relationship"
+            onChange={(e) => setRelationship(e.target.value)}
+            fullWidth
+          >
+            <MenuItem value={RELATIONSHIPS.Blank}></MenuItem>
```
leekahung: In React, you could close empty components without closing tags. But you'll need to include a `/` before the component to signify that it's closed, which is why the Lint is failing this line.

For example using what is shown here, the correct syntax should be `<MenuItem value={RELATIONSHIPS.Blank} />` and that should be fine.

As for what I think about adding these new inputs, see feedback from before.

src/constants/relationship_status.js
```diff
@@ -0,0 +1,8 @@
+const RELATIONSHIP_STATUS = {
+  BLANK: '',
+  ACTIVE: 'self',
```
leekahung: A bit confused over the choice and wording of key:value pairs. Shouldn't we just use `active` for `ACTIVE` and `archive` for `ARCHIVE`?

Also think we could use `other` for `OTHER` as a relationship status rather than `etc`/`ETC`.

src/constants/relationships.js
```diff
@@ -0,0 +1,8 @@
+const RELATIONSHIPS = {
+  Blank: '',
```
leekahung: A bit confused about the choice of casing? It would seem `src/contants/relationship_status.js` utilizes upper casing for its keys, lower casing for its values, but `src/constants/relationship.js` utilizes capitalization throughout. However, I think it's probably due to the old INTERACTION_TYPES from `src/constants/interaction_types.js` that's causing the confusion. We'll probably need to remove the older `src/constants/interaction_types.js` since it's no longer in use.

After looking over all our `constants` files, I think we should stick with the convention set by `src/contants/HMIS_ONTOLOGY_VALUES.js` and `src/constants/rdf_predicates.js` with camelCasing keys at the very least.

leekahung (changes requested):
Hey @ofu997! Just gone through some of the code. It looks like this aims to add new fields to the contacts object that we'll be storing with the contactsList, which I think is good. However, I'm not too sure about adding them as adjustable fields in `src/components/Modals/AddContactModal.jsx` just yet. Think we'll need more discussion about how to implement these new properties.

There are also other thing that need addressing like the unit tests. See feedback below.

src/components/Modals/AddContactModal.jsx
ofu997: I will work on the unit tests. So there will be another way to add 'relation' and 'relationshipStatus' to the contact properties?

leekahung: Was thinking about utilizing the contact list for those, especially with relationship status. However, we could potentially include the option to set relationships in the sign up flow on the sign up page.

src/constants/relationship_status.js
```diff
@@ -0,0 +1,8 @@
+const RELATIONSHIP_STATUS = {
+  BLANK: '',
+  ACTIVE: 'active',
+  ARCHIVE: 'archive',
+  OTHER: 'other'
```
leekahung: Could we get the keys camelCased as well? Think that should be sufficient.

src/components/Modals/AddContactModal.jsx
```diff
@@ -151,39 +154,56 @@ const AddContactModal = ({ addContact, showAddContactModal, setShowAddContactMod
...
+            value={relationshipStatus}
+            label="RelationshipStatus"
+            onChange={(e) => setRelationshipStatus(e.target.value)}
+            fullWidth
+          >
+            <MenuItem value={RELATIONSHIP_STATUS.BLANK} />
+            <MenuItem value={RELATIONSHIP_STATUS.ACTIVE}>Active</MenuItem>
+            <MenuItem value={RELATIONSHIP_STATUS.ARCHIVED}>Archived</MenuItem>
+            <MenuItem value={RELATIONSHIP_STATUS.ETC}>etc</MenuItem>
+          </Select>
```
leekahung: Personally, I think we should revert the modal back to what we had before until we're sure where to include the status updates for "relationship" (or "relationship type", which I think is the better wording) and "relationship status".

Like what I've mentioned earlier, "relationship status" (active/archive/other) feels like it should belong to the contacts list rather than in the add contacts modal and "relationship type" feels like it should belong to the sign up page, although, having it in the contacts list would also work.

If we were to include these `<Select>` components in this PR, I think should have them placed there instead.

src/constants/relationships.js
```diff
@@ -0,0 +1,8 @@
+const RELATIONSHIPS = {
```
leekahung: Rather than `RELATIONSHIPS`, I think `RELATIONSHIP_TYPE` would probably be better. It's more targeted as well.

leekahung (changes requested):
Made a few more change requests to this PR, think it should be good after getting those addressed.

leekahung:
Also, there seems to be linting errors. Do you run the lint or prettier in your branch? It doesn't seem like prettier is being picked up in your PR.

ofu997:
I ran ```npm run prettier:test``` without error.

leekahung:
> I ran ```npm run prettier:test``` without error.

We don't have a script for `prettier:test`, see screenshot.

Screen Shot 2023-11-15 at 13 06 15

`npm run prettier:run` is the script for running Prettier to update formatting. `npm run prettier:check` is the script for running Prettier to check formatting.

ofu997:
Weird
image

leekahung:
> Weird image

You need to run `npm run prettier:run`. For `npm run prettier:check` it's used by the CI to double check if formatting was missing.

Note the error found in the check:

Screen Shot 2023-11-16 at 21 17 50 (2)

src/components/Modals/AddContactModal.jsx
leekahung, line 1: Was wondering if we could simply revert this file back to the version from Development at this time? Other than that, I think it's good to go for potential merge.

leekahung:
Hey @ofu997 ! Wanted to check on the progress of this PR.

Believe we could start getting the discussions for where to place these new fields with the UX team once we got this merged and start an issue opened for implementation.

src/components/Modals/AddContactModal.jsx
```diff
@@ -20,9 +20,9 @@ import useNotification from '@hooks/useNotification';
 import { FormSection } from '../Form';
 
 /**
- * @memberof Modals
+ * @memberof Contcts
```
leekahung: This should be `@memberof Modals`

src/components/Modals/AddContactModal.jsx
```diff
@@ -20,9 +20,9 @@ import useNotification from '@hooks/useNotification';
...
  * @name renderWebId
- * @param {string} username - Username to convert into a webId
+ * @param {string} username - username to convert into a webId
```
leekahung: I don't think the correction is needed here.

src/components/Modals/AddContactModal.jsx
```diff
@@ -34,12 +34,12 @@ const renderWebId = (username) => {
...
+ * @memberof Contacts
  * @name AddContactModal
- * @param {object} props - React props
- * @param {Function} props.addContact - Function to add a contact
- * @param {boolean} props.showAddContactModal - Whether to display modal or not
- * @param {Function} props.setShowAddContactModal - Toggle modal
+ * @param {object} props - react props
+ * @param {Function} props.addContact  - function to add a contact
+ * @param {boolean} props.showAddContactModal - whether to display modal or not
+ * @param {Function} props.setShowAddContactModal - toggle modal
```
leekahung: The code here don't need to be changed.

src/components/Modals/AddContactModal.jsx
```diff
@@ -40,7 +40,6 @@ const renderWebId = (username) => {
  * @param {Function} props.addContact - Function to add a contact
  * @param {boolean} props.showAddContactModal - Whether to display modal or not
  * @param {Function} props.setShowAddContactModal - Toggle modal
- * @returns {React.JSX.Element} - The Add Contact Modal
```
leekahung: It looks like JSDoc return statement has been removed by mistake as pointed out by the linter.

```
* @returns {React.JSX.Element} - The Add Contact Modal
```

src/components/Modals/AddContactModal.jsx
leekahung: Alright, think we can just leave this for now. There are other JSDocs that still uses lowercasing, so I think that's fine for now

leekahung (approved):
Think it's good for merging. Think we could start an open discussion about how to utilize these new properties now that those statuses are now available.

timbot1789 (changes requested):
The structure of the "role" property seems incorrect. We're storing a string to it, but the definition on schema.org describes an object (technically a "thing") with a few different attributes.

The documentation actually links to a blog post about it: http://blog.schema.org/2014/06/introducing-role.html

ofu997:
I'll make it https://schema.org/roleName

timbot1789:
> I'll make it https://schema.org/roleName

That looks like a good solution to me

leekahung:
Hey @ofu997! Was wondering if this is still being worked on. If you are, let us know. Thanks!

ofu997:
> Hey @ofu997! Was wondering if this is still being worked on. If you are, let us know. Thanks!

Hi Kahung, 
It looks like I said "I'll make it https://schema.org/roleName" but didn't make a commit. I'm at work RN, so If it seems pretty much ready you can finish this up.

timbot1789 (approved):

PR #510, opened 2023-11-12
Use `getByRole` where appropriate and improve some test magic numbers

### LINKED ISSUE #449: Enhancement: Use `getByRole` or `testId` in tests over `getByLabelText`

We currently have tests that use `getByLabelText` in the tests when other selector methods would suffice.

We should convert `getByLabelText` tests to something less fragile like `getByRole/TestId`.

### Rationale:

Relying on `LabelText` for anything other than rare cases where it is necessary make tests fragile and possibly can imply that adding labels is the way to select elements.

Using `getByRole` in particular can also reinforce testing of functionality. For instance, in testing that a button maintains the expected accessible name, selecting it via `Role` of `button` helps reinforce that we are indeed selecting for a button, whereas selecting by `LabelText` can result in a false positive if the element loses the `button` role. This could help stop significant accessibility regressions.

Because label texts can change but functionality should change much more seldom, this also makes maintenance of the tests much easier.

For instances where selecting by role is applicable, do so. 

In instances where that doesn't work, giving elements a `data-testid` will give some of the same benefits and not rely so much on implementation details.

### Example of existing test

```diff
it('renders NavbarLoggedOut when user is not logged in', () => {
-    const { getByLabelText } = render(
+    const { getByRole } = render(
       <SessionContext.Provider value={{ session: { info: { isLoggedIn: false } } }}>
         <BrowserRouter>
           <NavBar />
         </BrowserRouter>
       </SessionContext.Provider>
     );
-    const loginButton = getByLabelText('Login Button');
+    const loginButton = getByRole('button');

     expect(loginButton).not.toBeNull();
   });
```

See https://testing-library.com/docs/queries/byrole/ for more info on more specific selection using the `name` option.

milofultz:
Resolves #449 
 
**1.** Uses `getByRole` where appropriate and updates components with appropriate roles where missing.
**2.** Removed some "magic" variables in tests.

### DIFF
--- a/src/components/NavBar/NavbarDesktop.jsx
+++ b/src/components/NavBar/NavbarDesktop.jsx
@@ -69,7 +69,7 @@ const NavbarDesktop = ({ setShowConfirmation }) => {
           </Link>
           <NavbarLinks aria-label="navigation links" />
           <Box sx={{ flexGrow: 1 }} />
-          <Box aria-label="menu" sx={{ display: { xs: 'none', md: 'flex' } }}>
+          <Box sx={{ display: { xs: 'none', md: 'flex' } }} data-testid="navbar-menu">
             {/* messages icon */}
             <IconButton
               component={NavLink}
--- a/src/components/NavBar/NavbarDesktop.jsx
+++ b/src/components/NavBar/NavbarDesktop.jsx
@@ -69,7 +69,7 @@ const NavbarDesktop = ({ setShowConfirmation }) => {
           </Link>
           <NavbarLinks aria-label="navigation links" />
           <Box sx={{ flexGrow: 1 }} />
-          <Box sx={{ display: { xs: 'none', md: 'flex' } }} data-testid="navbar-menu">
+          <Box sx={{ display: { xs: 'none', md: 'flex' } }} role="group">
             {/* messages icon */}
             <IconButton
               component={NavLink}
--- a/test/components/Contacts/ContactsListTable.test.jsx
+++ b/test/components/Contacts/ContactsListTable.test.jsx
@@ -68,5 +68,5 @@ it('sorts clients by familyName', () => {
   const client1 = getByRole('cell', { name: 'Zeigler' });
   const client2 = getByRole('cell', { name: 'Builder' });
 
-  expect(client1.compareDocumentPosition(client2)).toBe(2);
+  expect(client1.compareDocumentPosition(client2)).toBe(Node.DOCUMENT_POSITION_PRECEDING);
 });
--- a/test/components/NavBar/NavBar.test.jsx
+++ b/test/components/NavBar/NavBar.test.jsx
@@ -30,7 +30,7 @@ describe('resize tests', () => {
   it('renders NavbarDesktop when user is logged in on larger screen device', () => {
     window.matchMedia = createMatchMedia(1200);
 
-    const { getByTestId } = render(
+    const { getByRole } = render(
       <SessionContext.Provider value={{ session: { info: { isLoggedIn: true } } }}>
         <BrowserRouter>
           <NavBar />
@@ -38,7 +38,7 @@ describe('resize tests', () => {
       </SessionContext.Provider>
     );
 
-    const iconMenu = getByTestId('navbar-menu');
+    const iconMenu = getByRole('group');
 
     expect(iconMenu).not.toBeNull();
   });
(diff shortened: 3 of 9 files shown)


src/components/NavBar/NavbarDesktop.jsx
```diff
@@ -69,7 +69,7 @@ const NavbarDesktop = ({ setShowConfirmation }) => {
           </Link>
           <NavbarLinks aria-label="navigation links" />
           <Box sx={{ flexGrow: 1 }} />
-          <Box aria-label="menu" sx={{ display: { xs: 'none', md: 'flex' } }}>
+          <Box sx={{ display: { xs: 'none', md: 'flex' } }} role="group">
```
milofultz, line 72: Label isn't required, as each link will be announced when tabbed into it and `menu` doesn't provide other meaningful info.

test/components/Contacts/ContactsListTable.test.jsx
```diff
@@ -68,5 +68,5 @@ it('sorts clients by familyName', () => {
   const client1 = getByRole('cell', { name: 'Zeigler' });
   const client2 = getByRole('cell', { name: 'Builder' });
 
-  expect(client1.compareDocumentPosition(client2)).toBe(2);
+  expect(client1.compareDocumentPosition(client2)).toBe(Node.DOCUMENT_POSITION_PRECEDING);
```
milofultz, line 71: https://developer.mozilla.org/en-US/docs/Web/API/Node/compareDocumentPosition

test/components/NavBar/NavBar.test.jsx
```diff
@@ -30,31 +30,31 @@ describe('resize tests', () => {
...
+    const { getByRole } = render(
       <SessionContext.Provider value={{ session: { info: { isLoggedIn: true } } }}>
         <BrowserRouter>
           <NavBar />
         </BrowserRouter>
       </SessionContext.Provider>
     );
 
-    const iconMenu = getByLabelText('menu');
+    const iconMenu = getByRole('group');
```
xscottxbrownx, line 41: Is there something that by default has a role of group? I'm not seeing anything directly expressed as such.

This is a learning question for me, not questioning this logic.

test/components/NavBar/NavbarMobile.test.jsx
```diff
@@ -6,15 +6,15 @@ import NavbarMobile from '../../../src/components/NavBar/NavbarMobile';
...
+  const { queryByRole } = render(
     <BrowserRouter>
       <NavbarMobile />
     </BrowserRouter>
   );
 
   const logo = queryByRole('img', { name: /logo$/ });
-  const navLinks = queryByLabelText('navigation tabs');
-  const hamburgerMenu = queryByLabelText('mobile menu');
+  const navLinks = queryByRole('tablist');
```
xscottxbrownx, line 16: Same question here... curious where this role is found - or if it's a default of MUI tabs?

test/pages/Signup.test.jsx
```diff
@@ -49,8 +49,8 @@ describe('Signup Page', () => {
         }
       }
     };
-    const { getByText } = render(<MockSignupContexts session={sessionObj} />);
-    expect(getByText('Register for PASS', { exact: false })).not.toBeNull();
+    const { getByRole } = render(<MockSignupContexts session={sessionObj} />);
+    expect(getByRole('heading')).not.toBeNull();
```
xscottxbrownx: And this works because either component that can be rendered have a heading?  If that is true, is that too generic of a test?

xscottxbrownx (commented):
Questions to help my understanding...

timbot1789 (approved):
LGTM! I like PRs that remove more code than they add.

test/components/NavBar/NavBar.test.jsx
milofultz: Unfortunately, nothing has a `group` role by default. `div`'s are often used for grouping, but are otherwise devoid of any native `role`.

test/components/NavBar/NavbarMobile.test.jsx
milofultz: Yeah this is part of the [React tabs MUI component](https://mui.com/material-ui/react-tabs/). I was digging around looking to see if roles were applied and found it with the accessibility inspector in Firefox.

Screen Shot 2023-11-14 at 7 31 41 PM

test/pages/Signup.test.jsx
milofultz: True, I should probably add a `name` option to this to make it more specific. Thanks!

milofultz: Got it in there 👍

PR #512, opened 2023-11-12
Improve labels in `NavbarDesktop`

milofultz:
I made this without an issue because it's so minor.
 
**1.** Remove "new" from "show new messages/notifications" as both of these interactions will show _all_ messages/notifications, not just the new ones (at least that is my understanding).
**2.** Capitalize first word of labels for consistency.


xscottxbrownx (approved):
good enhancement

leekahung (approved):
Yeah. Think I'm alright with this update as well.

Issue #509, opened 2023-11-12
[Enhancement] - Migrate Message utility functions to custom hook with React Query


leekahung:
### Describe the Current Behavior/Feature:

With React Query in place, we could start migrating some of the utility functions we've previously built to a custom hook.

### Proposed Implementation (if applicable):

Move some of the functions from `session-helper`, `session-core`, or `MessageContext` over to a new hook called `useMessageList`.

Depending on how heavy the refactoring load would be, would likely implement this in parts.

PR #514, opened 2023-11-13
[Enhancement/Refactor]-  Issue 509/create new use message list hook

leekahung:
## This PR:

Partially addresses #509. This PR aims to migrate over/refactor the existing workflow for messages loading and rendering from utils and React Context to React Query.

Unit tests that's related to existing changes has been updated. A new test file has been made, mainly for the new custom hook `useMessageList`.

https://github.com/codeforpdx/PASS/assets/14917816/5e0170ba-d72c-46e9-b1c2-d8c3fc79db0f

https://github.com/codeforpdx/PASS/assets/14917816/3e62a8f1-1085-4523-af1b-42abf4bab5bc

Future steps for this will include (in no particular order):

- Updates to other Message functionalities in utils similar to what is done here, but for message building and message sending
- Splitting Inbox/Outbox into separate pages

**1.**    Issue #509 will be on-going even after this is merged (see Future Steps above)


src/hooks/useMessageList.js
```diff
@@ -0,0 +1,98 @@
...
+    title,
+    uploadDate,
+    readStatus,
+    sender,
+    senderWebId,
+    recipient
+  };
+};
+
+const useMessageList = (messageType) => {
```
leekahung, line 145: Think I'll include some JSDoc here as well to document what this hook is used for

timbot1789 (commented):
I'm a bit confused by this statement
> Had tried using just React Query for isLoading or isFetching, but the UI appears smoother with React Context and does not cause any noticeable re-renders of the message cards, so it remained with React Context.

React Query is built on top of a react context, so this shouldn't make a difference. If there's UI smoothness issues, they probably come from how the UI itself is handling the different loading states.

I notice you're hitting the network multiple times whenever you update a message.  This may be why you were having weird behavior, and it shouldn't be necessary. The react query `update` mutations can handle all of that for you. You don't need to manually refetch in the components either. The purpose of these hooks is to hide the network from the components. Put all the networking logic in the hooks, and expose simple interfaces to the components.

You may find it simpler not to handle both the inbox and the outbox in a single hook. Maybe try separating the inbox logic into its own `useInbox` hook, and the outbox into a `useOutbox` hook. I think you'll be able to get a clearer design that way.

leekahung:
> I'm a bit confused by this statement
> 
> > Had tried using just React Query for isLoading or isFetching, but the UI appears smoother with React Context and does not cause any noticeable re-renders of the message cards, so it remained with React Context.
> 
> React Query is built on top of a react context, so this shouldn't make a difference. If there's UI smoothness issues, they probably come from how the UI itself is handling the different loading states.

Was a bit confused myself about this actually, when I first noticed this (see clip when I've used `isFetching` again), but I think I know what might be happening. Since clicking on unread message previews updates the read status for message Thing on Solid, I that's what's causing the loading screen to pop up. The second press was fine because it's already read and `updateMessageReadStatus` is no longer going to trigger a state update.

https://github.com/codeforpdx/PASS/assets/14917816/a8c768f4-8992-4642-830e-38f04c2529f0

> I notice you're hitting the network multiple times whenever you update a message. This may be why you were having weird behavior, and it shouldn't be necessary.

Think I've managed to fix this behavior by moving the condition for `readStatus` along with `folderType === 'Inbox'` in the update I'm coding up now so `updateMessageReadStatus` won't be triggered at all if the message had already been read.

>The react query `update` mutations can handle all of that for you. You don't need to manually refetch in the components either. The purpose of these hooks is to hide the network from the components. Put all the networking logic in the hooks, and expose simple interfaces to the components.

Oh! That would be nice. I'll see if I could get mutations working for this. I simply wanted to update the readStatus for that one message. Like what's shown above in the clip, but without the loading screen appearing.

> You may find it simpler not to handle both the inbox and the outbox in a single hook. Maybe try separating the inbox logic into its own `useInbox` hook, and the outbox into a `useOutbox` hook. I think you'll be able to get a clearer design that way.

I was planning on splitting up Inbox and Outbox into their own routes and pages in a later PR.

leekahung:
Hey @timbot1789, think I've managed to get the React Query mutation working! Also got the loadMessages being handled directly by `isFetching` now without triggering the loading animation unless it's manually refreshed via the refresh button or when loading into the `Messages` page for the first time.

src/components/Messages/MessageFolder.jsx
leekahung, line 1: Will break up Inbox and Outbox in a follow-up PR to their own routes but will still be under messages (i.e. `/messages/inbox` and `/message/outbox`).

src/hooks/useMessageList.js
```diff
@@ -0,0 +1,219 @@
...
+    setStoredDataset(myDataset);
+    return parse(myDataset);
+  };
+
+  const { isLoading, isError, error, data, isSuccess, refetch, isFetching } = useQuery({
+    queryKey: [containerUrl?.toString()],
+    queryFn: fetchDocument
+  });
+
+  const addMutation = useMutation({
```
leekahung, line 199: Had custom `fetchDocument` and `addMutation` functions from `usRdfCollection.js` included for this use case.

src/pages/Messages.jsx
```diff
@@ -22,51 +20,29 @@ import { MessageButtonGroup, MessageFolder } from '../components/Messages';
  */
 const Messages = () => {
   localStorage.setItem('restorePath', '/messages');
-  const { podUrl } = useContext(SignedInUserContext);
-  const { session } = useSession();
+  const { updateMessageCountState } = useContext(MessageContext);
   const {
```
leekahung, line 22: The two hooks in `Message.jsx` will be moved into their own pages in a follow-up PR, one for "Inbox", one for "Outbox".

src/pages/Messages.jsx
```diff
@@ -22,51 +20,29 @@ import { MessageButtonGroup, MessageFolder } from '../components/Messages';
...
-    const messageList = folderType === 'Inbox' ? inboxList : outboxList;
-    const messagesInSolid = await getMessageTTL(session, folderType, messageList, podUrl);
-    messagesInSolid.sort((a, b) => b.uploadDate - a.uploadDate);
-    if (folderType === 'Inbox') {
-      setInboxList(messagesInSolid);
-    } else {
-      setOutboxList(messagesInSolid);
-    }
-    setLoadMessages(false);
+  const handleMessageRefresh = async () => {
```
leekahung: Probably want to create condition to trigger just one of the manual refetches rather than both. A simple if/else should do.

src/pages/Messages.jsx
```diff
@@ -22,51 +20,32 @@ import { MessageButtonGroup, MessageFolder } from '../components/Messages';
  */
 const Messages = () => {
   localStorage.setItem('restorePath', '/messages');
-  const { podUrl } = useContext(SignedInUserContext);
-  const { session } = useSession();
+  const { updateMessageCountState } = useContext(MessageContext);
```
timbot1789: Is this still needed? This value is directly computed from the inbox (and doesn't seem super expensive to calculate), so you should be able to store the computation locally and then you don't need the context at all here.

src/components/Modals/NewMessageModal.jsx
```diff
@@ -14,9 +14,9 @@ import Typography from '@mui/material/Typography';
 import useMediaQuery from '@mui/material/useMediaQuery';
 import { useTheme } from '@mui/material/styles';
 // Utility Imports
-import { sendMessageTTL, getMessageTTL } from '@utils';
+import { sendMessageTTL } from '@utils';
```
timbot1789, line 17: We should be able to move this function into the hook as well. Calling "add" (maybe renamed "send" or something similar) on the outbox should be able to handle it from a UI perspective. Is that something to do here, or does it make more sense to do that in a followup PR?

timbot1789 (commented):
I like this overall. Just a couple small comments. Nice work.

src/hooks/useMessageList.js
timbot1789: We can probably make some enhancements to `useRdfCollection` to better support this situation. But that can wait for a followup PR.

src/hooks/useMessageList.js
```diff
@@ -0,0 +1,219 @@
...
+  const { session, podUrl } = useSession();
+  const { fetch } = session;
+  const [storedDataset, setStoredDataset] = useState(null);
+  const queryClient = useQueryClient();
+  const containerUrl = podUrl && new URL(`PASS/${messageType}/`, podUrl).toString();
+  let messageList = [];
+
+  const parse = async (data) => {
+    const solidDataset = getThingAll(data);
+    const allMessagesDataset = solidDataset.filter((thing) => thing.url.endsWith('ttl'));
```
timbot1789, line 155: Ah so the messages are all stored as separate `.ttl` documents within a pod. We should be able to make the `useRdfCollection` hook support this case. I'll make a note and follow up.

src/components/Modals/NewMessageModal.jsx
leekahung: Yeah, I think that would be good too. Was thinking of have them moved over in follow-up PR.

src/pages/Messages.jsx
leekahung: That's true. I'll try out a commit for this

src/hooks/useMessageList.js
leekahung: Got it, think that would be great.

src/pages/Messages.jsx
leekahung: Think I've manage to get this working. Got MessageContext replaced as well in the process. Just need that unit test to get updated.

bingeboy:
I'm running this on my local and I'm noticing that when I click on messages from the top nav the Messages section loads with "No Messages Found" while the messages are being retrieved. I also have 20 pages of messages so that might be an edge case.

src/hooks/useMessageList.js
bingeboy: Reading up on https://tanstack.com/query/v3/docs/react/guides/updates-from-mutation-responses

leekahung:
> I'm running this on my local and I'm noticing that when I click on messages from the top nav the Messages section loads with "No Messages Found" while the messages are being retrieved. I also have 20 pages of messages so that might be an edge case.

Oh that's is strange. I'll take a look.

leekahung:
> I'm running this on my local and I'm noticing that when I click on messages from the top nav the Messages section loads with "No Messages Found" while the messages are being retrieved. I also have 20 pages of messages so that might be an edge case.

Not really seeing it in my case (see clip), so not sure what might be going on.

https://github.com/codeforpdx/PASS/assets/14917816/ab84d8b1-7a45-48af-bef1-785a3cc7eb3f

timbot1789:
> I'm running this on my local and I'm noticing that when I click on messages from the top nav the Messages section loads with "No Messages Found" while the messages are being retrieved. I also have 20 pages of messages so that might be an edge case.

This sounds like a performance bottleneck that we're not set up to handle yet. We haven't implemented any kind of server-side pagination, so the network load when fetching multiple pages of records is quite hefty. I don't think it's an issue we can fix in this PR.

src/components/Messages/MessagePreview.jsx
```diff
@@ -38,18 +34,13 @@ import { NewMessageModal } from '../Modals';
...
   const [showModal, setShowModal] = useState(false);
-  const { session } = useSession();
-  const { podUrl } = useContext(SignedInUserContext);
-  const { inboxList, setInboxList } = useContext(MessageContext);
+  const { add: updateReadStatus } = useMessageList('Inbox');
 
   const handleClick = async () => {
     setShowContents(!showContents);
-    if (folderType === 'Inbox') {
+    if (folderType === 'Inbox' && message.readStatus === false) {
```
timbot1789: !message.readStatus
```suggestion
    if (folderType === 'Inbox' && !message.readStatus) {
```

src/components/NavBar/NavMenu.jsx
```diff
@@ -52,7 +53,12 @@ const NavMenu = ({
   const theme = useTheme();
   const isSmallScreen = useMediaQuery(theme.breakpoints.down('sm'));
   const { setContact } = useContext(DocumentListContext);
-  const { numUnreadMessages } = useContext(MessageContext);
+  const { data } = useMessageList('Inbox');
+  const [numUnreadMessages, setNumUnreadMessages] = useState(0);
+
+  useEffect(() => {
```
timbot1789: You can compute the number of unread messages in the the `useMessageList` hook, or derive it directly where you display it in the component based on current state. Then you don't need to add this `useEffect` in consuming components, and you don't need to worry about the value in `numUnreadMessages` falling out of sync with what's in the hook.

src/components/NavBar/NavbarDesktop.jsx
```diff
@@ -32,7 +34,12 @@ import NavMenu from './NavMenu';
  */
 const NavbarDesktop = ({ setShowConfirmation }) => {
   const theme = useTheme();
-  const { numUnreadMessages } = useContext(MessageContext);
+  const { data } = useMessageList('Inbox');
+  const [numUnreadMessages, setNumUnreadMessages] = useState(0);
+
+  useEffect(() => {
+    setNumUnreadMessages(data?.reduce((a, m) => (!m.readStatus ? a + 1 : a), 0));
```
timbot1789: See above comment

src/contexts/SignedInUserContext.jsx
```diff
@@ -64,7 +64,9 @@ export const SignedInUserContextProvider = ({ children }) => {
         setProfileData(fetchedProfileData);
         await Promise.all([
           createPASSContainer(session, fetchedPodUrl, 'Documents'),
-          createPASSContainer(session, fetchedPodUrl, 'Profile')
+          createPASSContainer(session, fetchedPodUrl, 'Profile'),
+          createPASSContainer(session, podUrl, 'Outbox'),
```
timbot1789: You shouldn't need to create these containers in advance. You should be able to create them in the hook the first time the user tries to access the record.

See how we handle that in the `useRdfCollection` here: https://github.com/codeforpdx/PASS/blob/9b578d7a1ff96aa1f0eeb332e76e2997bc74d9e4/src/hooks/useRdfCollection.js#L51

src/components/Messages/MessagePreview.jsx
leekahung: Got this switched to using `!`

src/components/NavBar/NavMenu.jsx
leekahung: Ah, okay. Just tried it out without the `useEffect` and directly computing `numUnreadMessages` with:

```
const numUnreadMessages = data?.reduce((a, m) => (!m.readStatus ? a + 1 : a), 0);
```

https://github.com/codeforpdx/PASS/assets/14917816/fe93762a-f58b-4d65-b48c-9ce1760bb4b0

I'll replace the one in `NavbarDesktop.jsx` as well.

src/components/NavBar/NavbarDesktop.jsx
leekahung: Got this updated

src/contexts/SignedInUserContext.jsx
leekahung: I think I could remove the one for Outbox since it's internally for the user so users should still be able to access them without needing to generate them.

For the Inbox, I'll try switching up the function to just generate the necessary `.acl` file without the need for making the container so others could send message to them. Without it, could only send messages to themselves.

leekahung: Alright, think I've managed to get this working. I've got the new generateACL function nested inside the `fetchDocument` function for `useMessageList`. It'll only trigger when logging in for the first time right after the `saveSolidDatasetAt` function.

bingeboy:
> > I'm running this on my local and I'm noticing that when I click on messages from the top nav the Messages section loads with "No Messages Found" while the messages are being retrieved. I also have 20 pages of messages so that might be an edge case.
> 
> Not really seeing it in my case (see clip), so not sure what might be going on.
> Screen.Recording.2023-11-15.at.12.11.43.mov

I've pulled latest code and issue is resolved. Now showing "Loading Messages ..." 👍

xscottxbrownx:
I haven't tested this branch, but when creating a message - does the modal still behave like it did in very top video here?

The delay between clearing the form and closing the modal.

If so, is that necessary for some reason?

src/components/Modals/NewMessageModal.jsx
```diff
@@ -107,12 +107,7 @@ const NewMessageModal = ({ showModal, setShowModal, oldMessage = '', toField = '
         }, 2000);
```
bingeboy, line 107: @xscottxbrownx looks like this `setTimeout()` is causing that.

bingeboy:
This looks good to me. Cleans up Messages. Separates concerns between UI and Solid data. Nice work 👍

src/components/Modals/NewMessageModal.jsx
leekahung: Looks like the `setTimeout` was part of Development too. I don't remember the original reason, though I'm guessing it's related to one being async the other being synchronous, which was why they're placed inside a `finally` block. I've only touched on code related to the message lists here in this PR.

Think we could open up a new PR to handle this, we do have a problem with the modal closing regardless of error. We could get that resolved as well by how we handle the two set functions inside the `finally` block, whether we move them out or change the functions, etc. That'll be good.

https://github.com/codeforpdx/PASS/assets/14917816/fc7cf06d-25a4-4ab9-ad73-9b846273ebc7

timbot1789 (approved):

PR #515, opened 2023-11-13
Issue 513: civic profile housing [Part 1]

timbot1789:
First part of a housing section for the civic profile. Uploading this in order to fix some bugs with the `useCivicProfile` hook, and to provide an example of how to use it.

This is a simple form right now. It just allows you to upload an address. It hasn't gone through any UI design pass, and we will need a somewhat sophisticated design to figure out simple ways to ask about a person's homelessness history.

Also added a new devtool to allow you to start up both the pod server and the react server at the same time. Run `npm start` and you'll launch both. We use a small package called `concurrently` to manage both commands in one process.


src/hooks/useCivicProfile.js
```diff
@@ -10,18 +10,83 @@ import { RDF_PREDICATES } from '@constants';
 import useSession from './useSession';
 import useRdfCollection from './useRdfCollection';
 
+const prefix = 'urn:hud:hmis:owl#';
+
+const civicProfileConfig = {
+  firstName: {
+    type: 'string',
+    predicate: 'urn:hud:hmis:owl#FirstName'
```
andycwilliams: Can't `prefix` be added as a template literal for these four as well?

src/hooks/useCivicProfile.js
```diff
@@ -10,18 +10,83 @@ import { RDF_PREDICATES } from '@constants';
...
+    predicate: `${prefix}TimeToHousingLoss`
+  }
+};
+
+const convertDataToThing = (data, thingName, config) => {
+  let thingInProgress = buildThing(createThing({ name: thingName }));
+  Object.keys(data).forEach((key) => {
+    const definition = config[key];
+    if (!definition) return;
+    try {
```
leekahung: I don't think we need a try/catch here, there isn't any asynchronous calls in this particular function.

src/hooks/useCivicProfile.js
```diff
@@ -10,18 +10,83 @@ import { RDF_PREDICATES } from '@constants';
...
+const civicProfileConfig = {
+  firstName: {
+    type: 'string',
+    predicate: 'urn:hud:hmis:owl#FirstName'
+  },
+  lastName: {
+    type: 'string',
+    predicate: 'urn:hud:hmis:owl#LastName'
+  },
+  dateOfBirth: {
```
leekahung, line 24: Think I'll start refactoring DOB out of what we have now in the public profile in another PR.

src/components/CivicProfileForms/HousingInfo.jsx
```diff
@@ -1,8 +1,81 @@
 // React Imports
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 // MUI Imports
-import Typography from '@mui/material/Typography';
+import { TextField, Button } from '@mui/material';
+import { useCivicProfile } from '@hooks';
 
-const HousingInfo = () => <Typography>Housing Info</Typography>;
+const HousingInfo = () => {
```
andycwilliams, line 14: Will these forms require JSDocs?

andycwilliams:
> This is a simple form right now. It just allows you to upload an address. It hasn't gone through any UI design pass, and we will need a somewhat sophisticated design to figure out simple ways to ask about a person's homelessness history.

Are these being left for [Part 2]? What will that PR entail?

src/components/CivicProfileForms/HousingInfo.jsx
leekahung: Think it'll be good to include them, perhaps a small snippet of what fields users likely be filling out in these forms

src/hooks/useCivicProfile.js
timbot1789: Sure

src/hooks/useCivicProfile.js
timbot1789: I'm not sure what your comment about async means. Try...catch and async are orthogonal concepts. They don't really relate to each other.

Either way I refactored this section not to need it.

src/components/CivicProfileForms/HousingInfo.jsx
timbot1789: Added

timbot1789:
> > This is a simple form right now. It just allows you to upload an address. It hasn't gone through any UI design pass, and we will need a somewhat sophisticated design to figure out simple ways to ask about a person's homelessness history.
> 
> Are these being left for [Part 2]? What will that PR entail?

I haven't really thought about it too much. I'd like to be able to include the other 3 fields referenced in the PR. But we also need a separate question "Are you currently homeless?" because that would exclude the field "How long until you lose your house." There are also additional questions we can ask about their history in shelters. How worthwhile those would be I don't know for sure.

src/hooks/useCivicProfile.js
leekahung: Ah okay, all good.👍 

Thought try catch was mainly used for async await and promises, but I guess it's still good to use for any operations that throws an error object.

andycwilliams:
> I haven't really thought about it too much. I'd like to be able to include the other 3 fields referenced in the PR. But we also need a separate question "Are you currently homeless?" because that would exclude the field "How long until you lose your house." There are also additional questions we can ask about their history in shelters. How worthwhile those would be I don't know for sure.

Makes sense.

So would updating the UI be next or wait until later on?

src/components/CivicProfileForms/HousingInfo.jsx
```diff
@@ -1,8 +1,88 @@
...
+ * @name HousingInfo
+ * @returns {React.JSX.Element} The HousingInfo Component
+ */
+const HousingInfo = () => {
+  const { data, add, isSuccess } = useCivicProfile();
+  const [formData, setFormData] = useState({
+    lastPermanentStreet: '',
+    lastPermanentCity: '',
+    lastPermanentState: '',
+    lastPermanentZIP: ''
```
andycwilliams, line 20: Are all inputs supposed to be string? Not sure if HMIS requires that or anything. If not, we can maybe change ZIP  to a number (as well as adding `type="number"` to its `TextField`).

timbot1789:
> > I haven't really thought about it too much. I'd like to be able to include the other 3 fields referenced in the PR. But we also need a separate question "Are you currently homeless?" because that would exclude the field "How long until you lose your house." There are also additional questions we can ask about their history in shelters. How worthwhile those would be I don't know for sure.
> 
> Makes sense.
> 
> So would updating the UI be next or wait until later on?

I don't think there's any rush. We can pass this kind of thing over to the UI team as well and see what their thoughts are. Ultimately we as a project team need to figure out what data points we want to capture, and what use we have for them, then send that over to both UI and dev.

My goal in the meantime is to get the Civic profile to a point where it has the information you need to fill out an I9 form or other employment forms. An address is a necessary part of that.

src/components/CivicProfileForms/HousingInfo.jsx
timbot1789: They are. All the data types are defined in the HMIS ontology. We can't perfectly comply with all of them ATM, but recording most things as strings is a good way to get close (and make sure we can convert them to better types later). 

And the zip code may look like a number, but it is not a number. If you can't do math with it, it's not a number. Zip codes can, and do, have '+' signs in them, which you can't put in a number.

andycwilliams:
> I don't think there's any rush. We can pass this kind of thing over to the UI team as well and see what their thoughts are. Ultimately we as a project team need to figure out what data points we want to capture, and what use we have for them, then send that over to both UI and dev.
> 
> My goal in the meantime is to get the Civic profile to a point where it has the information you need to fill out an I9 form or other employment forms. An address is a necessary part of that.

I assume you meant UX that first time rather than UI? They have asked what the difference(s) will be between the civic profile and the user profile in terms of design. If they're intended to be mostly the same then that would save them some effort.

leekahung (approved):
Just took another look at the changes, think it's good for now.

The simple text fields should be fine. We could further improve them later in future PRs once we get designs from UX.

leekahung:
> They have asked what the difference(s) will be between the civic profile and the user profile in terms of design. If they're intended to be mostly the same then that would save them some effort.

I think it's safe to say that the "Civic Profile" is the personal profile of people where we'll have sensitive information like housing, financial, etc. For what we have as "Profile" or user profile is something akin to an online account for a web service, where only things like names or emails are available and can be seen publicly.

Think "Account" would be a better term for what we have as "Profile" right now and we could actually have "Civic Profile" be named as "Profile" proper.

andycwilliams:
> I think it's safe to say that the "Civic Profile" is the personal profile of people where we'll have sensitive information like housing, financial, etc. For what we have as "Profile" or user profile is something akin to an online account for a web service, where only things like names or emails are available and can be seen publicly.
> 
> Think "Account" would be a better term for what we have as "Profile" right now and we could actually have "Civic Profile" be named as "Profile" proper.

I wonder if "Profile" is a bit too analogous to "Account" in peoples' minds. Maybe something as simple as making "Profile" into "Profiles" would help differentiate the two.

leekahung:
> I wonder if "Profile" is a bit too analogous to "Account" in peoples' minds. Maybe something as simple as making "Profile" into "Profiles" would help differentiate the two.

Not having them both with the same base wording, it'll would be too confusing, more so than now. Even if "Profile" is "too analogous" to "Account", we could still make a distinction that they should be two separate things.

Personally, I do see the distinction with "Account" related more on the administrative side of things, namely related with PASS, with "Profile" related more to information related to the user themselves.

For example, it's like having a bank "Account" which is associated with the "bank" (like account number and names), and user information (i.e. "Profile") which is associate with you (like mailing address, phone numbers, email addresses) which banks asks for after you have an account.

src/constants/HMIS_ONTOLOGY_VALUES.js
```diff
@@ -1,8 +1,17 @@
+const prefix = 'urn:hud:hmis:owl#';
+
 const HMIS_ONTOLOGY_VALUES = {
   legalFirstName: 'urn:hud:hmis:owl#FirstName',
   legalLastName: 'urn:hud:hmis:owl#LastName',
   legalDOB: 'urn:hud:hmis:owl#DOB',
-  legalGender: 'urn:hud:hmis:owl#Gender'
+  legalGender: 'urn:hud:hmis:owl#Gender',
```
andycwilliams, line 7: `${prefix}` can probably be used with all the above

andycwilliams (approved):

timbot1789:
> > I don't think there's any rush. We can pass this kind of thing over to the UI team as well and see what their thoughts are. Ultimately we as a project team need to figure out what data points we want to capture, and what use we have for them, then send that over to both UI and dev.
> > My goal in the meantime is to get the Civic profile to a point where it has the information you need to fill out an I9 form or other employment forms. An address is a necessary part of that.
> 
> I assume you meant UX that first time rather than UI? They have asked what the difference(s) will be between the civic profile and the user profile in terms of design. If they're intended to be mostly the same then that would save them some effort.

This is a good question. Let's follow up in discord about it.

PR #458, opened 2023-10-15
Improve accessible name for images on home page and prop names for HomeSection

milofultz:
## This PR:

Improves accessible name for images on home page.
 
**1.**  Adds context for all logo images
**2.** Removes unnecessary `aria-label`s
**3.** Mark decorative (non-informative images) as such using an empty `alt` attribute

### Components

```
src/components/Footer/RenderCompanyInfoSection.jsx
src/components/NavBar/NavbarDesktop.jsx
src/components/NavBar/NavbarLoggedOut.jsx
src/components/NavBar/NavbarMobile.jsx
src/pages/Home.jsx
```

### Tests

```
test/components/NavBar/NavbarDesktop.test.jsx
test/components/NavBar/NavbarLoggedOut.test.jsx
test/components/NavBar/NavbarMobile.test.jsx
test/pages/__snapshots__/Home.test.jsx.snap
```

## Screenshots (if applicable):

Should be no difference visually.

## Additional Context (optional):

Following guidance around [decorative](https://www.w3.org/WAI/tutorials/images/decorative/#example-1-image-used-as-part-of-page-design)/[informative](https://www.w3.org/WAI/tutorials/images/informative/) images from W3. Verified using [this extension](https://accessibilityinsights.io/) and DevTools.

PR #516, opened 2023-11-13
Issue 456/home page structure enhancements

### LINKED ISSUE #456: Bug: Different heading levels used for styling purposes instead of semantically on the home page
### Bug Description:

Heading levels are used in ways to affect their styling and not being used semantically. This leads to poor experiences for users of assistive technologies. See [MDN's notes on navigation using headings](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#navigation).

---

### Reproduction Steps:    

**1.** Go to the home page and turn on your screen reader (Voiceover, in this instance).
**2.** Go to the list of headings (using the rotor in Voiceover)
**3.** Note that they are not linear (e.g. 2, 5, 2, 5; instead of 2, 3, 2, 3)

What should be expected is a linear tree of heading levels. Each child should be a direct descendant of the parent (e.g. an `<h2>` element can only have succeeding `<h3>`s as direct children).

Screen Shot 2023-10-13 at 4 55 53 PM

Screen Shot 2023-10-13 at 4 55 55 PM

Screen Shot 2023-10-13 at 4 55 57 PM

Screen Shot 2023-10-13 at 4 56 00 PM

Screenshots are from [this extension](https://accessibilityinsights.io/).

Use CSS to style elements using something like a class name or the tag name itself (for example, `.subheader` or applied directly to `h2`) and modify this based on zoom level, width of browser, etc. 

This would likely mean we need to decouple our styling from the MUI `Typography` component. Unsure how this is done, but I see the `sx` prop. 

Thinking it would probably make sense to either make components that are styled correctly that we can then apply a heading level to, or something of that nature, to make a more modular and more easily maintained structure (if these styles are only used in small parts of the app, I don't want to go full design system and build out unnecessary infra).

xscottxbrownx:
## This PR:
Resolves #456 
 
**1.** Added PASS logo and text to use as `h1` - my understanding is we are switching to this logo. Need a higher quality version.   
**2.** Refactored headers of home page to follow semantic outline.  
**3.** Added `<main>` to layout - only thing marked as main semantically previously was this home logged out page, nothing else.
**4.** Added some conditional rendering to solve issue of adding empty elements to DOM.

## Screenshots:

https://github.com/codeforpdx/PASS/assets/71395931/c47cc438-056a-49c7-bc3d-d7f5f0071013

 - Need to figure out what to do for proper headers in footer - wasn't sure because it depends on what page is being rendered currently (I think.)
 - Also believe the `Home.jsx` page can be refactored. Seems like some props are being passed to child components, that could just be initialized in those components themselves. Could also possibly be arrays of objects and .map()'ed - but not sure that would cut down code or improve readability until attempted. I do believe it would make for easier maintainability/changes in future.

**1.**  Styling feedback
**2.** Code review


src/components/Home/HomeSection.jsx
```diff
@@ -19,6 +20,7 @@ import Button from '@mui/material/Button';
  * @param {string} Props.href - section button href
  * @param {string} Props.label - section button aria-label
  * @param {boolean} Props.isReallySmallScreen - screen size
+ * @param {boolean} Props.noMargin - controls marginBottom
```
andycwilliams: (This is actually in regards to line 12 but GitHub doesn't seem to allow comments on unchanged lines)

Small typo, should be `HomeSection`.

src/pages/Home.jsx
```diff
@@ -38,6 +40,20 @@ const Home = () => {
...
         <section id="home">
           <Stack alignItems="center" justifyContent="center">
+            <Box my={8}>
+              <Box
+                component="img"
+                src="/assets/PASSLogolightmode.png"
+                alt=""
+                sx={{ width: isReallySmallScreen ? '80%' : '300px', color: 'primary.main' }}
+              />
+              <Typography variant="h1" color="primary" fontWeight="600">
```
andycwilliams: Maybe a slightly smaller size title here and move to above the logo? It may take up too much of the screen, in my opinion. Plus having the title on top would make it a little more prominent.

Or otherwise reduce the logo size. Lots of options.

src/pages/Home.jsx
```diff
@@ -38,6 +40,20 @@ const Home = () => {
       >
         <section id="home">
           <Stack alignItems="center" justifyContent="center">
+            <Box my={8}>
+              <Box
+                component="img"
+                src="/assets/PASSLogolightmode.png"
```
andycwilliams: If we go with this logo, then presumably we'll also want to update the smaller one in the navbar as well as the favicon.

src/pages/Home.jsx
```diff
@@ -38,6 +40,20 @@ const Home = () => {
...
+                alt=""
+                sx={{ width: isReallySmallScreen ? '80%' : '300px', color: 'primary.main' }}
+              />
+              <Typography variant="h1" color="primary" fontWeight="600">
+                PASS
+              </Typography>
+              <Typography variant="h2" component="p" color="primary.main" fontWeight="500">
```
andycwilliams: I think this is better suited as `h3`. `h2` is too close to the `h1` of the PASS title, in my opinion.

src/pages/Home.jsx
```diff
@@ -38,6 +40,20 @@ const Home = () => {
       >
         <section id="home">
           <Stack alignItems="center" justifyContent="center">
+            <Box my={8}>
+              <Box
+                component="img"
+                src="/assets/PASSLogolightmode.png"
+                alt=""
```
andycwilliams: There are a few empty `alt`s like here and `componentImageAlt=""` below.

andycwilliams (commented):
Just some thoughts. I possibly overlooked a few things in this initial look through.

The tests are failing here, but I'm assuming they're waiting to be updated once we settle on what direction we want to go?

Also, I believe we discuss (a little) about slightly changing the desktop layout by having the images and associated text alternate side-by-side. That could free up a bit of space and make it flow a little more nicely but wouldn't really work on the mobile version, which should probably just remain as is. Assuming we even want this change.

src/pages/Home.jsx
milofultz: We need to keep it as an `h2` to maintain the heading hierarchy, so if we want to differentiate it _visually_, I think we need to use the `variant` prop. (I know I'm probably mansplaining a little here, just want to make sure we maintain that hierarchy)

src/pages/Home.jsx
milofultz: Yeah, this should have something like "PASS logo." as the value.

https://sc.edu/about/offices_and_divisions/digital-accessibility/guides_tutorials/best_practices/alternative_text/logo-alt-text/index.php

src/components/Home/HomeSection.jsx
```diff
@@ -29,9 +31,10 @@ const HomeSection = ({
   button,
   href,
   label,
-  isReallySmallScreen
+  isReallySmallScreen,
+  noMargin
```
milofultz: nit: I'd probably opt for a positive and for a boolean-esque naming (`hasMargin`) over `noMargin`. `noMargin={false}` or `{noMargin ? null : 8}` is slightly harder to grok than `hasMargin={true}` or `{hasMargin ? 8 : null}`

src/layouts/Layout.jsx
```diff
@@ -35,7 +35,7 @@ const Layout = ({ ariaLabel, children }) => {
     >
       <NavBar />
       {session.info.isLoggedIn && <Breadcrumbs />}
-      {children}
+      <main>{children}</main>
```
milofultz, line 38: 🙏

src/components/Home/HomeSection.jsx
xscottxbrownx: Switching... I think my initial approach was for purposes of less code, but this suggestion makes sense & matches what I've seen from component libraries.

src/components/Home/HomeSection.jsx
xscottxbrownx: Will have this updated on next fix. Good catch, I kind of ignored previous code.

src/pages/Home.jsx
xscottxbrownx: Correct, but I'm not sure this is the exact logo desired. My understanding is it is the correct design desired, but not sure on color/size/quality/etc. So, I didn't want to go too far.

src/pages/Home.jsx
xscottxbrownx: @milofultz 
Ok, I was following the info that was all discussed/linked previously here: https://github.com/codeforpdx/PASS/pull/458

But maybe it's different for the logo?

@andycwilliams 
What I linked above is the discussion on all the empty alts for the other images that were already on this page (the alts were previous PR, I didn't change them.)

src/pages/Home.jsx
xscottxbrownx: @milofultz Do we?  I had it as a `<p>` as you can see on line 53.

milofultz: Gah, disregard that above comment, got totally mixed up on the props and things.

Hmm, actually am now wondering if this should be part of the `h1`, as then the `h1` would have context in terms of what PASS is ("PASS: Personal Access System for Services"). Don't know how nesting Typography elements works, but could do something like this:

```tsx
<Typography variant="h1" color="primary" fontWeight="600">
  PASS<span class="sr-only">: </span>
  <Typography variant="h2" component="span" color="primary.main" fontWeight="500">
    Personal Access System for Services
  </Typography>
</Typography>
```

with `sr-only` being some variant of https://webaim.org/techniques/css/invisiblecontent/

src/pages/Home.jsx
milofultz: Yeah logos are special. TBH I still don't quite grok the why, but every space I check says you should add an alt attribute for the logo. I think you are generally correct that this is decorative, but seems like logos have a special case

src/pages/Home.jsx
xscottxbrownx: I will dig into this today sometime

xscottxbrownx:
Latest attempt:
https://github.com/codeforpdx/PASS/assets/71395931/1d89fa06-7f5e-46c6-a9ae-8fa78803f14c

Still have some code work to do for proper headings, then to tackle tests.

---

Let me know any feedback on this design.

src/pages/Home.jsx
andycwilliams: Yeah I don't think there's been a real, definitive discussion about what logo to go with

andycwilliams:
> Latest attempt: https://github.com/codeforpdx/PASS/assets/71395931/1d89fa06-7f5e-46c6-a9ae-8fa78803f14c
> 
> Still have some code work to do for proper headings, then to tackle tests.
> 
> Let me know any feedback on this design.

The video isn't rendering but I downloaded it.

Overall I think it's getting better.

---

Line 70 I think could be an `h4` on smaller viewports. On the desktop version it's fine.

Also, when I try out mobile I see this, where its not formatted as nicely as in the video.

Capture

I've tried reloading and resizing but it doesn't change. Not sure what's causing it.

---

Using our primary theme color instead of black would be nicer and more consistent in my opinion. I don't recall if we have a version of the PASS logo with that yet, though.

Also, we use a version of the logo with the PASS text within the image. So I'm not sure if we want to use that here as well. But it doesn't really seem necessary and would maybe limit our responsiveness a bit.

---

In case you weren't already aware, the `Learn More` button does not function correctly. Plus it overlaps with the other elements on certain viewports.

---

Honestly, at this point I'd just like to get UX's perspective. We may be winging it a bit too much.

xscottxbrownx:
> > Latest attempt: https://github.com/codeforpdx/PASS/assets/71395931/1d89fa06-7f5e-46c6-a9ae-8fa78803f14c
> > Still have some code work to do for proper headings, then to tackle tests.
> > Let me know any feedback on this design.
> 
> The video isn't rendering but I downloaded it.
> 
> Overall I think it's getting better.
> 
> Line 70 I think could be an `h4` on smaller viewports. On the desktop version it's fine.
> 
> Also, when I try out mobile I see this, where its not formatted as nicely as in the video.
> 
> Capture
> I've tried reloading and resizing but it doesn't change. Not sure what's causing it.
> 
> Using our primary theme color instead of black would be nicer and more consistent in my opinion. I don't recall if we have a version of the PASS logo with that yet, though.
> 
> Also, we use a version of the logo with the PASS text within the image. So I'm not sure if we want to use that here as well. But it doesn't really seem necessary and would maybe limit our responsiveness a bit.
> 
> In case you weren't already aware, the `Learn More` button does not function correctly. Plus it overlaps with the other elements on certain viewports.
> 
> Honestly, at this point I'd just like to get UX's perspective. We may be winging it a bit too much.

**video not rendering**
 yeah it seems when github changed that you can't just click in the textbox and add an image/video, but now have to click the paperclip button to do so - they also limited video size. Any larger files (haven't figured out exact size yet) seem to make you download file instead of just play it.
 
 ---
 
 Changed line 70 per your thoughts, looks better in my opinion.
 
 ---
 
 Changed to green, but we don't have logo in that color yet that I could find.
 
 ---
 
 I was told the logo that has the PASS text combined is not properly optimized and doesn't scale well, so I didn't attempt it.
 
 ---
 
**Learn More**
 Yeah, I haven't seemed to figure out anchor "scroll to" links in MUI yet. I'll dig into it - if you happen to know by chance, let me know.
 
 ---
 
 **UX perspective**
 yeah pretty much most of the discussion now is not technically in the scope of this PR to fix the headings.

src/pages/Home.jsx
xscottxbrownx: This link also has an idea for breadcrumbs under the red "Examples" section.

timbot1789 (commented):
Looks like this is failing a bunch of unit tests. If we still havent settled on a design for the front page, it may be worthwhile to loosen up these tests so they don't fail on areas we haven't finalized yet.

src/pages/Home.jsx
xscottxbrownx: > Gah, disregard that above comment, got totally mixed up on the props and things.
> 
> Hmm, actually am now wondering if this should be part of the `h1`, as then the `h1` would have context in terms of what PASS is ("PASS: Personal Access System for Services"). Don't know how nesting Typography elements works, but could do something like this:
> 
> ```tsx
> <Typography variant="h1" color="primary" fontWeight="600">
>   PASS<span class="sr-only">: </span>
>   <Typography variant="h2" component="span" color="primary.main" fontWeight="500">
>     Personal Access System for Services
>   </Typography>
> </Typography>
> ```
> 
> with `sr-only` being some variant of https://webaim.org/techniques/css/invisiblecontent/

@milofultz  if you feel like taking an attempt at a solution here, I've had zero luck in getting it to render the way it is now and be part of the `h1`. I think that is the correct idea, I just can't seem to solve.  You can just make a commit/push to this branch if you'd like.

I believe everything but this is currently sufficient in the branch (I still need to tackle the tests - which I'm digging into because I don't think they are the best tests for this component and may need to rewrite.)

xscottxbrownx:
> Looks like this is failing a bunch of unit tests. If we still havent settled on a design for the front page, it may be worthwhile to loosen up these tests so they don't fail on areas we haven't finalized yet.

Just tagged in milo to attempt to solve our idea for a better `h1` for this page, and I'm moving onto the tests. Honestly, looking into them I don't believe they are the best written and may need to write a new approach.  But I'm just starting this now.

src/pages/Home.jsx
milofultz: Here's a PR with the changes: https://github.com/codeforpdx/PASS/pull/531/files

Feel free to merge them in if you feel they make sense

timbot1789:
I think you can remove the HomeSection unit tests. It seems to me that this is a section in too heavy development for those to be all that useful (and there's not too much interesting behavior on this page anyway). 

The `Home.jsx` snapshot test will need to either be updated with `npm run test -- --update-snapshot`, or replaced with a similar render test that shows it's rendering the content.

xscottxbrownx:
> npm run test -- --update-snapshot

Screenshot 2023-11-19 at 12 11 43 PM

I don't know when this all happened, but seems like tests are failing all over on Development branch too. I can start working on those next.  I'd like to learn more about testing.

So, my important question is - when do we want tests? When is it valuable to have them?

Like in the case of HomeSection:
1) switch it to see if the component returns anything
2) switch it to see if the component returns what we are currently expecting - maybe image and title text
3) delete the test

leekahung:
> > npm run test -- --update-snapshot
> 
> Screenshot 2023-11-19 at 12 11 43 PM
> I don't know when this all happened, but seems like tests are failing all over on Development branch too. I can start working on those next. I'd like to learn more about testing.
> 
> So, my important question is - when do we want tests? When is it valuable to have them?
> 
> Like in the case of HomeSection:
> 
> 1. switch it to see if the component returns anything
> 2. switch it to see if the component returns what we are currently expecting - maybe image and title text
> 3. delete the test

We would like to check to see if an element within a component is rendered, that's one important test, especially if that element would be something users would interactive with like a button or a text field.

The next thing should be if there are any significant rendering changes between breakpoints (like if widths would change from 25% to 100%) think we should test their CSS properties as well. For these tests, think we could make it into a binary test, one where we use the base value for Desktop as a basis for those tests. So the tests would either `.toBe(<what we expect the value to be on Desktop>)` or `.not.toBe(<what we expect the value to be on Desktop>)`. It'l be more flexible as well which simply shows that "rendering changes has occur".

Of course, if we change the base case (i.e. say Desktop for now), we'll have to update the test, but if we change the non-base case (i.e. mobile), the test should still run fine.

For snapshots, which has been in discussion for a while, I think we should outright replace with either #1 of what you've mentioned for the `Home` page. For #2, I think we could handle those when testing for smaller components like in the `HomeSection` component itself.

leekahung:
> I think you can remove the HomeSection unit tests. It seems to me that this is a section in too heavy development for those to be all that useful (and there's not too much interesting behavior on this page anyway).
> 
> The `Home.jsx` snapshot test will need to either be updated with `npm run test -- --update-snapshot`, or replaced with a similar render test that shows it's rendering the content.

Just to clarify, I'm in support of changing the snapshot the `Home` tests, but not removing the `HomeSection` tests (switch doesn't use snapshots), since they're mainly for detecting rendering changes between breakpoints, which can be quite significant from Desktop to Mobile. Think there's a better way of doing that rather than removing them.

Replacing the unit test from snapshots for `Home.jsx` would be good, I agree that there's simply too much in development for Home page at this moment. If we do end up removing one of the tests, think we can have `Home.test.jsx` removed instead of `HomeSection.test.jsx` since  the unit tests for the contents within that page, like `<HomeSection>` and `<KeyFeatures>`, can be handled in the unit tests in those components.

timbot1789:
> > npm run test -- --update-snapshot
> 
> Screenshot 2023-11-19 at 12 11 43 PM
> I don't know when this all happened, but seems like tests are failing all over on Development branch too. I can start working on those next. I'd like to learn more about testing.
> 
> So, my important question is - when do we want tests? When is it valuable to have them?
> 
> Like in the case of HomeSection:
> 
> 1. switch it to see if the component returns anything
> 2. switch it to see if the component returns what we are currently expecting - maybe image and title text
> 3. delete the test

@xscottxbrownx  Hmm I'm not getting those failures when I run `npm run test` on dev. Are all your dependencies up to date? You may want to run `npm ci` just to be safe.

I also see I gave you the wrong snapshot update command. It's actually `npm run test -- --update`. My bad. That shouldn't cause the test failures though. The bad argument should just cause the system to crash.

xscottxbrownx:
> > > npm run test -- --update-snapshot
> > 
> > 
> > Screenshot 2023-11-19 at 12 11 43 PM
> > I don't know when this all happened, but seems like tests are failing all over on Development branch too. I can start working on those next. I'd like to learn more about testing.
> > So, my important question is - when do we want tests? When is it valuable to have them?
> > Like in the case of HomeSection:
> > 
> > 1. switch it to see if the component returns anything
> > 2. switch it to see if the component returns what we are currently expecting - maybe image and title text
> > 3. delete the test
> 
> @xscottxbrownx Hmm I'm not getting those failures when I run `npm run test` on dev. Are all your dependencies up to date? You may want to run `npm ci` just to be safe.
> 
> I also see I gave you the wrong snapshot update command. It's actually `npm run test -- --update`. My bad. That shouldn't cause the test failures though. The bad argument should just cause the system to crash.

yeah most of them are this:
TypeError: Object.hasOwn is not a function

searching makes it seem that it's related to an outdate node version (older than 16.9),  but that would be strange as didn't have this before with tests and my node.  So, trying to update node now to new version - just having issues "switching" to the new version.   Says version 20 is installed, but when running `node -v` it still lists 16.4

Also did `npm ci` with no change.

---

I also found the snapshot update option, but going to to cut to instead write an actual test without snapshot entirely.

leekahung (commented):
Going over new changes, I think it looks good to go and alright for approval.

Of course, we would consider what type of transition when we have users click on the "Learn More" button, but that could be resolved later.

Other than that, think we just need to resolve these unit tests.

timbot1789:
> So, my important question is - when do we want tests? When is it valuable to have them?

@xscottxbrownx how do we address this without writing a book? I'll start by recommending a short video on the topic: https://www.youtube.com/watch?v=EZ05e7EMOLM

Unfortunately there's no easy answer to this, and  on top of that you have a bit of an edge case with the home page. It doesn't really fit with the rest of the app.

IMO, the question I start with is "What do I want the thing to do? What's interesting about this thing?" The answer to that question tells me what I should test. So if I have a document uploader where I can name documents and set expiry dates, then I want the uploader to do 3 things:

1. Upload documents
2. name documents
3. set expiry dates on documents

Then I write 3 tests focused on those 3 things. If I'm interested in another thing, like it rejects documents larger than 1 gb. Then I add a 4th test for that. This process is often called "behavior-driven development."

For the homepage, it seems you're interested in one thing: it renders the expected home page. So you need at least 1 test.

What's not entirely clear to me is if there's anything else interesting on this page, or if you've really decided what you want this page to do yet. If there's something else about this page that you think interesting, like it switches layouts when it's on a mobile device, then you probably want a test for that as well. Generally, static content doesn't need significant testing, since it's not doing too many interesting things.

If there are sections where you're still undecided what they should do, you don't need tests for them, but you should make it obvious that they are placeholders.

timbot1789:
> > > > npm run test -- --update-snapshot
> > > 
> > > 
> > > Screenshot 2023-11-19 at 12 11 43 PM
> > > I don't know when this all happened, but seems like tests are failing all over on Development branch too. I can start working on those next. I'd like to learn more about testing.
> > > So, my important question is - when do we want tests? When is it valuable to have them?
> > > Like in the case of HomeSection:
> > > 
> > > 1. switch it to see if the component returns anything
> > > 2. switch it to see if the component returns what we are currently expecting - maybe image and title text
> > > 3. delete the test
> > 
> > 
> > @xscottxbrownx Hmm I'm not getting those failures when I run `npm run test` on dev. Are all your dependencies up to date? You may want to run `npm ci` just to be safe.
> > I also see I gave you the wrong snapshot update command. It's actually `npm run test -- --update`. My bad. That shouldn't cause the test failures though. The bad argument should just cause the system to crash.
> 
> yeah most of them are this: TypeError: Object.hasOwn is not a function
> 
> searching makes it seem that it's related to an outdate node version (older than 16.9), but that would be strange as didn't have this before with tests and my node. So, trying to update node now to new version - just having issues "switching" to the new version. Says version 20 is installed, but when running `node -v` it still lists 16.4
> 
> Also did `npm ci` with no change.
> 
> I also found the snapshot update option, but going to to cut to instead write an actual test without snapshot entirely.

Hmm I'm using node version `16.20`. We currently only have version 16 pegged in the `.nvmrc`. We may need to update that if version 16.4 doesn't work.

If you have nvm installed, you can upgrade to 16.20 using that. `nvm install 16.20 && nvm use 16.20`

xscottxbrownx:
@timbot1789 I installed nvm.  And updated to 16.20, now tests working fine (except the 3 for home page and it's components, that I can now work on correctly.)

milofultz:
> @timbot1789 I installed nvm. And updated to 16.20, now tests working fine (except the 3 for home page and it's components, that I can now work on correctly.)

I updated the home section test because it was fairly straightforward from the code. Didn't want to update the snapshot test because it was a bit more involved 👍

xscottxbrownx:
> > @timbot1789 I installed nvm. And updated to 16.20, now tests working fine (except the 3 for home page and it's components, that I can now work on correctly.)
> 
> I updated the home section test because it was fairly straightforward from the code. Didn't want to update the snapshot test because it was a bit more involved 👍

ok, thanks... I finished new Home.test.jsx last night - I'll double-check KeyFeatures.tst.jsx now.

xscottxbrownx:
This is waiting on PR #531 to be merged in first. Then this could be merged, if approved.

src/pages/Home.jsx
leekahung, line 1: The "Key Features" components are out of alignment. The "Home Sections" looks fine though.

Screen Shot 2023-11-22 at 0 40 52

After looking at both components, I'm guessing it's because `<HomeSection>` contains their elements within a `<Stack>` with text-align center, while `<KeyFeatures>` simply utilizes a React Fragment.

src/pages/Home.jsx
```diff
@@ -36,48 +68,76 @@ const Home = () => {
...
+                color="primary"
+              >
+                Personal Access System for Services
+              </Typography>
+              <Typography color="secondary.main" sx={{ fontSize: '24px' }}>
+                Learn More
+              </Typography>
+              <IconButton aria-label="Scroll down" color="secondary" href="#about-pass">
+                <ArrowCircleDownOutlinedIcon sx={{ fontSize: '32px' }} />
+              </IconButton>
```
leekahung: Love the idea of a button to guide users along the site! But thinking about it again given we're already this far, think it'll be nice to have smooth scrolling down as well if we can have that included within this PR. (Disregard my earlier comment on getting this implemented later.)

I remember how to do this in CSS with a few lines. We just need to set `scroll-behavior: smooth` for the main `<html>` tag. Though in order to do that we'll need to create a CSS file again and import it to `index.jsx`.

https://github.com/codeforpdx/PASS/assets/14917816/3a4313bf-47d1-4d46-af9b-9d321b8f5323

The other way, of course, is to use the `useRef` hook to grab the element as a reference and make it run `scrollIntoView({ behavior: 'smooth' })` in React. A bit more complicated, but no CSS files required.

https://github.com/codeforpdx/PASS/assets/14917816/67d0b5b3-8828-4aa0-9f42-353fb6e30b46

leekahung (changes requested):
Finished going through the code, think it's about ready to go along with PR #531.

Just a few minor issues that needs addressing, but they seem pretty easy to fix/do (see feedback).

src/pages/Home.jsx
xscottxbrownx: I'm not sure what branch you are showing there.  The down arrow icon button has smooth scrolled to the first `HomeSection` below since this commit on 11/16: 103db67cb0524799f6709d447b78ae237530f2de

If you are on branch `issue-456/home-page-structure-enhancements`, have you pulled down latest?

make sure you are NOT on `...enhancements--h1`

src/pages/Home.jsx
xscottxbrownx: Don't know what I did to change it to misalign. Fixed now. Thanks for catch.

src/pages/Home.jsx
leekahung: Yeah, it's on `issue-456/home-page-structure-enhancements` with the latest changes (see clip of your recent changes with `KeyFeatures`). Didn't see the transition happening though.

https://github.com/codeforpdx/PASS/assets/14917816/1b55ac4e-9020-46f9-9b45-e7133a7e8849

Thought it might be a bug then?

xscottxbrownx: Sorry, I could've sworn I implemented smooth scroll via a ref previously!!! 

I'm out of it. 

Pushed commit now.

leekahung: No problem, just checked the new commits now. They're up and running. 👍

leekahung (approved):
From my end, think this is ready for merging.

andycwilliams:
I'm getting this error. Not sure if it's really app breaking or not. If it doesn't interfere then this should be good. The home page looks much cleaner and more professional now.



And I'd like to have a primary color version of the logo but that can wait.

Also, just remembered that the "Request a demo" button goes directly to Hugh's email. Is that what we'll ultimately want?

leekahung:
> I'm getting this error. Not sure if it's really app breaking or not. If it doesn't interfere then this should be good. The home page looks much cleaner and more professional now.
> 
> 
> And I'd like to have a primary color version of the logo but that can wait.

Ah, good catch! Yeah, it looks like there's an extra ref set within one of the `<HomeSection>` components (Line 111 in Home.jsx). That should be removed to get rid of the error. Only the parent `<div>` needs it.

src/pages/Home.jsx
```diff
@@ -36,48 +71,77 @@ const Home = () => {
...
+              <HomeSection
+                isReallySmallScreen={isReallySmallScreen}
+                componentImageSrc="/assets/web-security-green.png"
+                componentImageAlt=""
+                title="Keep Your Documents Safe and Secure Using Decentralized Technology"
+                description="Our innovative solution empowers individuals to manage their critical documents and control access for trusted organizations. PASS simplifies service access, enabling seamless documents requests and secure data sharing for a smoother support process."
+                button="Request a Demo"
+                href="mailto:hugh@codeforpdx.org"
+                hasMargin
+                ref={aboutRef}
```
leekahung: Andy caught this. Looks like there's a duplicate `ref` on Line 111, causing error in the browser console (see clip).

https://github.com/codeforpdx/PASS/assets/14917816/37ec1076-72fa-4108-893b-cebed08734a3

Error should disappear once we get that removed (see end of clip above).

leekahung (commented):
See feedback for error in console

src/pages/Home.jsx
xscottxbrownx: so this is where I had ref setup previously! I don't know what all happened there, but removed now.

leekahung (approved):

test/components/Home/KeyFeatures.test.jsx
```diff
@@ -7,19 +7,19 @@ import createMatchMedia from '../../helpers/createMatchMedia';
 const MockKeyFeaturesDefault = () => <KeyFeatures description="Example Text" />;
 const MockKeyFeaturesMobile = () => <KeyFeatures isReallySmallScreen description="Example Text" />;
 
-it('renders 67% width default', () => {
+it('renders 85% width default', () => {
```
timbot1789: This is fine, but tests tied to specific CSS values like this probably won't be very resilient. I don't think we'd want this test to fail if someone decides that actually 84% is better for desktop than 85%. We may want to associate the test with more significant values/ranges.  Here it looks like mobile always renders `100%`, which seems like a pretty set value. So something like `expect(descriptionStyles.width).not.toBe('100%')` may be more reliable.

In general we may be able to make tests like 'It renders desktop display on desktop' which checks for a `desktopLayout` class, and then a 'It renders mobile display on mobile' which checks for a `mobileLayout` class. Then we can update the precise details of the class as we need without causing the tests to fail.

test/pages/Home.test.jsx
```diff
@@ -1,24 +1,30 @@
...
-import { ThemeProvider } from '@mui/material/styles';
+import { Home } from '@pages';
+import createMatchMedia from '../helpers/createMatchMedia';
 
-import Home from '../../src/pages/Home';
-import theme from '../../src/theme';
+describe('Home Page', () => {
+  afterEach(cleanup);
+  // if tests are failing check this value vs value in Home.jsx
+  const h1FontSize = '144px';
```
timbot1789: Same as above

timbot1789 (approved):

test/pages/Home.test.jsx
xscottxbrownx: This is the same thing.  This is the current base value - so .toBe(basevalue) & other tests expect .not.toBe(basevalue).

xscottxbrownx: I'd rather it be something like this, so I can check the ordered of the rendered elements:

```js
it('renders mobile', () => {
    window.matchMedia = createMatchMedia(599);
    render(<Home />);
    const h1 = screen.getByTestId('testStack');
    const { children } = h1;
    
    expect(children[0].toEqual(<Typography />));
  });
  ```
  
  then desktop would just be like `expect(children[0].not.toEqual(<Typography />));`
  
  but can't seem to figure out syntax or where in docs to find much besides the matchers.

Comment (xscottxbrownx) [reply test/pages/Home.test.jsx]: Ok this seemingly works for the intention:

```js
describe('Home Page', () => {
  afterEach(cleanup);

  it('renders with correct order of logoSection on mobile', () => {
    window.matchMedia = createMatchMedia(599);
    render(<Home />);
    const h1 = screen.getByTestId('testStack');
    const children = Array.from(h1.children);

    expect(children.length).toBe(3);
    expect(children[0].tagName.toLowerCase()).toBe('span');
    expect(children[1].tagName.toLowerCase()).toBe('img');
    expect(children[2].tagName.toLowerCase()).toBe('span');
  });

  it('rrenders with correct order of logoSection on desktop', () => {
    window.matchMedia = createMatchMedia(1200);
    render(<Home />);
    const h1 = screen.getByTestId('testStack');
    const children = Array.from(h1.children);

    expect(children.length).toBe(2);
    expect(children[0].tagName.toLowerCase()).toBe('span');
    expect(children[1].tagName.toLowerCase()).toBe('span');
  });
});
```

test/components/Home/KeyFeatures.test.jsx
xscottxbrownx: Fully agreed and something discussed in depth in our discord DM. 

Changed all to have a base value so .toBe(basevalue) & .not.toBe(basevalue)

This is SLIGHTLY more resilient, but not by much as when base value changes you have the same effect. 
I think all CSS value based tests are no good.  Maybe adjusting tests to be more resilient is a future issue for the code base?

xscottxbrownx:
@milofultz feel like taking a look at these tests to see if anything sticks out to you as well?

test/components/Home/KeyFeatures.test.jsx
milofultz: I think the bigger question is what exactly is this supposed to be testing? If we are testing whether or not our styles change at certain values, I think it would make more sense to create a function that tests whether this does what it's supposed to to set `isReallySmallScreen` in `src/components/Messages/MessageButtonGroup.jsx`:

```ts
  const isReallySmallScreen = useMediaQuery('(max-width: 480px)');
```

If we can test `useMediaQuery`, then we won't really need to use these tests, will we?

milofultz: I agree that they are very temperamental. Since they are directly coupled with hard coded values, it will either be annoying or ignored by future devs, which kind of invalidates the purpose of tests, which is an alarm for unexpected changes. All that being said, trying to ensure the DOM doesn't change based on what we perceive to be unrelated changes *is* useful, but that's a great case for snapshots, not behavior-driven tests.

If we can test *behaviors* for the user, like [black-box testing](https://en.wikipedia.org/wiki/Black-box_testing), then that will be more resilient. And I think will also let us ask whether or not a test is useful, but that is only my opinion.

test/pages/Home.test.jsx
milofultz: This to me still feels very fragile. We're testing the exact order of the DOM, when for the user, none of this matters. If the goal is to ensure that the DOM doesn't change, let's just rely on snapshots as that seems more accurate for the purpose. Otherwise, I think tests should be testing the behavior that the user will experience

milofultz:
Honestly, an overhaul of making our test suites more focused, resilient, and effective to their intended purposes should probably be a different issue/PR, as this can become *really* side quest-y

test/pages/Home.test.jsx
xscottxbrownx: > This to me still feels very fragile. We're testing the exact order of the DOM, when for the user, none of this matters. If the goal is to ensure that the DOM doesn't change, let's just rely on snapshots as that seems more accurate for the purpose. Otherwise, I think tests should be testing the behavior that the user will experience

But the goal for this was not to ensure the DOM doesn't change. Quite the contrary.  In this particular component that is being tested, the DOM should be different for mobile and desktop. This tests just that, without the reliance on a css style check.

This is your h1 section that you figured out the nested syntax of.

test/components/Home/KeyFeatures.test.jsx
xscottxbrownx: I'm not sure I follow 100%, but I also believe that testing that react is doing react is silly and unnecessary. The maintainers of react and MUI are testing them, we don't need to as well.  I don't think we need to test if `useMediaQuery` works either. Kind of the point of using a library. 

I feel like the behavior (testing onclicks and such) should be done.

I'm not sold on checking css/styling or if mobile component is rendering in mobile screen or if desktop component is rendering in desktop screen.  It will render whatever we tell it to render. 

I'm not sure, I'm still studying.

PR #520, opened 2023-11-15
Make footer headings the appropriate heading level

milofultz:
Was mentioned in #456, but was small enough I thought could make it a small PR.

Since the footer is separate from the main content, and a page should only have one `h1`, this means that a footer should have their headings start at `h2`.

### DIFF
--- a/src/components/Footer/RenderCallToActionSection.jsx
+++ b/src/components/Footer/RenderCallToActionSection.jsx
@@ -16,7 +16,7 @@ import Typography from '@mui/material/Typography';
  */
 const RenderCallToActionSection = ({ isReallySmallScreen }) => (
   <Stack width={isReallySmallScreen ? 1 : 3 / 5} alignItems="center" justifyContent="center">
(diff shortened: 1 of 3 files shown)


xscottxbrownx (commented):
This is what I would've done to it too, but see my discord DM comments - as I'm not sure it's quite right for pages besides the logged out home page.

leekahung (approved):
These changes looks good to me. Think it's good to go.

milofultz:
> This is what I would've done to it too, but see my discord DM comments - as I'm not sure it's quite right for pages besides the logged out home page.

Because the footer stays the same throughout all the pages (afaict) and there will only ever be a single `h1`, that means that an `h2` will always be semantically appropriate.

Sounds like [it doesn't matter for SEO](https://iloveseo.com/seo/john-mueller-explained-how-one-could-use-h2-tags-in-the-footer/) and each heading has relevant content beneath it, so seems legit to me.

PR #529, opened 2023-11-16
Enhancement 489: Migrate existing document table list to mui data grid

### LINKED ISSUE #489: [Enhancement] - Migrate Existing Document List Table to MUI Data Grid

As detailed in the title, we could migrate what we have now for our document list table into an MUI Data Grid. Given what we could see from the Data Grid implemented for Contact List, I believe it'll be quite useful for the Document List as well.

Not only does it have a filter function, which could help users sort out/find specific document types on the fly, it's also mobile-friendly. With regards to the number of columns Document List currently has, I think we can consider that in a separate PR.

Similar to Contact List, we could implement the `<DataGrid />` component from MUI and pass in the documents list into the property `rows` as a list of document objects, with each row rendering the object properties as columns.

bingeboy:
## This PR:
Adds support for the MUI DataGrid Component in the DocumentTable component and removes the MUI Table and other MUI components associated to the table element.

Untitled


src/components/Documents/DocumentTable.jsx
```diff
@@ -32,47 +32,142 @@ import { EmptyListNotification, LoadingAnimation } from '../Notification';
...
+      field: 'description',
+      minWidth: 120,
+      flex: 1,
+      headerAlign: 'center',
+      align: 'center'
+    },
+    {
+      field: 'upload date',
+      type: 'dateTime',
+      valueGetter: ({ value }) => value && new Date(value),
```
leekahung: Think we can adjust the date objects into strings. Just the date would be fine. The same for Line 80 as well.

leekahung (changes requested):
Good work! This is a nice start.

2 things, could we get the Data Grid to be responsive? And could we change the date objects into proper date strings?

src/components/Documents/DocumentTable.jsx
bingeboy: This type of column throws an error when the date is converted to a string. I can remove the type `type: 'dateTime'`. Here is a screen render the differences:
image

leekahung: Was thinking along the lines of `toLocaleDateString`

src/pages/Profile.jsx
```diff
@@ -17,7 +17,7 @@ import { useTheme } from '@mui/material/styles';
 import { DocumentListContext } from '@contexts';
 // Component Imports
 import { ConfirmationModal, UploadDocumentModal, SetAclPermissionsModal } from '@components/Modals';
-import { DocumentTable } from '@components/Documents';
+import DocumentTable from '@components/Documents';
```
bingeboy, line 20: We should be able to delete `DocumentTableRow.jsx` with this PR.

src/components/Documents/DocumentTable.jsx
```diff
@@ -32,47 +32,166 @@ import { EmptyListNotification, LoadingAnimation } from '../Notification';
...
+    <Container>
+      <DataGrid
+        columns={columnTitlesArray}
+        rows={documentListObject?.docList.map((document) => ({
+          id: document.name,
+          type: document.type,
+          name: document.name,
+          description: document.description,
+          delete: document,
+          'upload date': document?.uploadDate.toLocaleDateString(),
```
bingeboy, line 200: @leekahung let me know if this is the format you wanted.

src/pages/Profile.jsx
leekahung: Yes, I agree. With the new grid, it's no longer needed.

src/components/Documents/DocumentTable.jsx
leekahung: Yep, that would be good for `upload date`.

For `expiration date` since it could potentially be null from the user not including one, think we could have it leave an empty string if no expiration date was given. So something like:

```
'expiration date': document?.endDate?.toLocaleDateString() || '',
```

Or something that gives a similar result.

leekahung: Looks like just `'expiration date': document?.endDate?.toLocaleDateString()` works too

bingeboy: Ok, I'll add a test case for that.

src/components/Documents/DocumentTable.jsx
```diff
@@ -32,47 +32,166 @@ import { EmptyListNotification, LoadingAnimation } from '../Notification';
...
-          </Table>
-        </TableContainer>
-      </Container>
-    </ThemeProvider>
+    <Container>
+      <DataGrid
+        columns={columnTitlesArray}
+        rows={documentListObject?.docList.map((document) => ({
+          id: document.name,
+          type: document.type,
```
leekahung, line 196: For the document type, think we could find a way to return a formatted string using `src/constants/doc_types.js`.

So something along the lines of `type: DOC_TYPES[document.type]` with DOC_TYPES imported.

Of course, it would look like there has been adjustments to how `document.type` was pushed into Solid from previous PRs so not all the types are consistently camelCase:

Screen Shot 2023-11-19 at 15 28 46

But I think we could reset everything by clearing whatever document list you're currently testing with by deleting the documents from PASS and adding them back in with the add documents button to see the changes.

andycwilliams:
Could we add the same options for filtering and density that the Contacts table has?

Also (and this relates to making the grid more responsive), could we remove some of the empty space on the sides? That will help fill out the space and give the user information more readily. And again this is what the Contacts table already does.

bingeboy:
@andycwilliams I agree, making this UI more like contacts makes sense to me.

src/components/Documents/DocumentTable.jsx
bingeboy: Updated to use DOC_TYPES.

leekahung (approved):
Just checked out the branch on my computer, loving the changes! Let me know if you still working on anything else for this PR. I think you've addressed what I've touched on before, so I'm good with approving as is.

Good work!

bingeboy:
@leekahung I pushed minor change to the test file. Removed 2 commented out lines and added a few more checks. I'm going to go ahead and merge.

PR #530, opened 2023-11-17
Version Update from v0.10.0 to v0.11.0

leekahung:
## This PR:

Version update here will be from v0.10.0 to v0.11.0 for the Master branch. Will create a new v0.11.0 Release after we got this merged into Master.


src/components/CivicProfileForms/HousingInfo.jsx
```diff
@@ -1,6 +1,88 @@
...
+    e.preventDefault();
+    if (!isSuccess) {
+      return;
+    }
+    add(formData);
+  };
+
+  return (
+    <form onSubmit={handleSubmit}>
+      <div
```
xscottxbrownx, line 45: Can switch to `<Box>` or even better - `<Stack>` which already has flex and direction of column

src/components/Messages/MessageButtonGroup.jsx
```diff
@@ -0,0 +1,59 @@
...
+      sx={{
+        display: 'flex',
+        padding: '20px 30px 10px',
+        flexDirection: isReallySmallScreen ? 'column' : 'row',
+        alignItems: isReallySmallScreen ? 'center' : ''
+      }}
+    >
+      <Button
+        variant="contained"
+        onClick={() => setShowModal(!showModal)}
```
xscottxbrownx: works this way, but since we only want this button to open the modal - 
`onClick={() => setShowModal(true)}` ?   More predictable/understandable.

src/components/Modals/UploadDocumentModal.jsx
```diff
@@ -97,7 +102,11 @@ const UploadDocumentModal = ({ showModal, setShowModal }) => {
   return (
     <Dialog open={showModal} aria-labelledby="upload-document-dialog" onClose={clearInputFields}>
       <FormSection title="Upload Document">
-        <form onSubmit={handleDocUpload} autoComplete="off">
+        <form
+          onSubmit={handleDocUpload}
+          autoComplete="off"
+          style={{ width: isSmallScreen ? '100%' : '100%' }}
```
xscottxbrownx: No need for ternary

src/constants/doc_types.js
```diff
@@ -5,16 +5,16 @@
  * @memberof utils
  * @name docTypes
  * @type {Array<string>}
- * @property {string} BankStatement - Bank Statemet
- * @property {string} Passport - Passport
- * @property {string} DriversLicense - Drivers License
- * @property {string} Other - Anything Else
+ * @property {string} bankStatement - Bank Statemet
```
xscottxbrownx: typo

src/model-helpers/Document.js
```diff
@@ -34,29 +21,12 @@ import {
...
-  await saveSolidDatasetAt(`${containerUrl}signature.ttl`, signatureDataset, {
-    fetch: session.fetch
-  });
-};
-
 /**
  * Helper Function that returns Driver's License ttl file based off of image passed
  *
- * @function createDriversLicenseTtlFile
+ * @function addDriversLicenseInfo
```
xscottxbrownx: Do we need both 27 and 29 ?

src/utils/barcode/barcode-scan.js
```diff
@@ -66,7 +66,7 @@ const csvToJson = (csvData) => {
 
   // TODO: Use more accurate naming convention here (i.e. obj is too generic)
   const obj = {};
-  rows.forEach((row, i) => {
+  rows.forEach((_, i) => {
```
xscottxbrownx: what is the `_` ?

src/components/CivicProfileForms/HousingInfo.jsx
leekahung: Think we could have this addressed in the next version in a dedicated PR like what we're doing for PR #533 with one of the other forms.

src/components/Messages/MessageButtonGroup.jsx
leekahung: Think we can do that I'll start a PR on minor revisions and cleanup for v10.

src/components/Modals/UploadDocumentModal.jsx
leekahung: Think we can remove this entirely. Just tested it without having much effect on the styling desktop or mobile.

src/constants/doc_types.js
leekahung: fixed

src/model-helpers/Document.js
leekahung: Removing duplicate line.

src/utils/barcode/barcode-scan.js
leekahung: While `row` is the intended element in the forEach iteration for `rows`, it's not used in the forEach loop.

Think I could add `row` back in, but simply include the underscore prefix to indicate that it's unused (`_row`) for better clarity.

xscottxbrownx: ahhh. I get it now. don't mind either way, just haven't seen before.

src/components/CivicProfileForms/HousingInfo.jsx
```diff
@@ -1,6 +1,88 @@
-import { Typography } from '@mui/material';
-import React from 'react';
+// React Imports
+import React, { useState, useEffect } from 'react';
+// MUI Imports
+import { TextField, Button } from '@mui/material';
```
xscottxbrownx: bloated mui import, unfortunately should be `import X from '@mui/material/X';`

src/components/Documents/DocumentTable.jsx
```diff
@@ -1,22 +1,41 @@
 // React Imports
 import React, { useContext } from 'react';
+
```
xscottxbrownx, line 3: Don't really care either way, but don't need these spacer lines between each category of imports. They don't bother me though.

src/components/NavBar/OidcLoginComponent.jsx
```diff
@@ -15,53 +18,116 @@ import { ENV } from '../../constants';
...
+        disablePortal
+        selectOnFocus
+        autoHighlight
+        blurOnSelect
+        openOnFocus
+        onOpen={() => setDropdownIsOpen(true)}
+        onClose={() => setDropdownIsOpen(false)}
+        value={selectedOidcFromDropdown}
+        inputValue={oidcIssuer}
+        onChange={(_, newValue) => {
```
xscottxbrownx, line 75: doesn't need to change but curious to know what the underscores are, and if they are necessary?

and differences between `onChange` and `onInputChange`.

leekahung: ^ @rioredwards

src/components/CivicProfileForms/HousingInfo.jsx
leekahung: Addressing issue in PR #552

xscottxbrownx (approved):

src/components/NavBar/OidcLoginComponent.jsx
rioredwards: I know! MUI didn't make this super intuitive...

The value of the input can change when the user selects from the dropdown OR types:

`onChange` triggers when the user selects a value from the dropdown.

`onInputChange` triggers when the value of the input changes (i.e. the user selects a value from the dropdown OR types in the text box).

[Source](https://mui.com/material-ui/react-autocomplete/#controlled-states)

The underscore is a naming convention for variables that aren't actually used. In this case, the `onChange` callback receives two parameters: `(event, value)`, but here, I only care about `value`. Since I need to put something in that first position, I put an underscore to indicate it's not important. Syntax highlighters, ESlint, and other devs know about this convention and won't consider it an accidental unused variable.

[Source](https://stackoverflow.com/a/11406839/18654411)

xscottxbrownx: ok, is this `(_, value)` in lieu of `e.target.value`? Or are they different?

xscottxbrownx: ok, is this `(_, value)` in lieu of `e.target.value`? Or are they different?

rioredwards: TL;DR:  In the case of `onInputChange`, they are the same, but in the case of `onChange` they are different...

Explanation:
In `onInputChange`: `e.target` is the `<input>` element, which has the value we are looking for.
In `onChange`: `e.target` is the element clicked on in the dropdown. This is just an `<li>` element, which doesn't have a value property. It **does** contain the display text for the option, but that **may** or **may not** be the actual value of the option. The **actual** value is stored internally and exposed in the second parameter.

PR #531, opened 2023-11-17
Make `h1` hold full title while maintaining formatting

milofultz:
Part of [PR 516](https://github.com/codeforpdx/PASS/pull/516).

This puts the whole heading and subheading within the `h1`, maintaining formatting and visuals.

Screen Shot 2023-11-16 at 8 45 23 PM

Screen Shot 2023-11-16 at 8 45 19 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "@mui/material": "^5.12.2",
     "@mui/styled-engine-sc": "^5.12.0",
     "@mui/system": "^5.12.1",
+    "@mui/utils": "^5.14.18",
     "@mui/x-data-grid": "^6.16.2",
     "@mui/x-date-pickers": "^6.5.0",
     "@tanstack/react-query": "^4.32.0",
--- a/src/pages/Home.jsx
+++ b/src/pages/Home.jsx
@@ -56,7 +56,7 @@ const Home = () => {
     </Typography>
   ) : (
     <Typography component="h1" color="primary" data-testid="testHomeH1">
-      <Stack direction="column" justifyContent="center" alignItems="center" spacing={4}>
+      <Stack justifyContent="center" alignItems="center" spacing={4} mb={18}>
         <Stack
           component="span"
           direction="row"
@@ -70,7 +70,7 @@ const Home = () => {
             {punctuation}
           </Typography>
         </Stack>
-        <Typography variant="h3" component="span" fontWeight="600" mb={18}>
+        <Typography variant="h3" component="span" fontWeight="600">
           {subheading}
         </Typography>
       </Stack>
--- a/src/pages/Home.jsx
+++ b/src/pages/Home.jsx
@@ -12,6 +12,7 @@ import SupportIcon from '@mui/icons-material/Support';
 import Typography from '@mui/material/Typography';
 import useMediaQuery from '@mui/material/useMediaQuery';
 import { useTheme } from '@mui/material/styles';
+import { visuallyHidden } from '@mui/utils';
 // Components Import
 import { HomeSection, KeyFeatures } from '@components/Home';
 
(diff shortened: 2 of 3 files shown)


src/pages/Home.jsx
```diff
@@ -28,25 +29,49 @@ const Home = () => {
...
-      </Typography>
-      <Box component="img" src="/assets/PASSLogolightmode.png" alt="PASS logo" width="50%" />
-    </Stack>
+    <Typography component="h1" color="primary">
+      <Stack component="span" justifyContent="center" alignItems="center" spacing={2} mb={2}>
+        <Typography variant="h1" component="span" fontWeight="500" fontSize="72px">
+          {heading}
+          {punctuation}
+        </Typography>
+        <Box component="img" src="/assets/PASSLogolightmode.png" alt="" width="50%" />
```
milofultz, line 60: Oh yeah, got some clarity from my coworkers about alt text for logos. You (@xscottxbrownx ) are correct that in this case, it is decorative so it should be left empty. Sorry about all the confusion with that 🙇

package.json
```diff
@@ -35,6 +35,7 @@
     "@mui/material": "^5.12.2",
     "@mui/styled-engine-sc": "^5.12.0",
     "@mui/system": "^5.12.1",
+    "@mui/utils": "^5.14.18",
```
timbot1789, line 38: What's this additional dependency? I can't find any documentation on it.

The npm repo makes it look like it may be private: https://www.npmjs.com/package/@mui/utils

src/pages/Home.jsx
```diff
@@ -28,25 +29,49 @@ const Home = () => {
...
-      <Typography variant="h1" fontWeight="500" fontSize="72px" color="primary">
-        PASS
-      </Typography>
-      <Box component="img" src="/assets/PASSLogolightmode.png" alt="PASS logo" width="50%" />
-    </Stack>
+    <Typography component="h1" color="primary">
+      <Stack component="span" justifyContent="center" alignItems="center" spacing={2} mb={2}>
+        <Typography variant="h1" component="span" fontWeight="500" fontSize="72px">
+          {heading}
+          {punctuation}
```
timbot1789: I'm not quite sure I understand what this punctuation is doing. Could you describe it a bit?

package.json
milofultz: This is what MUI recommended for screen-reader-only text: https://mui.com/system/screen-readers/

Turns out the dep is already in there (check out package lock), so this isn't actually much extra for the bundle

src/pages/Home.jsx
milofultz: The punctuation is largely just a nice thing for screen reader users, so when they get to the `h1`, they don't hear "PASS Personal Access..." as all one sentence. The colon will usually provide a pause between "PASS" and "Personal...". Otherwise, assistive technologies may say "colon" literally, which is also descriptive.

src/pages/Home.jsx
```diff
@@ -28,25 +29,49 @@ const Home = () => {
...
+          alignItems="center"
+          spacing={4}
+        >
+          <Box component="img" src="/assets/PASSLogolightmode.png" alt="" width="150px" />
+          <Typography variant="h1" component="span" fontWeight="500" fontSize="144px">
+            {heading}
+            {punctuation}
+          </Typography>
+        </Stack>
+        <Typography variant="h3" component="span" fontWeight="600" mb={18}>
```
xscottxbrownx: Can we move this `mb={18}` from here (line 70) to the `<Stack>` on line 56 instead?  This will look as intended. 

I would approve with this change.

src/pages/Home.jsx
```diff
@@ -28,25 +29,49 @@ const Home = () => {
...
-        variant="h1"
-        fontWeight="500"
-        fontSize={isReallySmallScreen ? '72px' : '144px'}
-        color="primary"
-      >
-        PASS
-      </Typography>
-    </Stack>
+    <Typography component="h1" color="primary">
+      <Stack direction="column" justifyContent="center" alignItems="center" spacing={4}>
```
xscottxbrownx: can remove `direction='column'` from line 56. `<Stack>` defaults to column.

xscottxbrownx (changes requested):
With these 2 small changes, can approve and merge.

src/pages/Home.jsx
xscottxbrownx: pushed this fix

src/pages/Home.jsx
xscottxbrownx: pushed this fix

xscottxbrownx (approved):

src/pages/Home.jsx
timbot1789: Is there a way to achieve this effect without writing screen reader specific code? Maybe wrap the section in a `header` tag and wrap the subtitle in a `p` element? https://mikebifulco.com/posts/semantic-html-heading-subtitle#use-this-code-to-add-subtitles-to-headings

PASS is volunteer-driven, and it doesn't have the resources to test PRs in multiple display environments. It's already a challenge to fully test our app in both desktop and mobile setups, so we can't guarantee that screen reader code will ever be validated, or maintained.

The ideal would be code that is fully agnostic to the display environment, and functions comparably in all environments. That way, a visual validation in the most recent version of Chrome can make us reasonably confident of the app's behavior in JAWS and NVDA. It's also easier to write automated tests for that kind of code.

I'm 100% on board with making PASS more usable for screen readers. We just need to do it in a way that's maintainable for a project this size. It's kind of a fun design challenge :)

milofultz: I hear you on maintainability. This code may not be the best solution (likely), but as you said, more maintainable and accessible solutions may require some significant changes elsewhere, when dealing with issues in the future.  

As for the h1/p suggestion, the text to be rendered needs to be _within_ the h1 for it to be presented to the screen reader. If someone hears “PASS”, it doesn’t actually say what it is. 

Another way would be leaving “PASS” out of the h1 and only containing the expanded acronym in the h1. Might be a bit cleaner overall.

milofultz: All this being said, I could leave some comments explaining the purposes of these elements.

milofultz: I cleaned it up a little and added a comment there. Happy to try something else, too, but this should be fairly simple to maintain, I think.

andycwilliams:
The test failed but once that's resolve I think this is good to go.

xscottxbrownx:
The re-written passing tests are in the branch we are trying to merge this code fix into.

andycwilliams (approved):

src/pages/Home.jsx
xscottxbrownx: @timbot1789

when you have a chance to re-look at this...

this PR is stalling #516

milofultz:
> The re-written passing tests are in the branch we are trying to merge this code fix into.

Rebasing this branch onto yours to get those tests

milofultz:
Assuming tests pass, I assume we could merge this and do a fast follow later for the concerns @timbot1789 has to not impede your branch merging @xscottxbrownx . Will leave it up to one of you three, as you have more PASS domain knowledge of what's best in a situation like this than I do.

src/pages/Home.jsx
timbot1789: This should be ok then. I think it would be good to go over accessibility standards with the project team at some point.

@milofultz  do you think you would be able to add [a small section to the testing wiki](https://github.com/codeforpdx/PASS/wiki/Testing) about testing the app with screenreaders and other assistive technologies? It's been a few years since I've done accessibility testing and I don't remember what the best practices are.

timbot1789 (approved):

PR #533, opened 2023-11-20
Issue 442/basic information form

### LINKED ISSUE #442: HMIS Profile Page: Basic Information Form
## Feature Description:
Implement the first form (basic information) for the HMIS profile page. We need this form to collect four pieces of data, and store them with the corresponding data tags:

Field Name | Field Type | Profile Predicate value
- First Name | Text Input | urn:hud:hmis:owl#FirstName
- Last Name | Text Input | urn:hud:hmis:owl#LastName
- Date of Birth |Date Input | urn:hud:hmis:owl#DOB
- Gender | Dropdown Selection | urn:hud:hmis:owl#Gender

On the form itself:

- If the profile already contains the corresponding value, we will display the value as plain text. 
- If the profile does not contain the value, we will show an input field.
- If the user clicks on a value they've already added, they will be able to edit it.

### Notes

- Date of Birth should be stored as an ISO-8601 date
- Gender is an enumerated list of values. We will store a number from this list in the profile, but we will display a corresponding description to the client:
```
0: Female
1: Male
2: Transgender male to female
3: Transgender female to male
4: Doesn't identify as male, female or transgender
8: don't know
9: Decline to answer
```

andycwilliams:
## This PR:
Resolves #442 

Just a draft for now to make sure I'm on the right track. 
 
**1.** Adds (basic) styling to `BasicInfo.jsx`
**2.** Adds logic from #515

Capture2
Capture

**1.** Ensure this component is properly implementing the `useCivicProfile` hook
**2.** Ensure the labeling (such as all the `id`s and `name`s) is consistent and adheres to any standards (if necessary)

**1.** What styling is needed for each form as well as overall Civic Profile section


timbot1789 (commented):
This all looks like it's on the right track.

- For the placeholder text, I'd say _legal first name_ and _legal last name_. We want to emphasize that the data put here should be usable in a legal context.
- I think there's a typo in the test that's causing it to fail. You're trying to select a textbox role, but the first name is actually an input role.

xscottxbrownx:
I haven't looked at this deeply yet...

But, personally don't like the date format - not a big deal though. Still logical.

When does submit button become active? Seems like it's disabled because `isSuccess` is false, but can't switch `isSuccess` to true without the button it seems.
Screenshot 2023-11-26 at 8 42 02 PM


And to hit on what Tim was saying more about the test...

I was able to get it passing by matching the `label` of each `<TextField>`:
`const firstNameField = getByRole('textbox', { name: 'Legal first name' });` worked.

xscottxbrownx:
I'm not really sure why but two things:

1) the margin between the first input and second is like double the others on mobile
2) the clear function doesn't seem to clear the date field

leekahung:
Think the direction of this PR is pretty good. With regards to date formatting, think we could follow what HMIS utilize.

andycwilliams:
> But, personally don't like the date format - not a big deal though. Still logical.

Do you mean `YYYY-MM-DD`? Because that's what HMIS states to use.

> the margin between the first input and second is like double the others on mobile

Fixed.

> the clear function doesn't seem to clear the date field

Yeah I haven't messed with that too much because I'm pretty sure all input must be rendered as string (again, per HMIS). But dates are entered as objects, so I assume that that will just have to be stringified? I've procrastinated this part a bit and need a bit of advice on the next step.

> I was able to get it passing by matching the `label` of each `<TextField>`: `const firstNameField = getByRole('textbox', { name: 'Legal first name' });` worked.

Still haven't finished the test. I think I need to resolve the above issue first.

timbot1789:
@andycwilliams  any updates on this one?

andycwilliams:
> @andycwilliams any updates on this one?

This is next on my to-do list

andycwilliams:
The Clear button should be good now. Still troubleshooting `BasicInfo.test.jsx` not full passing. Something to do with line 38: `const dateOfBirthField = getByRole('textbox', { name: 'Date of birth' });`

The old screenshots seem to be broken so here are some new ones.

src/components/CivicProfileForms/BasicInfo.jsx
```diff
@@ -1,8 +1,169 @@
...
+              onChange={handleChange}
+              value={formData.legalGender}
+              labelId="hmis-basic-info-gender"
+              defaultValue=""
+            >
+              <MenuItem value={0}>Female</MenuItem>
+              <MenuItem value={1}>Male</MenuItem>
+              <MenuItem value={2}>Transgender male to female</MenuItem>
+              <MenuItem value={3}>Transgender female to male</MenuItem>
+              <MenuItem value={4}>Doesn&apos;t identify as male, female or transgender</MenuItem>
```
timbot1789:
```suggestion
              <MenuItem value={4}>Don&apos;t identify as male, female or transgender</MenuItem>
```

src/components/CivicProfileForms/BasicInfo.jsx
```diff
@@ -1,8 +1,169 @@
...
+                onChange={handleChange}
+                value={formData.legalDOB}
+                format="YYYY-MM-DD"
+                type="date"
+                disableFuture
+                openTo="year"
+                views={['year', 'month', 'day']}
+              />
+            </LocalizationProvider>
+            <FormHelperText>YYYY-MM-DD</FormHelperText>
```
timbot1789: Not an issue here, but we don't need to obligate this format. As long as we properly convert to this format during storage, we can allow the user to enter whatever format they want.

We can improve this later though.

src/components/CivicProfileForms/BasicInfo.jsx
```diff
@@ -1,8 +1,169 @@
...
+            fullWidth
+          />
+        </Grid>
+        <Grid item xs={12} sm={6}>
+          <FormControl fullWidth>
+            <LocalizationProvider dateAdapter={AdapterDayjs}>
+              <DatePicker
+                id="hmis-basic-info-date-of-birth"
+                name="legalDOB"
+                label="Date of birth"
```
timbot1789: It think the issue with the test is that when <DatePicker /> renders to HTML, it does not add this label to the input element. Instead, it adds a default aria label, `choose date`. The test is grabbing this aria label for the input rather than the label name.

timbot1789: This MUI issue has a fix for it:
https://github.com/mui/mui-x/issues/8146

Adding that fixes the datepicker failure, but reveals some issues in other parts of the test.

The gender select has a role of "combobox" not "textbox"

And the "submitButton" is failing because there are multiple buttons on the page.  Using "queryByRole" instead of "getByRole" will return all buttons on the page. That should help with getting the test to behave properly.

src/components/CivicProfileForms/BasicInfo.jsx
andycwilliams: I'll remove line 113. Would it be fine to leave the rest unchanged for now (such as `format="YYYY-MM-DD"`)?

src/components/CivicProfileForms/BasicInfo.jsx
andycwilliams: I believe I'm making progress with the tests, but am still running into some. I've tried several different approaches but hesitate making any changes and so wanted to show what the current version is in the likely case I'm missing something.

src/components/CivicProfileForms/BasicInfo.jsx
timbot1789: If we are going to require that the user submits data as `YYYY-MM-DD`, then we need to keep line 113. We need to communicate what format we require the dates to be submitted it. 

My point is that we are not obligated to receive user input in that format if we think another format is easier for users. As long as we store the date in the format `YYYY-MM-DD`, we can have the user enter the date in whatever format we want.

andycwilliams: Okay, I'll return the helper text for the time being.

andycwilliams:
Opening up this PR now. That final test still isn't passing, so my hope right now is to comment that one out and create a separate issue to address it. 

Also please ignore my above mis-click.

leekahung:
> Opening up this PR now. That final test still isn't passing, so my hope right now is to comment that one out and create a separate issue to address it.

Think we could make an exception for this one. We could comment on the unit test for the basic information form for now, but like what you've mentioned we'll have to open an issue/bug report to fix it.

Can approve this after the merge conflict resolution.

test/components/CivicProfileForms/BasicInfo.test.jsx
```diff
@@ -1,11 +1,84 @@
...
+    await userEvent.type(legalFirstName, basicInfoProfile.legalFirstName);
+    await userEvent.type(legalLastName, basicInfoProfile.legalLastName);
+    await userEvent.type(legalDOB, basicInfoProfile.legalDOB);
+    await userEvent.type(legalGender, `${basicInfoProfile.legalGender}{enter}`);
+
+    await userEvent.click(clearButton);
+
+    expect(legalFirstName.value).toBe('');
+    expect(legalLastName.value).toBe('');
+    expect(legalDOB.value).toBe(null);
```
leekahung: I believe this test is failing because the input field for the date picker is not your typical text field. You can't use the same approach to test the data picker. The same would be said for `legalGender` which is a select field.

If you'd like, we can open up another issue to write unit tests for these inputs.

src/components/CivicProfileForms/BasicInfo.jsx
```diff
@@ -1,8 +1,181 @@
...
+ * @name BasicInfo
+ * @returns {React.JSX.Element} The BasicInfo Component
+ */
+const BasicInfo = () => {
+  const { data, add, isSuccess, storedDataset, refetch } = useCivicProfile();
+  const [formData, setFormData] = useState({
+    firstName: '',
+    lastName: '',
+    dateOfBirth: null,
+    gender: ''
```
leekahung: Not too big of an issue since the feature appears to be working as intended, however, MUI is complaining about the input fields if it's not using one of the values from the option elements.

Screenshot 2024-04-23 at 7 45 27 PM
Screenshot 2024-04-23 at 7 45 41 PM

Think we could simply set the defaultValue, state, and clear value for gender as the number `9` to work around the problem. So Line 36, Line 75, and Line 133.

leekahung (approved):
Hey Andy, I've finished reviewing the PR. Just one minor feedback on potential changes (see feedback), but it seems to be working overall, so I'm approving.

Issue #259, opened 2023-06-16
Update signatures system to use verifiable credentials


timbot1789:
**Is your feature request related to a problem? Please describe.**
Currently, we are adding cryptographically secure signatures to documents in PASS that can accurately claim if a document has been seen and certified by a certain entity.

However, the method we're using to assign the signatures is a bespoke method designed by me. I am not a security expert, and my design is not particularly robust. Because of this, it will be difficult for other institutions to read the signature, and even if they can, they may be hesitant to accept a nonstandard security system designed by an unpaid volunteer without expertise in the area.

**Describe the solution you'd like**
There is a W3C standard for issuing these types of certificates, [verifiable credentials](https://www.w3.org/TR/vc-data-model/). This system is conceptually similar to what I wrote up, but is an accepted industry standard with several supporting libraries. Switching the signature system to instead issue these verifiable credentials will make PASS easier to use, and more trustworthy.

**Describe alternatives you've considered**
There are several other individual credentialing standards for individual documents (like mobile driver's license). They are all useful, and PASS would benefit from adding them. However, verifiable credentials has a few advantages that make it a good starting point:
1) It's web native. Other standards tie the credential to a specific device that must be brought with you, or otherwise can't be used on the web
2) It can be used on any file. Other standards can only be used on certain file types
3) It's fully decentralized. Anyone can issue a verifiable credential. And anyone can anonymously verify a web credential. Other standards require the credential to be issued by certain bodies, and only allow verification by certain bodies

**Additional context**
We'll want to research a good library to use to generate web credentials. It needs:
- To be usable in the browser with javascript
- preferably be able to render encryption keys as JSON Web Token.

timbot1789:
We will probably need to run these signatures against RDF documents at some point. If we do, we should sign the canonical form of the graph:

https://www.w3.org/TR/rdf-canon/

Currently, we're signing the TTL file itself, which is somewhat limiting. We'll want to find a Javascript library that generates the canonical form for us.

PR #535, opened 2023-11-21
[No Issue] Remove unused credentialing system

timbot1789:
Remove our old demo credentialing system from the codebase. It's not currently used, and we want to move to a system based on [verifiable credentials](https://github.com/codeforpdx/PASS/issues/259) instead. We generally shouldn't try to roll our own security.

The related code also breaks on newer versions of node, and the tests related to it run slowly. Since it seems to be more a hindrance to the project at this point than an asset, we should probably just remove it.


leekahung (approved):
Yeah, think that would be good.

We could have a discussion in the upcoming meeting on how we could handle verification of documents moving forward.

PR #537, opened 2023-11-22
Issue 500/sign in modal

### LINKED ISSUE #500: Bug: mobile "sign in" modal side scroll
### Bug Description:

The mobile sign-in button needs styling to match desktop 'login' button (the text stays as "sign in" to keep button distinct from the login button). 

On top of that, the modal has side to side scroll, so the input field within the modal needs styling adjustments for mobile (at least down to 280px). (see screenshot below)

Screenshot 2023-11-07 at 7 13 07 PM

Screenshot 2023-11-07 at 7 13 13 PM

Update MUI styling for the sign in button and adjust the styling for the element with the input field to fit within the modal. The font size within that field could be adjusted, if needed.

xscottxbrownx:
## This PR:
Resolves #500 
 
**1.**  Changes styling of "Sign In" button to match "Log In" button
**2.** shrinks size of input on the SignInModal, so no side scrolling occurs   

Screenshot 2023-11-21 at 8 33 46 PM

Screenshot 2023-11-21 at 8 33 52 PM

### DIFF
--- a/src/components/NavBar/NavbarLoggedOut.jsx
+++ b/src/components/NavBar/NavbarLoggedOut.jsx
@@ -42,7 +42,6 @@ const NavbarLoggedOut = () => {
                 variant="contained"
                 color="secondary"
                 size="large"
-                // sx={{ opacity: '0.8', color: 'black' }}
                 sx={{ borderRadius: '20px' }}
                 onClick={handleSignIn}
               >
--- a/src/components/NavBar/NavbarLoggedOut.jsx
+++ b/src/components/NavBar/NavbarLoggedOut.jsx
@@ -40,9 +40,10 @@ const NavbarLoggedOut = () => {
             <>
               <Button
                 variant="contained"
-                color="tertiary"
+                color="secondary"
                 size="large"
-                sx={{ opacity: '0.8', color: 'black' }}
+                // sx={{ opacity: '0.8', color: 'black' }}
+                sx={{ borderRadius: '20px' }}
                 onClick={handleSignIn}
               >
                 Sign In
--- a/src/components/Modals/SignInModal.jsx
+++ b/src/components/Modals/SignInModal.jsx
@@ -37,7 +37,7 @@ const SignInModal = ({ showSignInModal, setShowSignInModal }) => (
       </DialogContentText>
     </DialogContent>
 
-    <DialogActions sx={{ display: 'flex', justifyContent: 'center', alignItems: 'center' }}>
+    <DialogActions>
       <OidcLoginComponent setShowSignInModal={setShowSignInModal} />
     </DialogActions>
   </Dialog>
--- a/src/components/Modals/SignInModal.jsx
+++ b/src/components/Modals/SignInModal.jsx
@@ -28,12 +28,12 @@ const SignInModal = ({ showSignInModal, setShowSignInModal }) => (
     onClose={() => setShowSignInModal(false)}
   >
     <DialogTitle id="dialog-title" textAlign="center">
-      Sign In?
+      Sign In
     </DialogTitle>
 
     <DialogContent>
       <DialogContentText id="dialog-description" textAlign="center">
-        Welcome to PASS. Sign in below
+        Please select a Solid Server to log in
       </DialogContentText>
     </DialogContent>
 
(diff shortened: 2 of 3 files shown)


src/components/NavBar/NavbarLoggedOut.jsx
```diff
@@ -40,9 +40,10 @@ const NavbarLoggedOut = () => {
             <>
               <Button
                 variant="contained"
-                color="tertiary"
+                color="secondary"
                 size="large"
-                sx={{ opacity: '0.8', color: 'black' }}
+                // sx={{ opacity: '0.8', color: 'black' }}
```
leekahung: Think we can remove this `sx` prop if we're going to use the styles below with the border radius.

leekahung (approved):
Think I'm good with this, can approve. Just two things to note, both are non-blockers.

The first one was related to the commented out `sx`, see written feedback below.

The second was related to the `<Autocomplete>`, see clip. Was wondering on how we wish to manage the internal vertical scrolling of the modal if users were to select say the servers lower in the recommended list. Could address this in this PR or in another one, either is fine with me.

https://github.com/codeforpdx/PASS/assets/14917816/0ce5e8ca-fd35-485e-890e-e83afb2f85d0

As for a solution, the first thing that comes in mind was to have the `<Autocomplete>` vertical scroll instead, but open to alternatives.

andycwilliams (approved):
Approving the changes. Only suggestions would be to make the CANCEL and LOGIN buttons fill out the space a little more to the left and right. Maybe adding `fullWidth` will do that. Can't test it right now.

I was wondering if we want to center the titles, as well. For some modals they're left aligned whereas modals with forms are centered. But that's not really within this PR scope.

> Was wondering on how we wish to manage the internal vertical scrolling of the modal if users were to select say the servers lower in the recommended list. Could address this in this PR or in another one, either is fine with me.

That might require a bit of discussion and definitely some testing. So I think it can be left for the future. But something we for sure ought to come back to.

src/components/NavBar/NavbarLoggedOut.jsx
xscottxbrownx: ahhh, missed it. Thanks!

xscottxbrownx:
@leekahung @andycwilliams 

- Centered title of "Sign In?"
- Adjusted buttons

---

#### 3 questions for you both:

1) Is there any need to have the button and modal say "Sign In" vs the "Login" that the desktop sees?
     - If not, I can cut a little code down
    
2) What do you think of making the "Login" button green **only** when it's in the modal?
     - Just to match all other modals of "cancel" being red, and "continue" being green
     
3) Do we need the subtitle text "Welcome to PASS. Sign in below"
      - Most other modals like this do not have it, and I'm not sure it provides any value at all

leekahung:
> 1) Is there any need to have the button and modal say "Sign In" vs the "Login" that the desktop sees?
>      - If not, I can cut a little code down

Technically, it's not a login button in mobile,  so having it named different seems right to me. The login in the modal is the proper one. It was why I originally made the sign up button explicitly different with wording and coloring to point that out.

> 2) What do you think of making the "Login" button green **only** when it's in the modal?
>      - Just to match all other modals of "cancel" being red, and "continue" being green

Yeah, this is fine. If we're switching up the coloring for the Login Button think we can leave the sign up button outside red. 
  
> 3) Do we need the subtitle text "Welcome to PASS. Sign in below"
>       - Most other modals like this do not have it, and I'm not sure it provides any value at all

We could change it to, "Please select a Solid Server for login" or the like.

xscottxbrownx:
> The second was related to the `<Autocomplete>`, see clip. Was wondering on how we wish to manage the internal vertical scrolling of the modal if users were to select say the servers lower in the recommended list. Could address this in this PR or in another one, either is fine with me.
> 
> 
> As for a solution, the first thing that comes in mind was to have the `<Autocomplete>` vertical scroll instead, but open to alternatives.

Want to make separate issue?

leekahung:
> 
> > The second was related to the `<Autocomplete>`, see clip. Was wondering on how we wish to manage the internal vertical scrolling of the modal if users were to select say the servers lower in the recommended list. Could address this in this PR or in another one, either is fine with me.
> > 
> > 
> > As for a solution, the first thing that comes in mind was to have the `<Autocomplete>` vertical scroll instead, but open to alternatives.
> 
> Want to make separate issue?
> 

Sounds good to me. Think we could have it set up after the merge.

andycwilliams:
I think this is all good now. I double approve.

PR #541, opened 2023-11-25
Add max height for autocomplete and signInModal components

### LINKED ISSUE #539: [Enhancement] - Vertical scrolling for sign in autocomplete

Was originally described in https://github.com/codeforpdx/PASS/pull/537#pullrequestreview-1743775571 this is mainly for the mobile view. But it looks like the issue is somewhat resolved by allowing the modal to stretch instead (see clip).

https://github.com/codeforpdx/PASS/assets/14917816/9f655e48-715b-4d3f-8277-260102d224d9

Still think it'll be better if we could keep the modal from changing and allow the autocomplete vertical scroll, but that's just a nit. If there's consensus with this, think we could get this implemented.

### Rationale:

Quality of life for mobile version of PASS.

Enable vertical scrolling for the element in question. Could keep the viewable height of the autocomplete list to a fixed height.

rioredwards:
## This PR:
Resolves #539

https://github.com/codeforpdx/PASS/assets/55720928/713f6846-be49-4fb2-a6df-447355ce373f

The issue this PR resolves is actually asking to add vertical scrolling functionality to the Autocomplete. However, adding a max height to the component addresses the issue... The Autocomplete component had scrollable functionality all along, so by limiting its height, this allows the component's content to be scrollable.

### DIFF
--- a/src/components/Modals/SignInModal.jsx
+++ b/src/components/Modals/SignInModal.jsx
@@ -26,6 +26,7 @@ const SignInModal = ({ showSignInModal, setShowSignInModal }) => (
     aria-labelledby="dialog-title"
     aria-describedby="dialog-description"
     onClose={() => setShowSignInModal(false)}
+    PaperProps={{ sx: { maxHeight: '20rem', overflow: 'clip' } }}
   >
     <DialogTitle id="dialog-title" textAlign="center">
       Sign In
--- a/src/components/NavBar/OidcLoginComponent.jsx
+++ b/src/components/NavBar/OidcLoginComponent.jsx
@@ -79,6 +79,7 @@ const OidcLoginComponent = ({ setShowSignInModal }) => {
         onInputChange={(_, newInputValue) => {
           setOidcIssuer(newInputValue);
         }}
+        ListboxProps={{ style: { maxHeight: '12rem' } }}
         renderInput={(renderParams) => (
           <TextField
             {...renderParams}


leekahung:
As stated from Issue #539:

> Still think it'll be better if we could keep the modal from changing and allow the autocomplete vertical scroll

Is it possible to just have the <Autocomplete> drop down without the modal changing?

timbot1789 (approved):

leekahung (changes requested):
See earlier comment about modal.

rioredwards:
> Is it possible to just have the drop down without the modal changing?

I think this must be possible, but I couldn't figure out how to do it...

I tried just limiting the height of the Autocomplete component, but the parent component was still resizing as if the Autocomplete's height wasn't being limited... From what I could find, MUI doesn't provide an intuitive way to limit the height of this component, so this was the best I could come up with. 

I'm still getting the hang of MUI, but if anyone here wants to dig in and find a better solution, I would be happy to see it done right!

leekahung:
> I think this must be possible, but I couldn't figure out how to do it...
> 
> I tried just limiting the height of the Autocomplete component, but the parent component was still resizing as if the Autocomplete's height wasn't being limited... From what I could find, MUI doesn't provide an intuitive way to limit the height of this component, so this was the best I could come up with.
> 
> I'm still getting the hang of MUI, but if anyone here wants to dig in and find a better solution, I would be happy to see it done right!

No worries @rioredwards , all good! Had poked around a bit locally too without much luck. Think we could just approve this for now as a tentative solution.

leekahung (approved):

PR #542, opened 2023-11-27
Add feature text description and button

### LINKED ISSUE #441: Following Pod creation, send user to Civic profile page
### Feature Description:
After a client finishes creating a pod, we want to send them to the HMIS profile page to begin creating their profile. Currently it looks like this:
image

We would like to add 2 things to it:
1. A section explaining the data we want to collect, why we're collecting it, and how it's secure and private
2. A button that navigates them to `<PASS host>/hmis_profile`

For part 1, change the text of the page to the following:

> We've created a pod for you. You can get to it by going here: <current pass URL>
> If anyone asks you to share your pod with them, you can give them this web ID: <Their web ID>
> Next we'd like to help you build a profile about yourself that you can share with your case managers. This will make it faster and easier for your case managers to get you connected to services. All of the data you put in this profile will be _yours_, and the only people who will be able to access it are yourself, and the people you choose to share it with. We will ask you some sensitive questions, and you may not feel comfortable answering some of them. All data in these forms is optional. If you want don't want to answer certain questions, or you feel they don't apply to you, skip them. If you don't have time to complete the whole profile, feel free to leave and come back. You can resume right where you left off.
> Click this button to begin: `The same style next button we use in the HMIS profile forms`

For part 2, the button should be the same as the `next` button used in the HMIS profile form

kabszac:
## This PR:
Resolves #441  
 
**1.**    Added the new text description and added the next button

## Screenshot:
Screenshot from 2023-11-27 08-18-35


xscottxbrownx:
Haven't dug into this, but stylistically this needs work.

1) Nobody is going to read that block of text
2) I think the first 2 sentence URLs should stand out - bold, color, etc
3) The bottom link (which I don't think people will notice due to the huge block of text) is only a link and not a button. Should either change it to a button or call it a link.  Button would be much more noticeable.
4) Not sure if we need the "Click this button to begin". Could just be a button that says "Continue", "Next", or something to that effect.

All in all, this just needs to be much more skimmable. I don't think user's are going to read all this. Needs to be broken up & grouped appropriately. Needs to be more concise where possible (first 2 and last sentences for sure.)

leekahung:
To add what Scott mentioned with make this concise, think we could include a piece where you can mention that you could have others send you messages at `http://solidprovider/username/` from the `messages` page or contact list `send message` button

xscottxbrownx:
@kabszac still interested in this issue?

leekahung:
Hey @kabszac! Just wondering if you're still working on the suggested changes. If you are, let us know. Thanks!

andycwilliams:
It looks like they've been unassigned from the issue. If so, and they're no longer working on this PR, I could contribute to it.

andycwilliams:
Unassigning due to lack of response. @kabszac if you would like to come back to this PR please let us know as soon as you can.

PR #544, opened 2023-11-27
Cleaned up v0.10.0 as well as removing unused functions from utils

leekahung:
## This PR:

Related to version upgrade to v0.11.0 from v0.10.0 (PR #530). Mostly cleaning up v0.10.0 before merger as well as removing unused functions from utils which functionalities has migrated over to React Context/React Query.


xscottxbrownx (approved):
Fixes from my comments on v.0.10.0 -> v.0.11.0

PR #545, opened 2023-11-27
Enhancement 511: Skip Navigation Button

### LINKED ISSUE #511: Enhancement: "Skip Navigation" links should be available to skip repetitive content

n/a

Skip link for users of assistive technologies to skip past repetitive content.

### Rationale:

Keyboard users often have to move past the same things on every page (navigation links, etc.) to get to the main content. Skip links lets a keyboard user skip past those repetitive elements and get to the main content. 

See [WebAIM's article](https://webaim.org/techniques/skipnav/) for more details. You can also see an example of it on their page; on refresh, press tab and you'll see a skip link appear and disappear when you tab again.

Screen Shot 2023-11-12 at 10 20 48 AM

One that is not visible until you tab is common and well vetted. See [WebAIM's article](https://webaim.org/techniques/skipnav/) for more details.

bingeboy:
## This PR:
Resolves #511
 
[screen-recorder-tue-nov-28-2023-13-43-35.webm](https://github.com/codeforpdx/PASS/assets/96048/64e69858-44a3-4531-a6df-66740234a323)

Follow best practice from [webaim](https://webaim.org/techniques/skipnav/)

**1.**   Where is a good location for the button to reveal? 
**2.** Add higher contrast styles.


leekahung:
Hm...think we could have the button appear like where you had it on the top. Think we just need to make it more noticeable with a contrasting background.

Also think you could do a similar approach to what was merged in PR #516 with the `useRef` and `scrollIntoView` for scrolling. That way we could avoid hashes in our URL.

bingeboy:
I could be wrong but I think semantically we want a `<a/>` tag. I can add a `event.preventDefault();` to the handler and the scrolling effect will work.

For styling I can reverse the colors and add a white background with the NavBar background color for the text link. Open to any suggestions.

xscottxbrownx:
> I could be wrong but I think semantically we want a `<a/>` tag. I can add a `event.preventDefault();` to the handler and the scrolling effect will work.
> 
> For styling I can reverse the colors and add a white background with the NavBar background color for the text link. Open to any suggestions.

The MUI `Link` component ends up creating an `<a>` tag in the HTML. I'd have to check about the React Router `Link` component, I assume it does as well.

For MUI `Link`, check the Footer's `RenderCopyrightAndLinksSection.jsx` to see example of it compiling down to `<a>`.

For example of the `useRef`, check my icon button on home page to scroll down. I'm pretty confident we could do the same with a `Link`.

xscottxbrownx:
Another thought:

With the current state of this button, I'm not seeing the benefit to the user. 

I expected the tabbing order to also be affected, but after clicking this "skip to main", if you tab again it starts back up at the logo in navbar. I would expect it to be part of the main content instead.  Is this something we could control??  

AND

should the breadcrumbs be inside the `<main>` or no? Should it be the next tabbable item? Just an honest question.

bingeboy:
@xscottxbrownx good point. Checkout how [mozilla]( https://developer.mozilla.org/en-US/) added the "skip to main" functionality. I'll have to research a little on how to get the tab index to work like this. This seems like a good implementation. They also seem to skip directly into the content so I'd recommend bypassing the breadcrumb. 

I'll also update the `<a/>` with a `<Link />`.

leekahung:
> I'll also update the `<a/>` with a `<Link />`.

Oh, about `<Link>`, think we should use React Router's version of it and update any potential styling with `style`. That way we could get a better handle on routing. Or you could wrap React Router's `<Link>` with MUI's `<Link>` but with a different name just for that file to get the MUI styling included.

xscottxbrownx:
> > I'll also update the `<a/>` with a `<Link />`.
> 
> Oh, about `<Link>`, think we should use React Router's version of it and update any potential styling with `style`. That way we could get a better handle on routing. Or you could wrap React Router's `<Link>` with MUI's `<Link>` but with a different name just for that file to get the MUI styling included.

Yeah we usually do something like:

```js
import { Link as RouterLink} from ‘react-router-dom’;
import Link from ‘@mui/material/Link;

<Link component={RouterLink}>
```

so we get the MUI styling, but react router functionality of interrupting the network call.

xscottxbrownx:
> @xscottxbrownx good point. Checkout how [mozilla](https://developer.mozilla.org/en-US/) added the "skip to main" functionality. I'll have to research a little on how to get the tab index to work like this. This seems like a good implementation. They also seem to skip directly into the content so I'd recommend bypassing the breadcrumb.
> 
> I'll also update the `<a/>` with a `<Link />`.

Interesting that it does it there too. The tabbing.   Just makes no sense to me.

So you use keyboard only and want to tab straight down to the main content (or scroll to main content via the "skip to main content button"),  then why would next tab take you right back up to the top. That makes zero sense to me.

See the "Manual Checks" section here: https://fae.disability.illinois.edu/rulesets/BYPASS_1/#:~:text=The%20id%20attribute%20is%20preferred,it%20from%20the%20tab%20order.

According to that, my thoughts are the way it's intended to be implemented. But honestly I have no idea what that link is - just a quick google search of topic.

bingeboy:
@xscottxbrownx I think the Mozilla site mirrors the focusing. The homepage might be an exception. Pretty sure we are both on the same page. I haven't looked into skipping the breadcrumb. Based on my research some sites skip "sub nav" , some  don't, some have a dropdown with options. I'll just need a consistent node with an ID to focus.

Example skipping to the content with focus: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object

@leekahung do you think I can continue to use a hash with the location hook from react-router-dom? IMHO using a `<a/>` instead of a `<button />` seems more semantic and allows for history api. The only link that seems to not work is the "login" for the podserver which I haven't researched into why.

The link @milofultz referenced recommeneded:
```
<body>
<a href="#maincontent">Skip to main content</a>
...
<main id="maincontent">
<h1>Heading</h1>
<p>This is the first paragraph</p> 
```


This is working on my local:
```

      <Link
        component={RouterLink}
        to={{ pathname: location.pathname, hash: '#main-content' }}
        tabIndex={0}
        onClick={handleLinkClick}
        ... >
          Skip to main content
         </Link>

```

leekahung:
> @leekahung do you think I can continue to use a hash with the location hook from react-router-dom? IMHO using a `<a/>` instead of a `<button />` seems more semantic and allows for history api. The only link that seems to not work is the "login" for the podserver which I haven't researched into why.

Hm...looking at this more closely and reading over the WebAIM site for "Skip Navigation", we are not exactly navigating between pages, but rather across the same page.

The main purpose was to help mouse-less users and screen readers skip over elements that could be of little or no benefit to them going over the site, so I think a `button` would work just as well as an internal `anchor` tag (see clip, first with clicking, the second time with just using the enter key without the mouse).

https://github.com/codeforpdx/PASS/assets/14917816/09d915dc-b91b-415e-b3ad-6c9258001280

The benefit here of course is to allow the user to navigate within the same route without altering the existing URL. And since they're not really leaving the page, guess we don't need to worry about navigation history.

If we wish to make this more semantic, we could fallback on using `aria-label`.

https://github.com/codeforpdx/PASS/assets/14917816/15c4dc22-3787-4e7f-95b1-c2fce34f5389

One thing I've notice though is how the tabs behaves with "Skip Navigation":

https://github.com/codeforpdx/PASS/assets/14917816/4c82524d-b00a-4a1a-8de9-007d6dd597c2

Since the purpose for a "Skip Navigation" is to help ease access for potential mouse-less users and screen readers, we would probably want the tab right after the skip to be those in `main` and beyond instead of jumping back. The tab to `nav` should return last. So we might need to think of a way to make the tabIndex dynamic or jump ahead if "Skip Navigation" is triggered.

milofultz:
Hey all, 

Re: tab order, we need to move our _focus_ to the element in question. The easiest way to do that is to use the default in-browser method of supplying an ID and letting the browser navigate there. Using a standard `a` element without removing the default behavior is the way to go, as it handles _all_ of what we're trying to reverse engineer automatically. On the WebAIM site, tab to the skip navigation button, press "Enter", and it will work as we're discussing.

While this does change the URL, but if the goal is to ensure it gets added to the history, it looks like there is a solution: https://github.com/rafgraph/react-router-hash-link

Re: using `a` or `button`, `a` tags are explicitly for navigation and are the most semantically clear to the user that this goes somewhere. A button is usually some kind of functionality aside from navigation, so I'd probably opt for an `a` tag in this instance. 

> so we get the MUI styling, but react router functionality of interrupting the network call.

A nit: because this is targeting an element within the page, there is no network call. ~~React router~~ The network should not be involved because we're not having to load up another page through clicking this link.

milofultz:
Also, for navigating using the keyboard, you want to use <kbd>Enter</kbd>, not <kbd>Space</kbd> on links. See https://www.w3.org/WAI/ARIA/apg/patterns/link/examples/link/#kbd_label

src/components/NavBar/NavBarSkipLink.jsx
```diff
@@ -0,0 +1,63 @@
...
+  };
+
+  // handles blur on link
+  const handleBlur = () => {
+    setAnimateOut(true);
+  };
+
+  // handles clicking on link and scrolling to content
+  const handleClick = (event) => {
+    event.preventDefault();
```
milofultz: This should be removed once we implement a solution using React Router and hash links, e.g. https://github.com/rafrex/react-router-hash-link

src/components/NavBar/NavBarSkipLink.jsx
```diff
@@ -0,0 +1,63 @@
...
+    setAnimateOut(true);
+    const targetElement = document.getElementById('main-content');
+    if (targetElement) {
+      targetElement.scrollIntoView({ behavior: 'smooth' });
+    }
```
milofultz: Depending on the resulting behavior of a React Router solution, you probably won't need to have this either.

I'd also say that if we should delegate scroll behavior to the user or the site itself. If we have smooth scroll behavior set for the whole site (or not), we should use that setting.

src/components/NavBar/NavBarSkipLink.jsx
```diff
@@ -0,0 +1,63 @@
...
+  };
+
+  return (
+    <a
+      href="#main-content"
+      onClick={handleClick}
+      onFocus={handleFocus}
+      onBlur={handleBlur}
+      tabIndex={0}
```
milofultz: This is set by default, as it is an interactive element, and is unnecessary so can be removed. See interactive elements: https://html.spec.whatwg.org/multipage/dom.html#interactive-content-2

test/components/NavBar/NavBarSkipLink.test.jsx
```diff
@@ -0,0 +1,41 @@
...
+      const afterTabKeyPressTransform = window.getComputedStyle(skipLink).transform;
+
+      expect(afterTabKeyPressOpacity).toBe('1');
+      expect(afterTabKeyPressTransform).not.toBe('translateY(0)');
+    });
+
+    await userEvent.click(skipLink);
+
+    await waitFor(() => {
+      const skipLinkAfterClick = window.getComputedStyle(skipLink).transform;
```
milofultz: Should we test opacity here as well, since it's tested in both places above?

leekahung:
>Re: tab order, we need to move our focus to the element in question. The easiest way to do that is to use the default in-browser method of supplying an ID and letting the browser navigate there. Using a standard a element without removing the default behavior is the way to go, as it handles all of what we're trying to reverse engineer automatically. On the WebAIM site, tab to the skip navigation button, press "Enter", and it will work as we're discussing. While this does change the URL, but if the goal is to ensure it gets added to the history, it looks like there is a solution: https://github.com/rafgraph/react-router-hash-link

Considering we're simply creating a method to "skip navigation" to the main content (the main focus for this PR), I don't think we need to include that to the browser history.

> Re: using `a` or `button`, `a` tags are explicitly for navigation and are the most semantically clear to the user that this goes somewhere. A button is usually some kind of functionality aside from navigation, so I'd probably opt for an `a` tag in this instance.

This feels more like user interaction rather than navigation with the user interacting with the site to "scroll" within a page (without a mouse or they're using a screen reader) rather than "navigate" from one page to another, which was why I think `button` seems to be the more appropriate choice rather than `a`. We could with the `button` element, we can do what we can with the `a` tag,  with the "Enter" key jumping straight to the main content without involving the browser history, so I'm in favor of using a `button` tag instead, at least with this case. I would favor the `a` tag otherwise though if we are jumping between specific routes.

But I could be wrong about the definitions though. (Been trying to dig around with little luck) 😆 Would be good to discuss about it. If it ends up being anchor tags being the right choice, I think we could as @milofultz suggested, follow through with that.

milofultz:
> Considering we're simply creating a method to "skip navigation" to the main content (the main focus for this PR), I don't think we need to include that to the browser history.

That makes sense. I am curious the behavior of trying to use the back button after clicking that and if it would be any different than a non-SPA page. If only from my own experience, when a website has different behavior of navigating history, it can result in super frustrating experiences (losing form data, refreshing unnecessarily on low-bandwidth connections when expecting to go to a `#` link, etc.).

> This feels more like user interaction rather than navigation with the user interacting with the site to "scroll" within a page (without a mouse or they're using a screen reader) rather than "navigate" from one page to another, which was why I think `button` seems to be the more appropriate choice rather than `a`. We could with the `button` element, we can do what we can with the `a` tag, with the "Enter" key jumping straight to the main content without involving the browser history, so I'm in favor of using a `button` tag instead, at least with this case. I would favor the `a` tag otherwise though if we are jumping between specific routes.
> 
> But I could be wrong about the definitions though. (Been trying to dig around with little luck) 😆 Would be good to discuss about it. If it ends up being anchor tags being the right choice, I think we could as @milofultz suggested, follow through with that.

I'm going to respectfully disagree on using `button` over `a`. We have literal decades of precedent in intra- and inter-page navigation using `a` elements and functionality that does something (submitting forms, triggering Javascript functions, etc.) being triggered with `button` elements. That kind of affordance, if only cultural, is a _huge_ boon for good design and UX that we don't ever have to worry about ever again. All browsers and assistive technologies will change with these affordances and entrenched ideas in mind and fighting that current will be annoying at best and detrimental at worst (long term, think lawsuits).

If the only thing we do is look at every page talking about implementing skip links, that will show us that the vast majority, if not totality, of implementations use links. Note first that [they are called skip **links** by w3 themselves](https://www.w3.org/WAI/WCAG22/Understanding/bypass-blocks.html#examples), which is the committee that writes the guidelines this issue is looking to adhere to. Second, if we take the first few results on DDG of "skip links", all implementations are links:

* https://webaim.org/techniques/skipnav/#overview
* https://www.w3schools.com/accessibility/accessibility_skip_links.php (note: I hate to reference w3)
* http://web-accessibility.carnegiemuseums.org/code/skip-link/
* https://a11y-101.com/development/skip-link

And probably the most important thing: if users of assistive technology expect a skip link to be activated with "Enter" and then we have a button which by convention is activated with "Space", they will be likely very confused. If only that every skip link they have probably ever experienced was a link, but also that a button is expected to _do_ something, not navigate.

I appreciate you reading this, I just fear the unnecessary reinvention of the wheel with this and think we could follow the existing guidelines and make a slam dunk for our users here.

leekahung:
> And probably the most important thing: if users of assistive technology expect a skip link to be activated with "Enter" and then we have a button which only activates with "Space", they will be likely very confused. If only that every skip link they have probably ever experienced was a link, but also that a button is expected to _do_ something, not navigate.
> 
> I appreciate you reading this, I just fear the unnecessary reinvention of the wheel with this and think we could follow the existing guidelines and make a slam dunk for our users here.

Thanks Milo for the clarification! Yeah, re-inventing the wheel is probably something we don't really need in our case. I guess navigation was the term I was most confused about. I've always expected that navigation only refers to inter-page interactions. I've never thought intra-page interaction like with the internal anchor tags as navigation (definitely clears up my confusion over internal anchor tags to begin with if you could just use a button to scroll to a destination in the same page). Again, I didn't thought "navigating" would be the same as "scrolling" until now.

Edit: Oh! Got it working with space too. I see the support for buttons and space key for button (https://www.w3.org/WAI/ARIA/apg/patterns/button/examples/button/) and the enter support for link (https://www.w3.org/WAI/ARIA/apg/patterns/link/)

I'm guessing the discussions with keeping the "Enter" key with buttons when they want to keep buttons and links interactions separate was a can of worms topic for the w3.

milofultz:
> I'm guessing the discussions with keeping the "Enter" key with buttons when they want to keep buttons and links interactions separate was a can of worms topic for the w3.

Yeah, I'm very curious about the history! I'm sure there's some "let's put this on temporarily" stuff along with some IE "do it our way" stuff, too haha.

src/layouts/Layout.jsx
```diff
@@ -36,7 +36,9 @@ const Layout = ({ ariaLabel, children }) => {
       <NavBarSkipLink />
       <NavBar />
       {session.info.isLoggedIn && <Breadcrumbs />}
-      <main id="main-content">{children}</main>
+      <div id="main-content" tabIndex={-1}>
```
bingeboy: Oops reverting this back to `main`. I had been playing around making this a component on my local.

test/components/NavBar/NavBarSkipLink.test.jsx
```diff
@@ -1,18 +1,27 @@
...
+        <NavBarSkipLink />
+      </BrowserRouter>
+    );
     const skipLink = screen.getByText('Skip to main content');
     expect(skipLink).not.toBeNull();
   });
 
-  it('moves the button in and out of view', async () => {
-    render(<NavBarSkipLink />);
+  it.skip('moves the button in and out of view', async () => {
```
bingeboy: Not sure why this broke... I'm assuming from adding MUI or React Router but I'll update in my next commit

src/layouts/Layout.jsx
```diff
@@ -33,9 +33,12 @@ const Layout = ({ ariaLabel, children }) => {
         minHeight: '100vh'
       }}
     >
+      <NavBarSkipLink />
```
milofultz, line 37: Love how easy this is to implement 👨‍🍳 💋

src/layouts/Layout.jsx
```diff
@@ -33,9 +33,12 @@ const Layout = ({ ariaLabel, children }) => {
...
       <NavBar />
       {session.info.isLoggedIn && <Breadcrumbs />}
-      <main>{children}</main>
+      <main id="main-content" tabIndex={-1}>
```
milofultz: To ensure we don't have to add `main-content` id to every `main`, might be a good new issue to make a `Main` component that this stuff is hardcoded into.

milofultz: Also, it doesn't need a tabIndex of -1. That's for if we want to allow _focusing_, but we are just navigating to that area of the page.

test/components/NavBar/NavBarSkipLink.test.jsx
```diff
@@ -0,0 +1,55 @@
...
+      expect(afterTabKeyPressTransform).toBe('translateY(0%)');
+    });
+
+    await userEvent.click(skipLink);
+
+    await waitFor(() => {
+      const afterTabKeyPressOpacity = window.getComputedStyle(skipLink).opacity;
+      const skipLinkAfterClick = window.getComputedStyle(skipLink).transform;
+      // wait for animation to finish
+      setTimeout(() => {
```
bingeboy: While this is passing I'm open to a better idea on how to wait for the animation to finish.

src/layouts/Layout.jsx
xscottxbrownx: I'm missing the point of this logic here.  Why would there be multiple `<main>`'s? 

It's a wrapper for the entire content between the header and footer basically.

bingeboy: 😃  I had done this then reverted it...
Added `<Main />` with no `tabIndex`.

xscottxbrownx: > 😃 I had done this then reverted it... Added `<Main />` with no `tabIndex`.

Yeah my question is why the suggestion and implementation to make it it's own component, instead of just using native html `<main id='main-content'>`

Just trying to understand/learn what the thought process / thing I'm missing here is.

bingeboy: I think I posted that comment at the same time you posted yours. 🤣  
Now that the `<NavBarSkipLink />` is relying on the `main-content` id we have broken that chunk of code out into a smaller testable file. I think you could make a case for not making it a component too. I don't feel strongly one way or another. It's abstracted and encapsulated and will be nice to have going forward if more functionality is added.

test/components/NavBar/NavBarSkipLink.test.jsx
leekahung: It think we could just check to see if the setTimeout which hosts the animation gets triggered? I'm not too sure with regards to animation testing.

xscottxbrownx: Or maybe if any of our testing libraries have an ability to check if an element is in viewport, instead of just checking if element exists on DOM?   setup so element is not in viewport, then after click check that it is...  not sure.

src/layouts/Layout.jsx
milofultz: I'm just thinking that wrapping a `main` that essentially requires a specific `id` to work properly that we're going to have in various places around our app into an abstracted component would be good. Then future devs wouldn't have to worry about putting in the `id` and there would be context around _why_ it is done this way, which is good for future education and avoiding regressions.

test/components/NavBar/NavBarSkipLink.test.jsx
milofultz: Might be able to use `fireEvent.animationEnd` 
* https://klaviyo.tech/hitting-a-moving-target-testing-javascript-animations-in-react-with-jest-8284a530a35a#4a82
* https://stackoverflow.com/a/66465145

```tsx
await fireEvent.animationEnd(skipLink);

// do tests here
```

bingeboy:
Improved test and resolved merge conflict.

bingeboy:
I believe I have addressed all the provided feedback.

milofultz:
I'm still getting an unexpected tab order when using the keyboard. 

1. Tab to the skip link
2. Activate it with enter
3. It scrolls to correct place
4. Press tab
5. I _expect_ it to start within the `main` element. What actually happens is it starts at the top of the page again.

milofultz:
I'm still getting an unexpected tab order when using the keyboard. 

1. Tab to the skip link
2. Activate it with enter
3. It scrolls to correct place
4. Press tab
5. I _expect_ it to start within the `main` element. What actually happens is it starts at the top of the page again.

milofultz (approved):
Looks good and is working for me 👍

bingeboy:
Squashing

PR #546, opened 2023-11-27
536 enhancement confirmation modal for uploading docs

### LINKED ISSUE #536: Enhancement: Dialog modal when adding document of same name
Currently, there is a window.confirm() when a user tries to upload a document that has same name of an already uploaded document.

**File:** `UploadDocumentModal.jsx` in the `handleDocUpload` function's `catch` block.

---

Should be an MUI modal to match all other user feedback styling of app.

---

### Rationale:
It provides consistent and trustworthy feedback to the user.

---

Just something similar to other dialog modals that popup with 2 button options.

---

### Additional context:

https://github.com/codeforpdx/PASS/assets/71395931/c0da56da-fa63-444b-93b5-7737e1d637d7

rioredwards:
## This PR:
Resolves #536
 
**1.**    Extends/Refactors ConfirmationModal component to accommodate more parameters for improved re-use across the app. Adds default parameters for props.
**2.** Refactors UploadDocumentModal to use the updated ConfirmationModal component. This replaces the use of window.confirm() in the component, as requested in #536.
**3.** Updates other components throughout the app that use ConfirmationModal to accommodate new params.
**4.** Updates/Extends tests for these updated components ([ConfirmationModal.test.jsx](https://github.com/codeforpdx/PASS/blob/536-Enhancement-Confirmation_Modal_For_Uploading_Docs/test/components/Modals/ConfirmationModal.test.jsx) and [UploadDocumentModal.test.jsx](https://github.com/codeforpdx/PASS/blob/536-Enhancement-Confirmation_Modal_For_Uploading_Docs/test/components/Modals/UploadDocumentModal.test.jsx)).

### Demo

https://github.com/codeforpdx/PASS/assets/55720928/871093ff-a523-4025-9541-4a9e4c0f5172

### Upload Document Flow

Screenshot 2023-11-27 at 2 53 57 PM

- The original issue that this PR addresses (#536) didn't ask for a refactor of the ConfirmationModal component, but as I started working on it, the changes seemed necessary...
- The flow involves displaying the ConfirmationModal when an error is caught while attempting to add a new document. Currently, the only error that opens the confirmationModal is triggered when an existing file in the pod has the same name as the one being uploaded ("File already exists"), however, this can be extended to handle other errors as well if needed.

- More tests for the UploadDocumentModal component would be great. There are some basic tests, but they don't cover any user behaviors involving uploading documents... I'm not ready to tackle this at the moment, but could make an issue if needed. The diagram I included above could be useful for this.


leekahung:
Love the chart! Definitely breaks down the flow of what we're aiming to do with the UploadDocumentModal. Will have a look over the code shortly.

src/components/Modals/UploadDocumentModal.jsx
```diff
@@ -64,46 +81,78 @@ const UploadDocumentModal = ({ showModal, setShowModal }) => {
...
+  const onFormSubmit = async (e) => {
+    e.preventDefault();
+    setProcessing(true);
+    handleDocUpload('add');
+  };
+
   return (
     <Dialog open={showModal} aria-labelledby="upload-document-dialog" onClose={clearInputFields}>
+      <ConfirmationModal
+        {...confirmationModalContentVariant[confirmationModalType]}
```
leekahung: While I do like that we're able to make the confirmation modal dynamic for the  `<UploadDocumentsModal>`, was wondering if we should just make it into its own specialized component.

src/components/Modals/UploadDocumentModal.jsx
```diff
@@ -64,46 +81,78 @@ const UploadDocumentModal = ({ showModal, setShowModal }) => {
...
-    } catch (error) {
-      const confirmationMessage =
-        'A file of this name and type already exists on the pod. Would you like to replace it?';
-
-      switch (error.message) {
-        case 'File already exists':
-          if (window.confirm(confirmationMessage)) {
-            await replaceDocument(fileDesc, file);
-            addNotification('success', `File updated on Solid.`);
+    switch (uploadType) {
```
leekahung: I like this. 😄  It's a very nice refactor/optimization. Makes the case for adding or replace documents quite cleanly in the handler function.

src/components/Modals/UploadDocumentModal.jsx
rioredwards: I agree that would probably make things more clear. I'll give it a try!

rioredwards: Okay, I made a specialized component, like you suggested, and I refactored some of the code for clarity...

leekahung (approved):
Nice work! I'm okay with this being merged.

src/components/Modals/UploadDocumentModal.jsx
```diff
@@ -64,45 +67,78 @@ const UploadDocumentModal = ({ showModal, setShowModal }) => {
...
+        onCancel={handleUploadCancelled}
+        onConfirm={confirmationModalType === 'add' ? handleDocAdd : handleDocReplace}
+        uploadType={confirmationModalType}
+      />
       <FormSection title="Upload Document">
-        <form onSubmit={handleDocUpload} autoComplete="off">
+        <form
+          onSubmit={onFormSubmit}
+          autoComplete="off"
+          style={{ width: isSmallScreen ? '100%' : '100%' }}
```
xscottxbrownx: no need for ternary in this case

test/components/Modals/ConfirmationModal.test.jsx
```diff
@@ -21,7 +24,7 @@ describe('Default screen', () => {
 
   it('renders logout button container flex-direction as row default', () => {
     const { getByRole } = render(<MockConfirmationModalLogout />);
-    const cancelButton = getByRole('button', { name: 'Cancel' });
+    const cancelButton = getByRole('button', { name: 'Log Out' });
```
xscottxbrownx: this would be `const confirmButton` wouldn't it?

test/components/Modals/ConfirmationModal.test.jsx
```diff
@@ -42,10 +45,74 @@ describe('Mobile screen', () => {
 
   it('renders logout button container flex-direction as column mobile', () => {
     const { getByRole } = render(<MockConfirmationModalLogout />);
-    const cancelButton = getByRole('button', { name: 'Cancel' });
+    const cancelButton = getByRole('button', { name: 'Log Out' });
```
xscottxbrownx: same here... `confirmButton`?

test/components/Modals/UploadDocumentModal.test.jsx
```diff
@@ -24,3 +25,21 @@ it('renders cancel/upload group as column mobile', () => {
 
   expect(cssProperty.flexDirection).toBe('column');
 });
+
+it('renders initial state and elements correctly', () => {
+  const { getByText, getByRole, getByLabelText } = render(<MockUploadDocumentModal />);
+
+  // Check for initial state values
+  const modalTitle = getByText('Upload Document');
```
xscottxbrownx: would this not be `const modalTitle = getByRole('heading', {name: whatever const you initialize earlier for it});` ?

xscottxbrownx (changes requested):

src/components/Modals/UploadDocumentModal.jsx
leekahung: Oh, good catch!

Yeah, if both cases are just width: 100%, do't think we need to make this ternary. Just `width: 100%` should do.

test/components/Modals/UploadDocumentModal.test.jsx
xscottxbrownx: Or am I confusing you with somebody else that seemed to be against `getByText`?  This works, but I feel somebody brought up a point about it.

rioredwards: Good catch! I had the same thought. The title in the Modal isn't being rendered as a header tag at the moment, I'll look into changing that though because that would be better sematic HTML.

src/components/Modals/UploadDocumentModal.jsx
rioredwards: Fixed!

test/components/Modals/ConfirmationModal.test.jsx
rioredwards: Fixed!

test/components/Modals/ConfirmationModal.test.jsx
rioredwards: Fixed!

xscottxbrownx: it honestly doesn't matter, but this one is still off.  line 48 should be confirmButton and line 49 should be buttonContainer, similar to the test above it.

xscottxbrownx (approved):
Approving because code works and issues have been changed, but small issue in one test. No cause for it to fail or anything. Just a consistency thing for easier understanding.

test/components/Modals/ConfirmationModal.test.jsx
rioredwards: Woops! I'm gonna blame this on lack of coffee 😂

PR #547, opened 2023-11-28
463 enhancement your signup link user experience optimization

### LINKED ISSUE #463: Enhancement:  "Remove Profile Image" user experience optimization
The "Remove Image" button on the Profile page uses default window prompt instead of a modal in project style. 

https://github.com/codeforpdx/PASS/assets/37914436/0ba44ff5-8d59-4e6f-9aad-b3830a24a6ca

Add a modal similar to delete contact with 'cancel' and 'delete' options

### Rationale:
This should improve the user experience of editing their profile image, and make the UI more consistent.

TheChunkyPumpkinhead:
## This PR:
Resolves (#463) 
 
**1.**    I fixed the card issue. It is now one card and all items are centered.
**2.** I am having trouble with the copy-and-paste API. 

##Screenshot 2023-11-28 at 3 33 46 PM

I need formal help with API

**1.**    API for copy-and-paste


src/components/NavBar/NavbarDesktop.jsx
```diff
@@ -67,6 +67,7 @@ const NavbarDesktop = ({ setShowConfirmation }) => {
           <NavbarLinks aria-label="navigation links" />
           <Box sx={{ flexGrow: 1 }} />
           <Box aria-label="menu" sx={{ display: { xs: 'none', md: 'flex' } }}>
+            lhgjasdgdk
```
leekahung: Think we can remove this from the file. Not sure what this would be for in the navbar.

src/components/Signup/PodRegistrationForm.jsx
```diff
@@ -18,7 +18,8 @@ import CardHeader from '@mui/material/CardHeader';
 const cardStyle = {
   padding: '8px',
   margin: '8px',
-  boxShadow: '0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19)'
+  textAlign: 'center'
+  // boxShadow: '0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19)'
```
leekahung: Think we can remove the unused styling for this component.

src/components/Signup/PodRegistrationForm.jsx
```diff
@@ -69,7 +70,7 @@ const PodRegistrationForm = ({ register, caseManagerName }) => {
         <p>You are not registering with anyone</p>
       )}
 
-      <Card variant="outlined" sx={cardStyle}>
+      <Card variant="none" sx={cardStyle}>
```
leekahung, line 72: Not sure if the "none" variant just the default card, but think we can remove this prop it is just the default value.

src/pages/Signup.jsx
```diff
@@ -103,6 +102,13 @@ const Signup = () => {
             minWidth: '400px',
             boxShadow: '0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19)'
           }}
+          // /* <Paper
+          //   sx={{
+          //     boxShadow: '0px 10px 4px 0px rgba(0, 0, 0, 0.2)',
+          //     padding: '20px',
+          //     minWidth: '400px'
+          //   }}
+          // >
```
leekahung: Can remove unused code (see feedback from before).

src/pages/Signup.jsx
```diff
@@ -1,9 +1,8 @@
 // React Imports
-import React, { useState, useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import { useSearchParams } from 'react-router-dom';
-
-// Inrupt Imports
-import { getThing, getWebIdDataset, getStringNoLocale } from '@inrupt/solid-client';
+// Inrupt Imports .
```
leekahung: Think we can remove the `.` from the import comments.

leekahung (changes requested):
Hey @TheChunkyPumpkinhead thanks for contributing to the issue! With regards to feedback, see request changes.

src/components/Signup/PodRegistrationForm.jsx
TheChunkyPumpkinhead: I am not sure what i am removing.

leekahung: `variant="none"`. Was wondering if remove that is the same as the default MUI card. If it is, think wouldn't need the prop for `variant`. If we end up needing the prop after all, think we can keep `variant="none"`.

TheChunkyPumpkinhead: Howzit, okay if i take that out, it goes back to the outlined box.



Screenshot 2023-11-29 at 4 55 49 PM

TheChunkyPumpkinhead: Yeah, so i believe we need it. Are you good with me leaving it?

leekahung: Yeah, if it's needed, think we can keep it.

leekahung (approved):
Looks good to me. Think we could get this merged.

One thing to note. Remember to keep your active branch updated in case of changes in Development. Had thought we have reverted the changes for a moment when I swapped branches until I've notice it's was just out of date.

Screen Shot 2023-11-29 at 19 48 36

xscottxbrownx:
I'd like to add one more thing...

lets not just have margin on the top, but also the bottom.  Currently, the footer rests directly up against the rendered parent component.

`src > pages > Signup.jsx` - maybe for the `<Box>` switch the `marginTop: 3` to `my: 3` so there's the same margin on bottom too.  (but honestly the `<Box>` should be a `<Stack>`, then we could get rid of the `display: flex` and the `flexDirection: column` too)

Screenshot 2023-11-29 at 8 05 00 PM

Ah, now that I think about it more and inspect it with Chrome dev tools... Seems like we may not even need the box/stack. May be able to put those stylings on the Paper component itself.  Seems to me it's just an extra div for no real reason.

xscottxbrownx (changes requested):
1 change - see above.

TheChunkyPumpkinhead:
Okay! i need to stop. I did what you asked: 
Screenshot 2023-11-30 at 3 43 32 PM
The only issue is the Typography(I think) needs to display correctly. How do I fix this?
Screenshot 2023-11-30 at 3 47 00 PM

leekahung:
>How do I fix this? Screenshot 2023-11-30 at 3 47 00 PM

Hm...think we could do something like splitting up the one line of text with the statement and webId into two lines of text, one line for the statement, one line for the webId.

To make it nicer, think we could also turn the 2nd line of text with the webId into a potential link (in case the person that's registering is interested in who they're registering/working with with for the new Solid Pod). If the lines get too wide for the screen, think we could employ ellipsis or the like.

xscottxbrownx:
> Okay! i need to stop. I did what you asked: Screenshot 2023-11-30 at 3 43 32 PM 

Actually you did the exact opposite - you put the Paper styling on the Stack, but that's okay it has same effect.

xscottxbrownx:
### Alright, I'm going to back up a lot because I'm a little lost here on this PR.

**This PR hasn't addressed number 1 or 3 from the Issue** #463 

---

**Number 1**
I'm not exactly sure what they are looking for, but it is showing the MUI copycontent rounded icon:
`import ContentCopyRoundedIcon from '@mui/icons-material/ContentCopyRounded';`

so I assume they wanted that icon to be next to the 'YOUR SIGNUP LINK', and maybe the onClick of the icon copies the signup link.
This way the user that wants to share that link with somebody doesn't have to click on it to open it, then manually copy it from address bar?

---

**Number 3**
Still using the window dialog box, instead of an MUI modal to keep with rest of app styling.

---

### With the above said, if you'd like to keep this PR only addressing number 2 from the issue - that would be fine, but we need to:
- edit the title and description of this PR
- edit the title and description of issue #463 
- create new issues for the rest that are not being addressed in this PR

---
---
---

### Now, back to the 1 thing from the issue this PR does address:

I agree with Ka Hung about splitting this text to two lines, and personally I would make the second line a little different (bold maybe 🤷🏻‍♂️) I also think his approach to minimize the text using an ellipsis (if it's too long) is a decent idea, but then it kind of loses all helpfulness and reason to be displayed at all in my opinion.  

This whole display problem becomes even more compounded when we consider mobile devices - which this fails to address or keep in mind. The form has a set width that doesn't fit for mobile whatsoever.

- So that's **issue 1** that I see here. This doesn't work for mobile at all, and is probably the basis to discuss/decide on what text to have here.  Maybe it doesn't render here, but rather on completion of registration?

---

- **Issue 2** is that there's a good amount of cleanup for styling that could still be done. Especially the `PodRegistrationForm`. I'm not sure what happened there, but...
    - It's using `<Card>` for seemingly no good reason, which is why it had card styling that you needed to get rid of with `variant='none'`. The only purpose I see for the `<Card>` is to use it's heading, this can be better accomplished with a `<Typography variant='h5' component='h4'>Set Up A New Pod</Typography>`
    - The `<Typography>` at the top of this component 'Register For PASS' has `display: flex, justifyContent: center, alignItems: center` which could all be replaced with a single `textAlign: center`, **BUT** looking at the card styling below also being `textAlign: center`, could just make the react fragment wrapper around this whole thing a `<Stack textAlign='center'>` and solve all of it. No longer need `cardStyle` or all the `flex` and stuff on the 'Register for PASS' element.
    - All the inputs and labels:  I'm not sure what's going on here.  The two `<FilledInput>'s` & `<InputLabels>'s` have props with aria-labels, but are already labeled by the label using the `htmlFor` and the `id` of the input.  And the top `email` input is a `<TextField>` which handles all of these things.  All 3 should just be `<TextField>'s` used properly. See here: https://mui.com/material-ui/react-text-field/    
    
---

- **Issue 3** is the functionality of the `PodregistrationForm`
    - The `handleSubmit` function run when submitting form has zero error handling. If user types in anything to the inputs and hits the button, they get no feedback what is wrong, why it isn't working, if it's working, etc.
    - The email input needs a `type='email'` for validation. Right now user can type in any text in any format.
    
---

- **Issue 4** the `Signup`:
    - The `<Stack>` doesn't need `display: flex, flexDirection: column`, that is the default for `<Stack>`.
    - With the other things I mentioned above, you wouldn't need the `justifyContent` or `alignItems` on there either.
    - From the new styling on the `<Container>` it is now defeating the purpose of `my: 3` on the `<Stack>` & that is already being defeated by the `margin: auto`.   So, 1 - `margin: auto` probably should be `mx:auto` as I think you are using it to center this item horizontally in the container. 2 - choose to either style the container padding or the stack marginY. They are currently competing each other. Nothing on container and `my: 12` on stack will be about the same visually. So, that large block of styling can be cut down to 5 lines:
    
Screenshot 2023-12-02 at 10 19 09 AM

Keep in mind this still doesn't address the mobile view issues.

leekahung:
Oh, to be able to pick up the potentially long webId, think we could of course use the Clipboard API (or copy and paste functionality) as mentioned in the main issue. That'll be nice I think, since we could reuse that in other places.

xscottxbrownx:
@TheChunkyPumpkinhead are you still interested in this issue?

TheChunkyPumpkinhead:
Yeah

andycwilliams:
@TheChunkyPumpkinhead Hey, just checking in to see what the current status is.

andycwilliams:
Closing this PR. Will open new one pending reassessment of issue.

PR #548, opened 2023-11-28
501/navbar hamburger menu and notification bell improvements

### LINKED ISSUE #501: Bug: Nav bar Hamburger menu and notification bell improvements
### Bug Description:

1. Hamburger menu and bell from nav bar has a green backdrop others have a black opacity. All backdrop colors when modals appear should be uniform

2. menus appear slightly overlapping nav, should start at bottom of nav bar.

3. Text is staggered and icons are different sizes.  (probably because icons are different sizes)

4. Add padding around divider separating settings and log out below routes.

Screenshot 2023-11-07 at 7 35 43 PM

Update existing styling for existing nav menus from MUI such that it removes the excess padding, make nav menu icons consistent, make nav menu appear below navbar instead of the element it's attached to, and update the backdrop color to match the black backdrop of modals when they're opened.

richestrinh:
## This PR:
Resolves #501
 
**1.** Updated positioning for hamburger menu and bell drop-down to appear below nav bar.
**2.** Updated backdrop for hamburger menu and bell.   
**3.** Fixed hamburger menu icons to have consistent sizing.
**4.** Updated spacing around divider seperating routes and settings.

PASS_Issue#501


src/components/NavBar/NavMenu.jsx
```diff
@@ -75,7 +75,14 @@ const NavMenu = ({
...
       onClose={handleMenuClose}
       onClick={handleMenuClose}
-      sx={{ mt: 5, backgroundColor: 'rgba(1, 121, 105, 0.2)' }}
+      slotProps={{
+        paper: {
+          style: {
+            marginTop: '10px'
+          }
+        }
+      }}
```
leekahung, line 88: I'm a bit curious about what this line does specific. Is this for updating the styling the underlying paper component from MUI?

src/components/NavBar/NavMenu.jsx
```diff
@@ -86,7 +93,7 @@ const NavMenu = ({
             >
               <MenuItem
                 component={Button}
-                startIcon={<ContactsIcon />}
+                startIcon={<ContactsIcon style={{ height: '24px', width: '24px' }} />}
```
leekahung: For MUI components, think we could utilize their specialized `sx` prop for custom styling instead of `style` since we're not using HTML elements.

src/components/NavBar/NavMenu.jsx
```diff
@@ -98,13 +105,13 @@ const NavMenu = ({
             >
               <MenuItem
                 component={Button}
-                startIcon={<AccountBoxIcon />}
+                startIcon={<AccountBoxIcon style={{ height: '24px', width: '24px' }} />}
```
leekahung: See comment above about `sx` vs. `style`

src/components/NavBar/NavMenu.jsx
```diff
@@ -113,7 +120,7 @@ const NavMenu = ({
             component={Button}
             startIcon={
               <Badge variant={numUnreadMessages > 0 ? 'dot' : 'standard'} color="error">
-                <EmailIcon />
+                <EmailIcon style={{ height: '24px', width: '24px' }} />
```
leekahung: See comment above about `sx` vs. `style`

src/components/NavBar/NavMenu.jsx
```diff
@@ -124,7 +131,7 @@ const NavMenu = ({
         {/* notifications */}
         <MenuItem
           component={Button}
-          startIcon={<NotificationsIcon />}
+          startIcon={<NotificationsIcon style={{ height: '24px', width: '24px' }} />}
```
leekahung: See comment above about `sx` vs. `style`

src/components/NavBar/NavMenu.jsx
```diff
@@ -157,19 +164,19 @@ const NavMenu = ({
             Profile
           </MenuItem>
         </Link>
-        <Divider />
+        <Divider sx={{ marginY: '10px' }} />
         {/* settings */}
         <MenuItem
           component={Button}
-          startIcon={<SettingsIcon />}
+          startIcon={<SettingsIcon style={{ height: '24px', width: '24px' }} />}
```
leekahung: See comment above about `sx` vs. `style`

src/components/NavBar/NavMenu.jsx
```diff
@@ -157,19 +164,19 @@ const NavMenu = ({
...
           sx={{ color: theme.palette.primary.main, width: '100%' }}
         >
           Settings
         </MenuItem>
         {/* logout */}
         <MenuItem
           component={Button}
-          startIcon={<LogoutIcon />}
+          startIcon={<LogoutIcon style={{ height: '24px', width: '24px' }} />}
```
leekahung: See comment above about `sx` vs. `style`

leekahung (changes requested):
Hey @richestrinh thanks for contributing to changes for PASS! The changes for this PR looks great, just a few details I want this PR to go over. (see feedback)

One other thing that could be addressed here or another PR is the extra padding from the MUI menu and MUI menu list (see clip). If we could get the padding away, think it'll be good.

https://github.com/codeforpdx/PASS/assets/14917816/0d83be54-b170-4406-ab80-8ba5f7abc658

xscottxbrownx (commented):
After Ka Hung's questions/concerns are addressed, I would approve as well. Don't see anything else not already mentioned.

src/components/NavBar/NavMenu.jsx
richestrinh: Yes, you’re correct :)

leekahung:
I think removing the one layer of padding should be sufficient. Removing both layers of padding makes the menu seem too crowded/squished? The spacing between sections (like that of Profile and Settings) looks just right.

Screen Shot 2023-11-30 at 17 35 56

Of course, there's a divider, but think that distance from the bottom of profile to the divider or from the top of settings to the divider looks pretty good. The problem before was that the padding was doubled up when only one layer of padding is needed.

leekahung (changes requested):
I think other than what I've just noticed, think every else looks alright. Think it'll be good to go after the issue with the padding gets resolved (see new comment).

leekahung (approved):
Oh, this is nice! We could of course go back in later and fix the padding of the notification menu once we got a working design (and not a placeholder), but I think the NavMenu is looking pretty great now.

Can certainly approve this for merging. Good work!

xscottxbrownx (approved):

xscottxbrownx (changes requested):
I approved, and then realized there is a difference on full mobile menu (when contacts and civic profile are also in the menu.)

The padding isn't consistent on that one compared to the others that were changed. 

I think it's still the double padding version.

richestrinh:
> I approved, and then realized there is a difference on full mobile menu (when contacts and civic profile are also in the menu.)
> 
> The padding isn't consistent on that one compared to the others that were changed.
> 
> I think it's still the double padding version.

I think it had to do with the min-height for the mobile view.
Here's what it looks like after my fix!

PASS#501_MinHeight

xscottxbrownx (approved):

PR #551, opened 2023-11-30
Issue 509/splitting messages inbox outbox

### LINKED ISSUE #509: [Enhancement] - Migrate Message utility functions to custom hook with React Query

leekahung:
## This PR:

Partially addresses #509 similar to PR #514. This PR is for splitting up the message pages into separate sub-routes, while retaining previous functionalities. Have updated unit tests accordingly and created separate tests for Inbox and Outbox.

https://github.com/codeforpdx/PASS/assets/14917816/84db6895-2000-4580-b361-5f4836d69d9f

- Updates to other Message functionalities in utils similar to what is done in PR #514, but for message building and message sending

**1.**    Issue #509

### DIFF
--- a/src/AppRoutes.jsx
+++ b/src/AppRoutes.jsx
@@ -46,15 +46,15 @@ const AppRoutes = () => {
           <Route path=":webId" element={<Profile />} />
         </Route>
         <Route path="/messages">
-          {MESSAGE_PAGES_LIST.map((messagePages) => (
+          {MESSAGE_PAGES_LIST.map(({ path: routePath, element }) => (
             <Route
-              path={messagePages.path}
-              key={messagePages.path}
+              path={routePath}
+              key={routePath}
               element={
-                messagePages.path === '/messages' ? (
+                routePath === '/messages' ? (
                   <Navigate to="/messages/inbox" replace />
                 ) : (
-                  <MessagesLayout path={messagePages.path}>{messagePages.element}</MessagesLayout>
+                  <MessagesLayout path={routePath}>{element}</MessagesLayout>
                 )
               }
             />
--- a/src/AppRoutes.jsx
+++ b/src/AppRoutes.jsx
@@ -4,9 +4,11 @@ import { Routes, Route, Navigate, Outlet } from 'react-router-dom';
 // Inrupt Imports
 import { useSession } from '@hooks';
 import { SessionProvider } from '@contexts';
+// Component Imports
 import { CIVIC_FORM_LIST, FormLayout } from '@components/CivicProfileForms';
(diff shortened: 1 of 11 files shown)


src/AppRoutes.jsx
```diff
@@ -43,7 +45,23 @@ const AppRoutes = () => {
           <Route index element={<Contacts />} />
           <Route path=":webId" element={<Profile />} />
         </Route>
-        <Route path="/messages" element={<Messages />} />
+        <Route path="/messages">
+          {MESSAGE_PAGES_LIST.map((messagePagesProps) => (
```
milofultz: nit: would this read nicer as `messagePage`? Then we would get `messagePage.path` and `messagePage.element`

src/components/Messages/MessageButtonGroup.jsx
```diff
@@ -23,6 +24,14 @@ const routesArray = [{ label: 'Inbox' }, { label: 'Outbox' }];
  */
 const MessageButtonGroup = ({ setShowModal, boxType, setBoxType }) => {
   const isReallySmallScreen = useMediaQuery('(max-width: 480px)');
+  const navigate = useNavigate();
+
+  const handleNavigate = (pathLabel) => {
+    const path = pathLabel.toLowerCase();
+
+    setBoxType(path);
+    navigate(`/messages/${path}`);
```
milofultz: To decouple the label and the path (thinking about the future, maybe internationalization, etc.), should we add a `path` property to the `routesArray`? e.g.

```
const routesArray = [
  {
    label: 'Inbox',
    path: 'inbox',
  },
  {
    label: 'Outbox',
    path: 'outbox',
  }
];
```

milofultz: Then below could send `handleNavigate(item)` and destructure

milofultz: I understand preoptimization is the death of all code, so take this as you will 🙃

milofultz (approved):

src/AppRoutes.jsx
leekahung: Agreed, definitely a nicer read. I'll have this switched up.

src/components/Messages/MessageButtonGroup.jsx
leekahung: Haha, there's definitely room for improvement. I can switch this up too.

PR #552, opened 2023-11-30
Cleaning up v0.10.0 in prep for v0.11.0 upgrade

leekahung:
## This PR:

Cleans up v10 and address issues in PR for version update (PR #530). Also includes minor changes to other files found when reviewing #530.

### DIFF
--- a/src/layouts/Layout.jsx
+++ b/src/layouts/Layout.jsx
@@ -7,8 +7,8 @@ import Box from '@mui/material/Box';
 import useMediaQuery from '@mui/material/useMediaQuery';
 import { useTheme } from '@mui/system';
 // Component Imports
+import { InactivityMessage, NotificationContainer } from '@components/Notification';
 import { NavBar } from '../components/NavBar';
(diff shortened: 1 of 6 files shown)


src/layouts/Layout.jsx
```diff
@@ -7,8 +7,8 @@ import Box from '@mui/material/Box';
 import useMediaQuery from '@mui/material/useMediaQuery';
 import { useTheme } from '@mui/system';
 // Component Imports
+import { InactivityMessage, NotificationContainer } from '@components/Notification';
```
xscottxbrownx, line 10: if this change was solely for the `@components`, should we do for `NavBar` as well (in the same file)?

xscottxbrownx (approved):
Just one question...

src/layouts/Layout.jsx
leekahung: Sure. Think we could also create a new `index.js` file for `<Footer>` as well.

PR #553, opened 2023-12-02
#549 No whitespace on form submission

### LINKED ISSUE #549: [Enhancement] - Some forms/modals need beginning and trailing whitespace deleted before running the function
Noticed this issue on 2 modals/forms:
**1. Add Contacts**
**2. New Message**

**ADD CONTACTS:**
If user has a whitespace before or after text in `First Name`, `Last Name`, or `Username`, it does not error but leaves the whitespace in the data - not great.

**NEW MESSAGE:**
If user has a whitespace before the text or after the text in the `To:` field, it will error.

---

Need to have these inputs validated, and have beginning and trailing whitespace deleted prior to running functions involved.

---

### Rationale:
Not super noticeable to user why failing (due to whitespace)

---

### Additional context:
https://github.com/codeforpdx/PASS/assets/71395931/a96ee9c9-0a26-4bf7-b132-f525301c094d

milofultz:
Resolves #549
 
**1.** Strips leading/trailing whitespace before submission
**2.** Cleans up some other cruft I found along the way

### DIFF
--- a/src/components/Form/FormSection.jsx
+++ b/src/components/Form/FormSection.jsx
@@ -14,7 +14,7 @@ import { useTheme } from '@mui/material/styles';
  * @name FormSection
  * @param {object} Props - A React prop that consists of
  * that consist of title and children (see {@link formSectionProps})
- * @param Props.id
+ * @param {string} Props.id - The `id` attribute of the element containing the `title`
  * @param {string} Props.title - Title of form section
  * @param {React.ReactElement} Props.children - JSX Element of the wrapped form
  * @returns {React.JSX.Element} - The FormSection Component
--- a/src/components/Form/FormSection.jsx
+++ b/src/components/Form/FormSection.jsx
@@ -14,12 +14,12 @@ import { useTheme } from '@mui/material/styles';
  * @name FormSection
  * @param {object} Props - A React prop that consists of
  * that consist of title and children (see {@link formSectionProps})
- * @param {string} Props.id - The `id` attribute of the element containing the `title`
+ * @param {string} Props.headingId - The `id` attribute of the element containing the `title`
  * @param {string} Props.title - Title of form section
  * @param {React.ReactElement} Props.children - JSX Element of the wrapped form
  * @returns {React.JSX.Element} - The FormSection Component
  */
-const FormSection = ({ title, id, children }) => {
+const FormSection = ({ title, headingId, children }) => {
   const theme = useTheme();
   const isSmallScreen = useMediaQuery(theme.breakpoints.down('sm'));
 
@@ -36,7 +36,7 @@ const FormSection = ({ title, id, children }) => {
         minWidth: '50%'
       }}
     >
-      <Typography variant="h2" align="center" mb={2} sx={{ fontSize: '20px' }} id={id}>
+      <Typography variant="h2" align="center" mb={2} sx={{ fontSize: '20px' }} id={headingId}>
         {title}
       </Typography>
       {children}
--- a/src/components/Form/FormSection.jsx
+++ b/src/components/Form/FormSection.jsx
@@ -14,11 +14,12 @@ import { useTheme } from '@mui/material/styles';
  * @name FormSection
  * @param {object} Props - A React prop that consists of
  * that consist of title and children (see {@link formSectionProps})
+ * @param Props.id
  * @param {string} Props.title - Title of form section
  * @param {React.ReactElement} Props.children - JSX Element of the wrapped form
  * @returns {React.JSX.Element} - The FormSection Component
  */
-const FormSection = ({ title, children }) => {
+const FormSection = ({ title, id, children }) => {
   const theme = useTheme();
   const isSmallScreen = useMediaQuery(theme.breakpoints.down('sm'));
 
@@ -35,7 +36,7 @@ const FormSection = ({ title, children }) => {
         minWidth: '50%'
       }}
     >
-      <Typography variant="h2" align="center" mb={2} sx={{ fontSize: '20px' }}>
+      <Typography variant="h2" align="center" mb={2} sx={{ fontSize: '20px' }} id={id}>
         {title}
       </Typography>
       {children}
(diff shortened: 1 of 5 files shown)


src/components/Modals/AddContactModal.jsx
```diff
@@ -85,14 +86,16 @@ const AddContactModal = ({ addContact, showAddContactModal, setShowAddContactMod
...
   return (
     <Dialog
       open={showAddContactModal}
       aria-labelledby="dialog-title"
       onClose={() => setShowAddContactModal(false)}
     >
-      <FormSection title="Add Contact">
-        <form onSubmit={handleAddContact} autoComplete="off">
+      <FormSection title={formTitle}>
+        <form aria-label={formTitle} onSubmit={handleAddContact} autoComplete="off">
```
xscottxbrownx: I believe `aria-label` should be used when there is no content in the DOM that can be used to label this element.

So, here I believe we should use `aria-labelledby` and reference the `<FormSection>` element which labels this form via it's title.

timbot1789 (approved):

leekahung (approved):
Think I'm good with this too.

src/components/Modals/AddContactModal.jsx
milofultz: Fixed, thank you!

src/components/Modals/NewMessageModal.jsx
```diff
@@ -129,103 +138,106 @@ const NewMessageModal = ({ showModal, setShowModal, oldMessage = '', toField = '
           minHeight: '90%'
         }}
       >
-        <form onSubmit={(e) => handleSubmit(e)} autoComplete="off">
```
milofultz, line 132: This looks like a lot, but I just wrapped the form in a FormSection and removed the typography below.

milofultz:
I added a FormSection to the NewMessage modal because I realized that should probably be done the same way as the other one, and now they both have labels.

src/components/Modals/NewMessageModal.jsx
```diff
@@ -89,7 +90,14 @@ const NewMessageModal = ({ showModal, setShowModal, oldMessage = '', toField = '
       addNotification('error', 'Please enter a message');
```
xscottxbrownx: This wasn't from you, but I don't think line 84-89 get used anymore. I think there is validation on each input now so leaving these blank are not an option.

src/components/Modals/NewMessageModal.jsx
```diff
@@ -129,103 +138,106 @@ const NewMessageModal = ({ showModal, setShowModal, oldMessage = '', toField = '
...
+              type="text"
+              name="title"
+              id="title"
+              onChange={(e) => handleChange(e)}
+              required
+              label="Subject"
+              inputProps={{ maxLength: '48' }}
+              fullWidth
+            />
+            <TextField
```
xscottxbrownx: This text field should be conditionally rendered. Even on a New Message modal, it still shows in the DOM currently.  Maybe `oldMessage && <TextField>`?

src/components/Modals/NewMessageModal.jsx
```diff
@@ -129,103 +138,106 @@ const NewMessageModal = ({ showModal, setShowModal, oldMessage = '', toField = '
...
-              >
-                Cancel
-              </Button>
-              <Button
-                variant="contained"
-                type="submit"
-                color="primary"
-                startIcon={<CheckIcon />}
-                fullWidth
+              InputLabelProps={{
```
xscottxbrownx: I'm not sure this label shrink is needed.  I didn't notice any overlap from label and input value on any device I tested.

src/components/Form/FormSection.jsx
```diff
@@ -14,11 +14,12 @@ import { useTheme } from '@mui/material/styles';
  * @name FormSection
  * @param {object} Props - A React prop that consists of
  * that consist of title and children (see {@link formSectionProps})
+ * @param {string} Props.id - The `id` attribute of the element containing the `title`
  * @param {string} Props.title - Title of form section
  * @param {React.ReactElement} Props.children - JSX Element of the wrapped form
  * @returns {React.JSX.Element} - The FormSection Component
  */
-const FormSection = ({ title, children }) => {
+const FormSection = ({ title, id, children }) => {
```
leekahung: Just a nitpick, but think we could use something like `htmlId` for the prop name as to not get confused for where the `id` attribute is attributed to (the component itself or an element within the component).

xscottxbrownx (changes requested):
See my previous comments, but the big one for me is to make that one conditionally rendered so it's not adding needlessly to the DOM.

milofultz:
> See my previous comments, but the big one for me is to make that one conditionally rendered so it's not adding needlessly to the DOM.

Got all these fixes in! If you're happy, feel free to merge it in.

src/components/Modals/AddContactModal.jsx
```diff
@@ -91,8 +92,8 @@ const AddContactModal = ({ addContact, showAddContactModal, setShowAddContactMod
       aria-labelledby="dialog-title"
       onClose={() => setShowAddContactModal(false)}
     >
-      <FormSection title="Add Contact">
-        <form onSubmit={handleAddContact} autoComplete="off">
+      <FormSection title="Add Contact" id="add-contact-form">
```
leekahung: With the update to using `headingId` for the prop for `<FormSection>` in place of `id` which we'll reserve for elements within the form component, we can update this to `headingId="add-contact-form"`

src/components/Modals/NewMessageModal.jsx
```diff
@@ -129,103 +130,105 @@ const NewMessageModal = ({ showModal, setShowModal, oldMessage = '', toField = '
...
-                fullWidth
-              >
-                Cancel
-              </Button>
-              <Button
-                variant="contained"
-                type="submit"
-                color="primary"
-                startIcon={<CheckIcon />}
+        <FormSection title={oldMessage ? 'Reply To' : 'New Message'} id="new-message-form">
```
leekahung: Think we can update this to use `headingId="new-message-form"`

src/components/Modals/AddContactModal.jsx
milofultz: gah, brain off right now, thank you

leekahung:
Looks like there was a pretty significant divergence that happened, but I got everything patched up. 👍

It should be good to go from my end @milofultz. Think you've address the points I've mentioned. Awaiting other reviewers.

xscottxbrownx (approved):

PR #560, opened 2023-12-09
Issue 524/update profile to account

### LINKED ISSUE #524: [Enhancement] - Update naming for Civic Profile and/or Profile

At the moment there is quite a bit of confusion about the differences in Profile and Civic Profile due to the wording, but I think I have a solution to this. We could have "Profile" be changed to "Account" and "Civic Profile" be changed to "Profile" proper.

[Update 12/09/2023] - Other naming conventions could be used as well to make the two distinct. Several other ideas has been proposed such as:

1. convention for things associated with webIds like "user menu/user settings"
2. utilize alternative wording such as "account"/"settings"
3. utilize prefix words like "Pod" or "PASS" to associate profiles like what's done with "Civic Profile"

### Rationale:

To clear up what we have now with regards to user information and account information.

For this implementation we could simply rename the pages and routes accordingly.

xscottxbrownx:
## This PR:
Resolves #524 
 
**1.**  Changes `profile` route path & file/function/variable names to `account`
**2.**  Changes `civic-profile` route path & file/function/variable names to `profile`

https://github.com/codeforpdx/PASS/assets/71395931/f1a6ef19-7a0b-409d-8f92-8160e689d36e

**1.**  If we also want to change the internal solid files or folders to match (.ttl)
**2.** Changes needed to `useCivicProfile` hook?
**3.** Changes needed to `src` > `model-helpers` > `Profile.js` code around line 130 'civic access'?
**4.** Changes needed to `localStorage.setItem('profileImage')` ? Couldn't tell where 'profileImage' was first stored/created and if it could be changed to 'accountImage'


leekahung:
> 4. Changes needed to localStorage.setItem('profileImage') ? Couldn't tell where 'profileImage' was first stored/created and if it could be changed to 'accountImage'

The initial point where `profileImage` is first stored is `SignedInUserContext`

timbot1789 (changes requested):
I don't think we actually want to change the name of the civic profile to this. IMO it makes the UI more confusing.

This is an object that contains specific, sensitive information that should only be shared with certain people. We need a name that communicates that to users. The purpose of the adjective "civic" in "civic profile" is to communicate this. "Profile" is vague, and not actually all that distinct from an "account." 

---

"Account" itself is also a weird name because that data isn't associated with an account. An account is an entry about you opened in the books at some organization. That's not what happening here. You're instead modifying data on an ID card that has been issued to you.

xscottxbrownx:
> I don't think we actually want to change the name of the civic profile to this. IMO it makes the UI more confusing.
> 
> This is an object that contains specific, sensitive information that should only be shared with certain people. We need a name that communicates that to users. The purpose of the adjective "civic" in "civic profile" is to communicate this. "Profile" is vague, and not actually all that distinct from an "account."
> 
> "Account" itself is also a weird name because that data isn't associated with an account. An account is an entry about you opened in the books at some organization. That's not what happening here. You're instead modifying data on an ID card that has been issued to you.

----

Fair.

1) Maybe we should have some sort of workflow for Issues, then. So people don't spend hours on things that haven't been agreed upon.

2) Do you have any suggestions for naming then?

3) I probably just don't understand the whole concept of why we have a 'civic profile' and a 'profile', as I've missed a lot of dev meetings.  But if it's just some data is private and some is public, why not just have a single profile with two sections - the public section and the private section?

timbot1789:
@xscottxbrownx Let's move this discussion over to the issue.

timbot1789:
@xscottxbrownx Are you opposed to closing this? I don't think we'll be acting on this PR as is. I think we need to talk more about the fundamental design. We can reopen this once we get the design more hashed out.

xscottxbrownx:
Continuing discussion before implementing any changes - closing PR for now.

PR #561, opened 2023-12-10
[Bug] - Include validation for webIds in Add Contacts Modal

### LINKED ISSUE #415: Bug: The "Add contact" modal does not validate if a contact actually exists
Currently, when you add a contact, you are required to add their web ID. However, no validation is done to see if this web ID actually exists, leading to confusing behavior.

Instead, we can make the web ID the first field input into the contact creation. Once they enter that, we can fetch the webId profile. If we get the profile, we can prepopulate the rest of the form with the data from the webId profile. If the profile fetch fails, we put an error in the form saying the web ID is invalid.

Inrupt's pod spaces has an example feature: https://podbrowser.inrupt.com/contacts/add 

image

leekahung:
## This PR:

Resolves #415 by including an early validation step that exits the `handleAddContact` function if `webId` being added does not exist. Minor clean-up to `AddContactModal.jsx` has also been included.

https://github.com/codeforpdx/PASS/assets/14917816/34243825-53e1-4a86-82bc-9a54cc637ad7

## Future Steps:

Related to a new bug, but if you were to include another user with a webId that already exists, but is already included in the contacts list, the new submission will overwrite the information from the existing contact from the list.

https://github.com/codeforpdx/PASS/assets/14917816/79c1b849-9b7d-497b-a90a-3469d7029dd0

### DIFF
--- a/src/components/Modals/AddContactModal.jsx
+++ b/src/components/Modals/AddContactModal.jsx
@@ -50,6 +50,7 @@ const AddContactModal = ({ addContact, showAddContactModal, setShowAddContactMod
   const [userFamilyName, setUserFamilyName] = useState('');
   const [username, setUsername] = useState('');
   const [webId, setWebId] = useState('');
+  const [invalidWebId, setInvalidWebId] = useState(false);
   const [processing, setProcessing] = useState(false);
   const theme = useTheme();
   const isSmallScreen = useMediaQuery(theme.breakpoints.down('sm'));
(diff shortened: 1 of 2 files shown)


leekahung:
Would need to update tests for the modal to include validation

xscottxbrownx:
Just looking at your videos above - I think I'd prefer the error be shown on the Username field instead of the notification hook toast popup.   It would directly show the user which field is incorrect and needs fixing.   

Like triggering the 'error' prop on the field and helper text of 'user does not exist' (or similar) as well.

leekahung:
> Just looking at your videos above - I think I'd prefer the error be shown on the Username field instead of the notification hook toast popup. It would directly show the user which field is incorrect and needs fixing.
> 
> Like triggering the 'error' prop on the field and helper text of 'user does not exist' (or similar) as well.

Got it included, see updated clip in description.

timbot1789 (approved):

PR #574, opened 2024-02-01
Simplify resize tests

timbot1789:
This PR removes a lot of unnecessary code around resizing, and some now unneeded test cases.

In general, I think we should be able to mock `useMediaQuery` instead of modifying the window object in future tests.

### DIFF
--- a/src/components/Footer/Footer.jsx
+++ b/src/components/Footer/Footer.jsx
@@ -2,7 +2,6 @@
 import React from 'react';
 // Material UI Imports
 import Box from '@mui/material/Box';
-import Container from '@mui/material/Container';
 import Divider from '@mui/material/Divider';
 import Stack from '@mui/material/Stack';
 import useMediaQuery from '@mui/material/useMediaQuery';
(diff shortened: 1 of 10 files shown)


src/components/Footer/Footer.jsx
```diff
@@ -21,7 +20,6 @@ import RenderCopyrightAndLinksSection from './RenderCopyrightAndLinksSection';
  */
 const Footer = () => {
   const theme = useTheme();
-  const isSmallScreen = useMediaQuery(theme.breakpoints.down('md'));
   const isReallySmallScreen = useMediaQuery(theme.breakpoints.down('sm'));
```
leekahung, line 23: It's more of a stylistic choice (so it'll probably be for a separate PR if we decide to tackle it), but I wonder if we should start switching the direction of the footer earlier at the `md` (at 900px) breakpoint instead of `sm` (at 600px) @andycwilliams 

Breakpoint at `md`:



Breakpoint at `sm`:

leekahung (approved):
Outside of comments, think the slim-down is good. I can approve this.

PR #580, opened 2024-02-20
[ no issue ] UI tweaks to Profile page

timbot1789:
Small tweaks to the profile page to make it look a little cleaner. We had a few links floating at the top just to have a place to put them. Now they're better integrated into the design.
 
- Moved the "web ID" field and the "signup link" field into the profile card itself.
- Renamed "signup link" to "invite link" to better describe its intended purpose
- added a copy button that copies the invite link to the system clipboard
- Removed some redundant text
- Removed the required field requirement from username in the contacts modal

Before:
image

After:
image

### DIFF
--- a/src/components/Profile/ProfileComponent.jsx
+++ b/src/components/Profile/ProfileComponent.jsx
@@ -124,7 +124,22 @@ const ProfileComponent = ({ contactProfile, webId }) => {
             setInputValue={setNickname}
             edit={edit}
           />
-          <ProfileInputField inputName="WebId" inputValue={webId} />
+          <ProfileInputField
+            inputName="WebId"
+            inputValue={webId}
+            endAdornment={
+              <IconButton
+                aria-label="Copy WebId"
+                edge="end"
+                onClick={() => {
+                  navigator.clipboard.writeText(webId);
+                  addNotification('success', 'webId copied to clipboard');
+                }}
+              >
+                <ContentCopyIcon />
+              </IconButton>
+            }
+          />
         </Box>
         {!contactProfile && (
           <Box sx={{ display: 'flex', flexDirection: 'row', gap: '10px' }}>
--- a/src/components/Profile/ProfileComponent.jsx
+++ b/src/components/Profile/ProfileComponent.jsx
@@ -21,10 +21,11 @@ import ProfileEditButtonGroup from './ProfileEditButtonGroup';
  * @name ProfileComponent
  * @param {object} Props - Props for ClientProfile component
  * @param {object} [Props.contactProfile] - Contact object with data from profile
+ * @param {string} [Props.webId] - The webId of the contact
  * or null if user profile is selected
  * @returns {React.JSX.Element} The UserProfile Component
  */
-const ProfileComponent = ({ contactProfile }) => {
+const ProfileComponent = ({ contactProfile, webId }) => {
   const { session } = useSession();
   const { updateProfileInfo, setProfileData, profileData, fetchProfileInfo } =
     useContext(SignedInUserContext);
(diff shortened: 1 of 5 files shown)


src/components/Profile/ProfileComponent.jsx
```diff
@@ -116,13 +124,29 @@ const ProfileComponent = ({ contactProfile }) => {
...
+              handleCancelEdit={handleCancelEdit}
+              handleEditInput={handleEditInput}
+            />
+            <Typography sx={{ marginTop: '8px' }}>
+              <a href={signupLink} rel="noopener noreferrer" target="_blank">
+                Your Invite Link
+              </a>
+              <IconButton aria-label="Copy Invite Link" edge="end">
+                <ContentCopyIcon
+                  onClick={() => {
```
andycwilliams: I'm getting this error:



It resolved by simply moving 

```
onClick={() => {
  navigator.clipboard.writeText(signupLink);
  addNotification('success', 'Invite link copied to clipboard');
}}
```

to the `ContentCopyIcon`. It still retains functionality.

src/components/Profile/ProfileComponent.jsx
```diff
@@ -116,13 +124,29 @@ const ProfileComponent = ({ contactProfile }) => {
             setInputValue={setNickname}
             edit={edit}
           />
+          <ProfileInputField inputName="WebId" inputValue={webId} />
```
andycwilliams: The `WebId` does not display in full. The user has to manually scroll to the right in order to see the entire thing. Not sure if we care about that. Or if we're just going to display the username.



There are some other things that come to mind, such as when clicking on the edit button it stretches out the box but none of the fields extend along with it.

The content copy icon could maybe be done in line with the WebId itself? MUI should has an option for that. The `AddContactModal` already does, although that is not currently functional.

It does this with InputProps

```
InputProps={{
  endAdornment: (
    <InputAdornment position="end">
      <IconButton aria-label="Copy WebId" edge="end">
        <ContentCopyIcon />
      </IconButton>
    </InputAdornment>
  )
}}
```

But I'm wondering if all this could just be left for another PR. I would like UX to go over the whole profile page, which would not be necessary to get this one done.

andycwilliams (commented):
Looking better. Just a couple ideas.

src/components/Profile/ProfileComponent.jsx
timbot1789: Fixed

src/components/Profile/ProfileComponent.jsx
timbot1789: Added a copy to clipboard button next to the web ID. 

The web ID can often get quite long, so we will need to cut it off somewhere. Cutting it off at the same point as the other fields seems reasonable.

Changing the webid when you click "edit" may make the user think the webID is editable, which it is not.

I'd also like to get feedback from UX about what they think about what they think of the app, but I haven't heard anything from them in a while. In the mean time, we can come up with our own designs.

src/components/Profile/ProfileComponent.jsx
```diff
@@ -116,13 +124,46 @@ const ProfileComponent = ({ contactProfile }) => {
...
+            inputName="WebId"
+            inputValue={webId}
+            endAdornment={
+              <IconButton
+                aria-label="Copy WebId"
+                edge="end"
+                onClick={() => {
+                  navigator.clipboard.writeText(webId);
+                  addNotification('success', 'webId copied to clipboard');
+                }}
```
leekahung, line 137: Think it's nice we got the clipboard now for these links.

Was wondering if we should convert this as a reusable function that could potentially be used on other pages like Contacts.

leekahung (approved):
Just saw the changes with the links and the copy function. It's pretty nice.

Code looks good, so I'm good with approving this. Just a quick comment on the copy function (see feedback).

src/components/Profile/ProfileComponent.jsx
timbot1789: Not a bad idea. Would be a good first issue for someone.

PR #581, opened 2024-02-20
[ no issue ] Remove unused buttons from navbar

timbot1789:
Remove unused buttons from navbar. 

They're placeholders for features we were considering at one point. However, it's not clear with the current feature set if these buttons would ever be used (the notifications bell doesn't match with current UX trends).

It also makes onboarding more difficult because new users can't tell if the button isn't supposed to work, or if they're doing something wrong.

Before:

image
 ---
image

After:
image
---
image


leekahung (approved):
Looks good to me. 👍

PR #583, opened 2024-02-21
[no issue] Update inrupt libraries to version 2

timbot1789:
Inrupt updated its libraries to version 2 a little while ago. It has, among other things, a way to log out of both the app and the pod server at the same time. Let's update, since the more advanced logout will be helpful. Also, the documentation has been updated to reflect the new version.

However, the documentation updates have been haphazard, so not all the tutorials have been moved to the new version.

### DIFF
--- a/src/contexts/SessionContext.jsx
+++ b/src/contexts/SessionContext.jsx
@@ -27,7 +27,7 @@ import {
   logout,
   handleIncomingRedirect,
   getDefaultSession,
-  onSessionRestore as onSessionRestoreClient
+  EVENTS
 } from '@inrupt/solid-client-authn-browser';
 
 import { getProfileAll, getPodUrlAll } from '@inrupt/solid-client';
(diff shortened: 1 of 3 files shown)


src/contexts/SessionContext.jsx
```diff
@@ -27,7 +27,7 @@ import {
   logout,
   handleIncomingRedirect,
   getDefaultSession,
-  onSessionRestore as onSessionRestoreClient
+  EVENTS
```
timbot1789, line 30: The main breaking change was getting rid of the "onSessionRestore", "onLogin" functions, and replacing them with `session.events.on()`. This is the only breaking change that affects our code.

leekahung (commented):
I've just tested out the features on PASS to confirm if anything broke. Most seems to work just fine.

There's one feature that seems to break though, that being new account registration via the invite link (see clip). I think that would need to be addressed, though probably in a separate PR.

https://github.com/codeforpdx/PASS/assets/14917816/a2b65179-9163-464b-be30-19f03dc32bac

timbot1789:
> I've just tested out the features on PASS to confirm if anything broke. Most seems to work just fine.
> 
> There's one feature that seems to break though, that being new account registration via the invite link (see clip). I think that would need to be addressed, though probably in a separate PR.
> Screen.Recording.2024-02-22.at.1.52.41.PM.mov

This upgrade is actually motivated by wanting to make enhancements to that flow, which I'm working on in a separate branch. However _that bug in your console_ shouldn't be appearing. I'll take a closer look and probably fix it here.

PR #584, opened 2024-02-25
464 - Simplify Signup Flow

### LINKED ISSUE #464: Bug: Sign-up Link registration signs up with wrong name and webId
### Bug Description:

Existing registration flow sign-up with the wrong name and webId. It uses the name and webId from the person they sign up with instead when adding to their Contact list.

### Reproduction Steps:    
**1.**    Registers new user
**2.**    See bug after navigating to Contacts page

### Expected Behavior:

Should sign up user with their own first and last name with their webId, not the first and last name and webId of the person that they sign up with. The information added to the Contact list should be related to the person that's being registered not the person who they sign up with.

https://github.com/codeforpdx/PASS/assets/114430929/ed68ca69-a9a5-4a40-8e34-713347a26617

https://github.com/codeforpdx/PASS/assets/114430929/1b66fd94-449f-4352-8c64-5b22a5f3e204

timbot1789:
## This PR:
Resolves #464 
 
**1.** Rebuilds the user registration logic using community solid server V7's [automation API](https://communitysolidserver.github.io/CommunitySolidServer/latest/usage/account/json-api/#controlsaccountclientcredentials). This allows us to build out a user's entire pod with no action needed from the user other than username and password. This system only works on Community Solid Server V7, but that has been the case.
**2.** Adds a couple more states to the registration page. Now the page can show: creation form, loading page, and finish page. A future PR will need to add an error page for if there's an error in pod creation
**3.** Added a signup button to the home screen that takes a new user to the signup process. This won't set them up with a case manager, but it will make new volunteer onboarding significantly easier.
**4.** Added some code to outline the IDP logout feature that comes with inrupt's V2 client auth libraries (allows logout from both pod and app in one click). However, this is proving complicated, especially to deploy, so finishing the feature will be left to another PR. I don't want the code to sit on my computer forever, though.

We will need to improve the error handling of the signup flow in a future PR. This PR only implements the happy path. It doesn't handle any errors during pod creation.


config/solid-server/solid-config.json
```diff
@@ -1,35 +1,37 @@
...
-    "css:CommunitySolidServer/config/ldp/modes/default.json",
-    "css:CommunitySolidServer/config/storage/keyvalue/storages/storages.json",
-    "css:CommunitySolidServer/config/util/auxiliary/empty.json",
-    "css:CommunitySolidServer/config/util/identifiers/suffix.json",
-    "css:CommunitySolidServer/config/util/index/default.json",
-    "css:CommunitySolidServer/config/util/logging/winston.json",
-    "css:CommunitySolidServer/config/util/representation-conversion/default.json",
-    "css:CommunitySolidServer/config/util/resource-locker/file.json",
-    "css:CommunitySolidServer/config/util/variables/default.json"
+    "css:config/app/init/static-root.json",
```
timbot1789, line 4: Created using new CSS  config generator:
https://communitysolidserver.github.io/configuration-generator/v7/

env.template
```diff
@@ -1,3 +1,4 @@
 VITE_SOLID_IDENTITY_PROVIDER="http://localhost:3000/"
 VITE_SOLID_POD_SERVER="http://localhost:3000/"
 VITE_SUGGESTED_OIDC_OPTIONS="http://localhost:3000/, https://opencommons.net/, https://solidcommunity.net/, https://login.inrupt.com/, https://inrupt.net/"
+VITE_CLIENT_ID_DOC="http://localhost:3000/clientAppId.json"
```
timbot1789, line 4: The client ID is necessary for allowing proper redirects when logging in or logging out. However, it must be hosted on the internet to properly use. This hosting issue won't be resolved here.

package.json
```diff
@@ -95,7 +96,7 @@
     "vitest": "^0.34.6"
   },
   "lint-staged": {
-    "*.{js,css,ts,tsx,jsx}": [
+    "src/**/*.{js,css,ts,tsx,jsx}": [
```
timbot1789, line 99: Fix for vite config getting linted and causing the commit to fail.

public/clientAppId.jsonld
```diff
@@ -0,0 +1,11 @@
+{
```
timbot1789, line 1: See [here](https://docs.inrupt.com/developer-tools/javascript/client-libraries/tutorial/authenticate-client/) for more information on this document

src/AppRoutes.jsx
```diff
@@ -32,14 +31,7 @@ const AppRoutes = () => {
...
       />
-      <Route
-        path="/signup"
-        element={
-          <SessionProvider>
-            <Signup />
-          </SessionProvider>
-        }
-      />
+      <Route path="/signup" element={<Signup />} />
```
timbot1789, line 34: The signup now manages its own state entirely (it doesn't interact with the session) so we don't need to mask the session or force the user to log out.

src/components/Signup/podSignupHelpers/initializePod.js
```diff
@@ -0,0 +1,90 @@
+// Inrupt Imports
```
timbot1789, line 1: This was incorrectly labeled as a `jsx` file. It's a pure JS file.

src/components/Signup/podSignupHelpers/registerPod.js
```diff
@@ -0,0 +1,102 @@
+import {
```
timbot1789, line 1: This was incorrectly labeled as a jsx file. It's a pure js style

src/contexts/SessionContext.jsx
```diff
@@ -57,7 +57,7 @@ export const SessionProvider = ({
 
   useEffect(() => {
     if (onSessionRestore !== undefined) {
-      onSessionRestoreClient(onSessionRestore);
+      session.events.on(EVENTS.SESSION_RESTORED, onSessionRestore);
```
timbot1789, line 60: Fixes for Inrupt's v2 library upgrade. `onSessionRestoreClient` and all other `onEvent` functions have been removed and replaced with this `session.events.on` style.

vite.config.js
```diff
@@ -2,40 +2,34 @@ import { defineConfig } from 'vite';
 import react from '@vitejs/plugin-react';
 import visualizer from 'rollup-plugin-visualizer';
 import { resolve } from "path";
-import { nodeResolve } from '@rollup/plugin-node-resolve'; 
-
-// Workaround for this open solid-client-authn bug:
-// https://github.com/inrupt/solid-client-authn-js/issues/3001
-// Once that issue is fixed and the solid-client-authn-js version is bumped
-// We can delete this function and all related code
-function withCustomResolver(
```
timbot1789, line 11: Inrupt's V2 libraries now support ES Modules out of the box, so this custom resolver is no longer needed.

src/components/Signup/ShowNewPod.jsx
```diff
@@ -32,6 +34,10 @@ const ShowNewPod = ({ oidcIssuer }) => {
           ? `You have registered with ${data[0].person}`
           : 'You have not registered with a case manager'}
       </Typography>
+      <Typography>Click here to return to the home page and log into PASS:</Typography>
```
andycwilliams, line 39: Maybe upon successful registration it could countdown and then redirect automatically? I've seen plenty of instances along those lines (e.g. "[action successful], you will be redirected in [X] seconds or click here")

src/components/Signup/podSignupHelpers/initializePod.js
andycwilliams: Is there a functional difference in this case, or is it more a matter of best practices?

src/pages/Signup.jsx
```diff
@@ -29,28 +27,34 @@ import { PodRegistrationForm, ShowNewPod, initializePod, registerPod } from '@co
...
+        caseManagerWebId,
+        caseManagerFirstName: caseManagerNames[0],
+        caseManagerLastName: caseManagerNames[caseManagerNames.length - 1]
+      },
+      registration.fetch
+    );
+    setStep('done');
   };
 
   const loadProfileInfo = async () => {
```
andycwilliams, line 62: Receiving this error when first clicking on Sign Up

andycwilliams (commented):
Haven't had a chance to go through all the changes. These are just immediate thoughts.

Slightly tangential: I'm thinking about contributing to #547, which I believe affects the same `Signup` component. Thus it might be an opportunity to reassess that and its related issue (#463), as the needs may have changed since then.

It may also be a good idea to add a Cancel/Return button on the Sign Up page. Or perhaps that could be part of the above PR instead.

leekahung:
I've cleared out my local_temp_server_files directory to test out the new build before diving into the codebase changes. The new signup functionality seems to be running fine, I see the new pods being created from the process. One small issue I've saw pop-up was the webId not being shown (see screenshot).

Screenshot 2024-02-25 at 4 37 17 PM

I've made a few more accounts to test things out, but it looks like the problem above was related to who logged in last as it show the previous logged in user's webId (see clips).

(Previous user logged in was `http://localhost:3000/newtestuser/profile/card#me`)

https://github.com/codeforpdx/PASS/assets/14917816/b36ba706-c738-4f1b-ade1-80c301949ce7

(Previous user logged in was `http://localhost:3000/testpod/profile/card#me`)

https://github.com/codeforpdx/PASS/assets/14917816/ff3aac94-18b6-4503-adcb-40b5892b1c93

If the server is fresh and no one had logged in last, it makes sense that the webId was blank for the first pod creation, but it should have been the webId of the user creating the webId, not the previous logged in user.

Another thing I've noticed when signing up using that button was that the .acl file for `/PASS/Documents` failed to generate for some of the new users. It returns a 409 error instead when logging in for the first time. Without it, we can't control the permissions for "Documents".

Screenshot 2024-02-25 at 4 39 04 PM

If we were to re-login (logout and login again), the .acl file gets generated as it should if the file was missing (see screenshot), but the function for generating the .acl file should have been running correctly the first time around, so I'm not sure what happened there.

Screenshot 2024-02-25 at 4 43 28 PM

If I were to signup for a pod using the signup forms by CSS instead, the 409 error doesn't seem to appear at all (see screenshot). The .acl file gets generated the first time around.
 
Screenshot 2024-02-25 at 4 45 41 PM

Something not related to this PR (but an earlier one), is that user deletion fails to trigger when the first and last name are null. I'll be posting a separate bug report for that (see issue #585).

Other than the issues I've seen above, all other functionalities seem to be work correctly like document uploads/preview/deletion, adding new contacts, sending messages, etc. Sharing documents still work as long as the .acl file exists within `/PASS/Documents/` inside the user's pod

src/components/Signup/podSignupHelpers/initializePod.js
timbot1789: Best practice. Also makes dev tools understand it better.

src/components/Signup/ShowNewPod.jsx
timbot1789: I agree we can definitely improve this page. It's a placeholder right now before we settle on a real design. A timed countdown is one approach.

I wrote up an issue here related to it:
https://github.com/codeforpdx/PASS/issues/441

One of the ideas I had was to include an overview of what the pod is, and what kind of things it lets you do. This wouldn't work with the timer idea, since the overview would take a lot of time to read, and users may want to be linked to other resources to read more.

Either way it's not related to this PR. We can continue the conversation on that issue.

src/pages/Signup.jsx
timbot1789: This should have only happened when clicking the signup button. Not when using the invite link.

This is now fixed.

timbot1789:
@leekahung Thanks for the thorough testing. Fixed the web ID display and the pod URL display (it was broken too).

Added the doclist to pod initialization. It's an empty document. The ACL doesn't need to be created at this time. The doclist inherits permissions from the PASS pod, and the document sharing feature creates an acl if one doesn't already exist.

leekahung:
> @leekahung Thanks for the thorough testing. Fixed the web ID display and the pod URL display (it was broken too).
> 
> Added the doclist to pod initialization. It's an empty document. The ACL doesn't need to be created at this time. The doclist inherits permissions from the PASS pod, and the document sharing feature creates an acl if one doesn't already exist.

All good, I'll give it another go through post changes.

leekahung:
Alright, just gone through the changes. The previous errors I was getting before for the features are now gone. So things are looking good there. I'll take a look over the code.

src/components/Signup/ShowNewPod.jsx
```diff
@@ -5,33 +5,38 @@ import React from 'react';
 import Typography from '@mui/material/Typography';
 
 // Hooks imports
-import useSession from '@hooks/useSession';
 import { useContactsList } from '@hooks';
+import { Button } from '@mui/material';
```
leekahung, line 9: Very minor, recommend import style to be `import Button from '@mui/material/Button'` to be consistent with other MUI imports

src/pages/Signup.jsx
```diff
@@ -11,14 +11,12 @@ import Box from '@mui/material/Box';
...
 
-// Custom Hook Imports
-import useContactsList from '@hooks/useContactsList';
-import useSession from '@hooks/useSession';
 // Constant Imports
 import { ENV } from '@constants';
 
 // Signup Form Imports
 import { PodRegistrationForm, ShowNewPod, initializePod, registerPod } from '@components/Signup';
+import { Typography } from '@mui/material';
```
leekahung, line 19: Very minor nitpick similar to the previous comment about MUI imports and consistency.

leekahung (approved):
Outside of two minor syntax comments, the rest of the code looks good to me!

It's not too important now for development, but for the sign-up process, I was wondering if we should give a disclaimer somewhere pointing out when this feature could be used since it's limited to just servers with CSS v7 vs. when users should just use the login button to get to the sign-up page for their respective servers like with solidcommunity.net's server or Inrupt's PodSpaces to make their Pods.

Screenshot 2024-02-25 at 11 50 41 PM
Screenshot 2024-02-25 at 11 50 01 PM

timbot1789:
> Outside of two minor syntax comments, the rest of the code looks good to me!
> 
> It's not too important now for development, but for the sign-up process, I was wondering if we should give a disclaimer somewhere pointing out when this feature could be used since it's limited to just servers with CSS v7 vs. when users should just use the login button to get to the sign-up page for their respective servers like with solidcommunity.net's server or Inrupt's PodSpaces to make their Pods.
> 

With the current design, the user doesn't get to choose their pod provider during pod creation flow. They're required to use the pod provider configured in the running PASS instance. In that case, the disclaimer would need to be sysadmin/dev facing. "You can only connect this part of the app to a CSS v7 server." I've added disclaimers in the comments of the registration function code, which is sufficient unless we want to write a sysadmin guide in the wiki.

You're right though that we need to display the Pod Provider that PASS signed the user up with at the end. I added a line to the last page that shows the URL of the pod provider.

We can also add an option to let a user enroll with an existing pod. But that's a separate flow where we can expect the user to be able to handle pod creation without our help, so no disclaimer is needed. I just added [an issue](https://github.com/codeforpdx/PASS/issues/586) to the project for it. I had this feature in a demo a while back, but haven't written up tests for it yet.

PR #594, opened 2024-03-01
Revert "576 readme.md and wiki getting started update"

timbot1789:
This change isn't up to project standards. Git is not an image hosting service, nor will we be using it as such. There are several other possible approaches that solve the problem of hosting images for wikis that use tools better fit for the task. If this image hosting functionality is needed, then you may research other available options.


andycwilliams:
Could we retain the changes to the README, or is that not an option when doing a revert?

timAtSource:
> Could we retain the changes to the README, or is that not an option when doing a revert?

Perhaps, but it's best to revert. The other issue here is permissions a user receives when they get a commit merged to the project. They gain permission to run actions. We should not grant this to people who have not demonstrated understanding of the basics of the project. 

This is not a quality commit for a new project member. They will need to find a new task.

timbot1789:
> Could we retain the changes to the README, or is that not an option when doing a revert?

Perhaps, but it's best to revert. The other issue here is permissions a user receives when they get a commit merged to the project. They gain permission to run certain actions on the repo. We should not grant this permission to people who have not demonstrated understanding of the basics of the project.

This is not a quality commit for a new project member. They will need to find a new task.

Jared-Krajewski (approved):
I think we can revert this and re submit a PR with just the readme changes. We will re host the images for the wiki in google drive, I think they are worth keeping and may make it easier for new people to understand the setup process. Although I'm not sure what the permissions have to do with this particular PR vs any other one from new members.. I'm going to suggest we review our contribution security and policy in the very near future, preferably Tuesdays dev meeting.

timbot1789:
> I think we can revert this and re submit a PR with just the readme changes. We will re host the images for the wiki in google drive, I think they are worth keeping and may make it easier for new people to understand the setup process. Although I'm not sure what the permissions have to do with this particular PR vs any other one from new members.. I'm going to suggest we review our contribution security and policy in the very near future, preferably Tuesdays dev meeting.

Why are we not able to upload the images to the wiki directly? I've done that on a few of the pages.

The readme updates themselves are also outdated. It doesn't reflect the actual startup flow since we upgraded to community solid server v7. The setup flow has been greatly simplified, and we should update our readme to reflect that. The changes here will quickly need to be rewritten.

Jared-Krajewski:
> > I think we can revert this and re submit a PR with just the readme changes. We will re host the images for the wiki in google drive, I think they are worth keeping and may make it easier for new people to understand the setup process. Although I'm not sure what the permissions have to do with this particular PR vs any other one from new members.. I'm going to suggest we review our contribution security and policy in the very near future, preferably Tuesdays dev meeting.
> 
> Why are we not able to upload the images to the wiki directly? I've done that on a few of the pages.
> 
> The readme updates themselves are also outdated. It doesn't reflect the actual startup flow since we upgraded to community solid server v7. The setup flow has been greatly simplified, and we should update our readme to reflect that. The changes here will quickly need to be rewritten.

I had issue when trying to add images directly to the wiki before so I had hosted everything from me on google drive. If they can go directly into the wiki that would be great. As far as I was aware the wiki's required linked image hosting.

Edit: apparently you can just drop the image into the wiki page you are editing now. that sure is convenient. 

as far as the setup flow changes.. I was not aware of that, but that makes sense.

PR #600, opened 2024-03-17
598 simplify add contact to use username

### LINKED ISSUE #598: [Enhancement] - Simplify the add Contact Form to only provide the username to look up the webid
In its current form PASS is still not very user friendly. You have to provide the whole web Id from the ```https``` to the ```#card``` to lookup the user. This is mainly for cases where the pod servers are different and URL needs to be changed to fit those pods.

image

### Proposed Changes:
One approach is to:

- Add a list selection for the most common servers such as ```solidcommunity```, ```opencommons```, etc
- Also have option to enter custom server URL such as ```Other```
- Then ask for the username and populate the webId string based on the server given and the username

### Possible Drawbacks:
Discuss any potential drawbacks or negative impacts this enhancement might have, including:

### Alternatives Considered:
Mention any alternative solutions or enhancements you've considered. Why were they not chosen or why are they less desirable than your proposal?

DionSat:
## This PR:
Resolves #598
 
In its current form PASS is still not very user friendly. You have to provide the whole web Id from the https to the #card to lookup the user. This is mainly for cases where the pod servers are different and URL needs to be changed to fit those pods.

This pull request replaces the WebID field with a username and OIDC provider dropdown to make the application more user friendly. The WebID can still be selected by clicking the ```Other``` dropdown for entering a custom WebID.

image

Tooltip:
image

image

image

### DIFF
--- a/src/components/Modals/AddContactModal.jsx
+++ b/src/components/Modals/AddContactModal.jsx
@@ -21,11 +21,10 @@ import useMediaQuery from '@mui/material/useMediaQuery';
 import { useTheme } from '@mui/material/styles';
 // Custom Hook Imports
 import useNotification from '@hooks/useNotification';
-// Component Imports
-import { ENV } from '@constants';
-import { FormSection } from '../Form';
-
 // Constant Imports
+import { ENV } from '@constants';
+// Component Imports
+import { FormSection } from '../Form';
 
 // @memberof Modals
 // @name renderWebId
@@ -63,14 +62,6 @@ const AddContactModal = ({ addContact, showAddContactModal, setShowAddContactMod
   const [oidcProviders] = useState([...ENV.VITE_SUGGESTED_OIDC_OPTIONS.split(', '), 'Other']);
   const [Oidc, setOIDC] = useState('');
 
-  const oidcForm = {
-    'http://localhost:3000/': 'http://localhost:3000/user/profile/card#me',
-    'https://opencommons.net/': 'http://opencommons.net/user/profile/card#me',
-    'https://solidcommunity.net/': 'https://user.solidcommunity.net/profile/card#me.',
-    'https://login.inrupt.com/': 'https://id.inrupt.com/user',
-    'https://inrupt.net/': 'https://id.inrupt.com/user'
-  };
-
   const clearInputFields = () => {
     setUserGivenName('');
     setUserFamilyName('');
@@ -105,7 +96,7 @@ const AddContactModal = ({ addContact, showAddContactModal, setShowAddContactMod
       };
     } else {
       userObject = {
-        webId: oidcForm[Oidc].replace('user', userName.trim()).trim(),
+        webId: JSON.parse(ENV.VITE_OIDC_WEBIDS)[Oidc].replace('user', userName.trim()).trim(),
         ...(addUserGivenName.value && { givenName: addUserGivenName.value.trim() }),
         ...(addUserFamilyName.value && { familyName: addUserFamilyName.value.trim() })
       };
--- a/src/components/Modals/AddContactModal.jsx
+++ b/src/components/Modals/AddContactModal.jsx
@@ -55,10 +55,12 @@ const AddContactModal = ({ addContact, showAddContactModal, setShowAddContactMod
   const [userFamilyName, setUserFamilyName] = useState('');
   const [webId, setWebId] = useState('');
   const [invalidWebId, setInvalidWebId] = useState(false);
+  const [userName, setUserName] = useState('');
+  const [customWebID, setCustomWebID] = useState(false);
   const [processing, setProcessing] = useState(false);
   const theme = useTheme();
   const isSmallScreen = useMediaQuery(theme.breakpoints.down('sm'));
-  const [oidcProviders] = useState(ENV.VITE_SUGGESTED_OIDC_OPTIONS.split(', '));
+  const [oidcProviders] = useState([...ENV.VITE_SUGGESTED_OIDC_OPTIONS.split(', '), 'Other']);
   const [Oidc, setOIDC] = React.useState('');
 
   const clearInputFields = () => {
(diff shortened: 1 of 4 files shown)


DionSat:
? I just ran all the tests on my side but its still failing the test check. That's weird it passes the tests it says it fails on my side. So I don't know what's going on with the checks.

DionSat:
I figured it out. There is no environmental variable when GitHub actions it setting up the environment. GitHub actions sets up the dependencies and builds the project and runs the test but doesn't create an ```.env``` file. Found a similar issue with GitHub actions. Because my test relies on the ```.env``` file we need to either create secret variable in GitHub or add ```env:``` parameters to the GitHub actions. Unfortunately this would require changing the workflow file or I would have to change my branch to not use the ```.env``` file. But I think this should be addressed now because I assume in the future we'll have tests that rely on the ```.env``` file so we should address it now.

test/components/Modals/AddContactModal.test.jsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render, screen } from '@testing-library/react';
+import { render, screen, getByRole as testGetByRole } from '@testing-library/react';
```
timbot1789, line 2: I _think_ the CI/CD test failures are caused by this `getByRole` import. That appears to be a function from the [DOM testing library](https://testing-library.com/docs/dom-testing-library/intro), which forms the basis for a lot of the framework specific libraries. I don't think this function is playing nicely with the React Version, [React Testing Library](https://testing-library.com/docs/react-testing-library/intro). The react library handles rendering differently than DOMTL, and I don't think these methods can be mixed and matched. (This is probably actually a bug in the framework itself. If we can get a minimally reproducible example, it may be worth reporting.)

I think if you change from importing `getByRole` to using the getByRole function on either the screen or the renderer, things should start behaving more consistently.

`screen.getByRole('combobox')`

or preferably:
```javascript
{getByRole} = <ReactComponent />
const elt = getByRole('combobox');
```

timbot1789:
> I figured it out. There is no environmental variable when GitHub actions it setting up the environment. GitHub actions sets up the dependencies and builds the project and runs the test but doesn't create an `.env` file. Found a similar issue with GitHub actions. Because my test relies on the `.env` file we need to either create secret variable in GitHub or add `env:` parameters to the GitHub actions. Unfortunately this would require changing the workflow file or I would have to change my branch to not use the `.env` file. But I think this should be addressed now because I assume in the future we'll have tests that rely on the `.env` file so we should address it now.

In that case, you can mock the ENV in tests.

Add whatever values you need here: https://github.com/codeforpdx/PASS/blob/077756394a32d60c0b2dcde8d4472838a9ca3970/__mocks__/src/constants/index.js#L1

Then add this line after the imports in your test file:
```javascript
vi.mock('@constants');
```

And the values will appear in the tests.

Note: This is actually why we wrap ENV as a constant variable in PASS. It lets us treat the env as an external service like a database or a network connection. Then it's more clear how to mock the ENV in tests.

src/components/Modals/AddContactModal.jsx
```diff
@@ -49,28 +55,61 @@ const AddContactModal = ({ addContact, showAddContactModal, setShowAddContactMod
...
   const [invalidWebId, setInvalidWebId] = useState(false);
+  const [userName, setUserName] = useState('');
+  const [customWebID, setCustomWebID] = useState(false);
   const [processing, setProcessing] = useState(false);
   const theme = useTheme();
   const isSmallScreen = useMediaQuery(theme.breakpoints.down('sm'));
+  const [oidcProviders] = useState([...ENV.VITE_SUGGESTED_OIDC_OPTIONS.split(', '), 'Other']);
+  const [Oidc, setOIDC] = React.useState('');
+
+  const oidcForm = {
```
timbot1789: Rather than hardcode these in the component, I think it would make sense to place them in the ENV variable in some way. That way both the options for default OIDC providers, and their web ID templates are in the same place.

DionSat:
I tried using that mock of the ENV file in the manual mocks folder. But it doesn't seem to work. I tried to mock('@constants') but it didn't really populate the ENV like you would expect. Now I did find a way using this:
```vi.mock('../../../src/constants/', async () => {
  const actual = await vi.importActual('../../../src/constants/');
  return {
    ...actual,
    ENV: {
      VITE_SOLID_IDENTITY_PROVIDER: 'https://www.testurl.com/',
      VITE_SUGGESTED_OIDC_OPTIONS:
        'http://testurl_1.com/, http://testurl_2.com/, http://testurl_3.com/'
    }
  };
});
```

This will work but another problem is that you would have to call it each time for the test file that needs the ENV. Why not just set the env value in the setupFile before each test. I found you can just call ```process.env.[ENV_VARIABLE] = [STRING]``` in the setupFile to setup the ENV variables before each test. The nice thing about the setupFile I added is that it runs before each test file. This seems a lot simpler and only requires you define the ENV variables once.

andycwilliams:
Just had a thought. We may want to have a tooltip or a clickable question mark next to the OIDC Provider field that explains what it does. We could have one for each other field as well, though those seem more intuitive.

src/components/Modals/AddContactModal.jsx
DionSat: Makes sense, I had a similar idea. I think I can move them out safely into the env file. I would have to update the test again to include the new env variables but should be easy.

DionSat: Okay I was able to pull out the WebID and store them in the ```.env``` file.

DionSat:
As for the tool tip I can add that as well. I've seen a couple of those on some websites and I do think the OIDC provider dropdown should have one to make it easier to understand what we're looking for.

leekahung (approved):
I had just tried out the branch locally and everything seems to run fine. 👍 The use of a select field and tooltips was a great choice.

Was a bit confused at first when it didn't work, but I remembered the environment variables had been updated.

andycwilliams:
Didn't check this until now. Is the commented out meant to stay in? (Lines 30-38 on the AddContactModal)

PR #601, opened 2024-03-19
No Issue - Fix sharing of one document

timbot1789:
The sharing of an individual document seems to have broken some time back in November. This PR fixes one document permission sharing.

Testing:

1. Upload a document to your pod (e.g. some_document.pdf)
2. share that document with another pod
3. Check that an ACL file has been properly generated for this file (some_document.pdf.acl)

### DIFF
--- a/test/utils/session-core.test.js
+++ b/test/utils/session-core.test.js
@@ -39,7 +39,7 @@ describe('setDocContainerAclPermission', () => {
     expect(sessionHelpers.setDocAclForUser).toBeCalledWith(
       session,
       expectedContainerUrl,
-      'update',
+      '',
       expectedWebId,
       permissions
     );
--- a/src/components/Documents/DocumentTable.jsx
+++ b/src/components/Documents/DocumentTable.jsx
@@ -154,7 +154,7 @@ const DocumentTable = ({ handleAclPermissionsModal, handleSelectDeleteDoc }) => 
       getActions: (data) => [
         <GridActionsCellItem
           icon={<ShareIcon />}
-          onClick={() => handleAclPermissionsModal('documents', data.name, data.type)}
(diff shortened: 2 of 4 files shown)


leekahung (approved):
Can confirm that the file-specific .acl file is being generated from the changes. 👍

Screenshot 2024-03-18 at 9 11 11 PM (3)

I assume `universalAccess.setAgentAccess` would neatly replace the custom `setDocAclForUser` made in PASS (at least in the case up create/updating file-specific .acl files)?

timbot1789:
> Can confirm that the file-specific .acl file is being generated from the changes. 👍
> 

Thanks for looking at this so fast!
> I assume universalAccess.setAgentAccess would neatly replace the custom setDocAclForUser made in PASS (at least in the case up create/updating file-specific .acl files)?

I believe so. It works for individual documents, but not quite for containers. This method doesn't currently have the ability to set default/inherited permissions. See this thread on the solid community forum: https://forum.solidproject.org/t/inherited-default-access-using-the-universal-api/7353/7. Whereever we are setting file-specific ACLs, we can use `setAgentAccess` instead.

If the limitation in `setAgentAccess` is removed in the future, then we could fully replace most of our custom ACL controls with it.

test/utils/session-core.test.js
```diff
@@ -39,7 +39,7 @@ describe('setDocContainerAclPermission', () => {
     expect(sessionHelpers.setDocAclForUser).toBeCalledWith(
       session,
       expectedContainerUrl,
-      'update',
+      '',
```
timbot1789, line 42: Changed this to '' to better communicate that this value doesn't do anything. The only value that matters here is "create," which triggers a separate creation flow. All other strings have the "update" effect.

PR #609, opened 2024-04-02
586 feature request add sign up with existing pod option to the sign up page

### LINKED ISSUE #586: [Feature Request] - Add "Sign up with existing POD" option to the Sign up page
### Feature Description:
Currently, if you click the signup button in PASS, or if you get an invite link from a PASS user, your only option is to create a new pod. We should allow people to sign up with an existing pod as well.

The page currently looks like this:
Screenshot 2024-03-03 at 3 19 59 PM

We should add a new form underneath that lets a user put in their existing pod:
Screenshot 2024-03-03 at 3 24 17 PM

The POD Provider area will take in the URL of the user's chosen pod provider. Once inputting this, they can click the login button to be directed to their provider for sign-in. This flow will be very similar to the current login to PASS flow. Once they are signed in, they will be redirected back to PASS, where we can begin the same POD setup flow we do when creating a new POD. Both flows, the "Get a new POD" signup and the "use an existing POD" signup, should end at the same screen, the `ShowNewPod` component, which currently looks like this:
Screenshot 2024-03-03 at 3 27 02 PM

Improvements to this screen are out of scope of this issue.

russfraze:
## This PR:
Created the register with existing pod form on the signup page which accepts the URL to a pod provider. 

## Test steps

1. Create a new pod with https://opencommons.net/
2. On the Pass signup page, enter https://opencommons.net/ in the pod provider field in the register with existing pod form. 
3. Click the login to pod provider button. 
4. You should be redirected to https://opencommons.net/
5. Login and you should be redirected back to pass. Confirm that you are signed in.


tim-transparent-ai:
This looks good overall. I was able to sign up with an existing pod, and the data was initialized.

2 requests:
1. Can the page switch over to the ShowNewPod component, or a slightly updated ShowSetup component?
2. Can you add a unit test that shows the RegisterWithExistingPod calls the expected functions?

timbot1789:
This looks good overall. I was able to sign up with an existing pod, and the data was initialized.

2 requests:

1. Can the page switch over to the ShowNewPod component, or a slightly updated ShowSetup component?
2. Can you add a unit test that shows the RegisterWithExistingPod calls the expected functions?

russfraze:
@timbot1789 I think everything should be working properly now. I could use a little help on the unit test, specifically mocking the login function. I am looking at the OidcLoginComponent as a reference and I can see that the login function is imported, but I can't see how it's mocked with vi.fn or spyOn. 

In that component, the assertion is just expect(login).toBeCalled(); which works there but throws an error in my test. The error is that it expects me to have called spy. I'd love your feedback and a little guidance on the test, and I can finish this up.

russfraze:
@timbot1789 ready for review.

src/components/Signup/ExistingPodForm.jsx
```diff
@@ -0,0 +1,73 @@
...
+
+    localStorage.setItem('oidcIssuer', oidcIssuer);
+    await login({
+      oidcIssuer,
+      redirectUrl
+    });
+  };
+
+  return (
+    <Card style={cardStyle}>
```
leekahung: Not a big issue since it still works as intended, but MUI components have a dedicated `sx` attribute for custom CSS instead of `style` used for regular HTML elements. It lets MUI recognize any other potential themes that might be in place. https://mui.com/system/getting-started/the-sx-prop/

src/components/Signup/ExistingPodForm.jsx
```diff
@@ -0,0 +1,73 @@
...
+      <CardHeader title="Register with Existing Pod" />
+      <form onSubmit={loginHandler} style={formRowStyle}>
+        <TextField
+          id="existing-pod-provider"
+          style={textFieldStyle}
```
leekahung: Same comment on `style` vs `sx`

leekahung (changes requested):
Hey Russ! Just a minor feedback about MUI props/attributes `style` vs `sx`. Otherwise, I think I could approve this after that is addressed.

russfraze:
@leekahung thanks for that feedback. I changed it so the styles are in the sx prop for the MUI components. Just a heads up, I think the reason I did that in the first place is because I copied the styles from the PodRegistration component which still uses the style prop for its MUI components. For consistency, we could change that component too. 

I think this PR should be good to go now.

leekahung (approved):
Of course, we could always come back later to update the styling more,  but it's good enough for now. Approving.

PR #611, opened 2024-04-03
Enhancement update document sharing modal to pull from contacts list (#573)

### LINKED ISSUE #573: Enhancement: Update document sharing modal to pull from contacts list
Currently, to share a document with someone, you must enter their pod URL. This is hard for users because it requires them to remember the Pod URL of everyone they want to share with. Conceptually it is incorrect as well, because you do not share with a pod, you share with a Web ID 
image

Rework the sharing modal to share with people off their contact list. This can use the same functionality as currently exists in the messaging system (a dropdown that searches the contact's first and last name).
image

### Additional context:
This may require removing logic around adapting the "pod url." This logic is unnecessary. If we have a person's web Id, we can use it directly.

joshua-cornett:
## This PR:
Resolves #573
 
**1.** Document sharing (SetAclPermissionsModal) uses Web ID instead of Pod URL   
**2.** Populates a dropdown of contacts and associated Web IDs
**3.** _Side improvement_: Header of FormSection no longer overflows

## Screenshots:

image

image

image

image

image

### DIFF
--- a/src/components/Form/FormSection.jsx
+++ b/src/components/Form/FormSection.jsx
@@ -28,7 +28,7 @@ const FormSection = ({ title, headingId, children }) => (
       mx: '3px',
       padding: '16px',
       minWidth: '50%',
-      maxWidth: '75vdw'
+      maxWidth: '75dvw'
     }}
   >
     <Typography variant="h2" align="center" mb={2} sx={{ fontSize: '20px' }} id={headingId}>
--- a/src/components/Form/FormSection.jsx
+++ b/src/components/Form/FormSection.jsx
@@ -27,7 +27,8 @@ const FormSection = ({ title, headingId, children }) => (
       alignItems: 'center',
       mx: '3px',
       padding: '16px',
-      minWidth: '50%'
+      minWidth: '50%',
+      maxWidth: '75vdw'
     }}
   >
     <Typography variant="h2" align="center" mb={2} sx={{ fontSize: '20px' }} id={headingId}>
--- a/src/components/Modals/SetAclPermissionsModal.jsx
+++ b/src/components/Modals/SetAclPermissionsModal.jsx
@@ -54,7 +54,7 @@ const SetAclPermissionsModal = ({ showModal, setShowModal, dataset }) => {
     label: `${contact.person} ${contact.webId}`,
     id: contact.webId
   }));
-  const shareeName = data?.filter(
+  const shareName = data?.filter(
     (contact) => permissionState.webIdToSetPermsTo === contact.webId
   )[0];
 
@@ -124,7 +124,7 @@ const SetAclPermissionsModal = ({ showModal, setShowModal, dataset }) => {
         }
       >
         <form onSubmit={handleAclPermission} autoComplete="off" style={{ width: '100%' }}>
-          <FormControl required fullWidth sx={{ maxWidth: '75vdw', marginBottom: '1rem' }}>
+          <FormControl required fullWidth sx={{ maxWidth: '75dvw', marginBottom: '1rem' }}>
             <InputLabel id="permissionType-label">Select One</InputLabel>
             <Select
               labelId="permissionType-label"
@@ -150,7 +150,7 @@ const SetAclPermissionsModal = ({ showModal, setShowModal, dataset }) => {
             fullWidth
             required
             autoFocus
-            value={shareeName?.person ?? permissionState.webIdToSetPermsTo}
+            value={shareName?.person ?? permissionState.webIdToSetPermsTo}
             disablePortal
             autoSelect
             options={contactListOptions}
--- a/src/components/Modals/SetAclPermissionsModal.jsx
+++ b/src/components/Modals/SetAclPermissionsModal.jsx
@@ -123,8 +123,8 @@ const SetAclPermissionsModal = ({ showModal, setShowModal, dataset }) => {
           dataset.modalType === 'container' ? 'Share All Documents' : `Share ${dataset.docName}`
         }
       >
-        <form onSubmit={handleAclPermission} autoComplete="off">
-          <FormControl required fullWidth sx={{ marginBottom: '1rem' }}>
+        <form onSubmit={handleAclPermission} autoComplete="off" style={{ width: '100%' }}>
+          <FormControl required fullWidth sx={{ maxWidth: '75vdw', marginBottom: '1rem' }}>
             <InputLabel id="permissionType-label">Select One</InputLabel>
             <Select
               labelId="permissionType-label"
@@ -145,6 +145,7 @@ const SetAclPermissionsModal = ({ showModal, setShowModal, dataset }) => {
             id="set-acl-to"
             name="setAclTo"
             data-testid="newShareWith"
+            sx={{ marginBottom: '1rem' }}
             freeSolo
             fullWidth
             required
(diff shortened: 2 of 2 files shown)


src/components/Form/FormSection.jsx
```diff
@@ -27,7 +27,8 @@ const FormSection = ({ title, headingId, children }) => (
       alignItems: 'center',
       mx: '3px',
       padding: '16px',
-      minWidth: '50%'
+      minWidth: '50%',
+      maxWidth: '75vdw'
```
leekahung: Minor thing, but I believe you mean the `dvw` unit here for device viewport width?

src/components/Modals/SetAclPermissionsModal.jsx
```diff
@@ -107,8 +123,8 @@ const SetAclPermissionsModal = ({ showModal, setShowModal, dataset }) => {
           dataset.modalType === 'container' ? 'Share All Documents' : `Share ${dataset.docName}`
         }
       >
-        <form onSubmit={handleAclPermission} autoComplete="off">
-          <FormControl required fullWidth sx={{ marginBottom: '1rem' }}>
+        <form onSubmit={handleAclPermission} autoComplete="off" style={{ width: '100%' }}>
+          <FormControl required fullWidth sx={{ maxWidth: '75vdw', marginBottom: '1rem' }}>
```
leekahung: Same unit here, think it should be `dvw` instead of `vdw`.

src/components/Modals/SetAclPermissionsModal.jsx
```diff
@@ -39,16 +40,27 @@ import useNotification from '../../hooks/useNotification';
...
     permissionType: ''
   });
   const [processing, setProcessing] = useState(false);
 
+  const { data } = useContactsList();
+  const contactListOptions = data?.map((contact) => ({
+    label: `${contact.person} ${contact.webId}`,
+    id: contact.webId
+  }));
+  const shareeName = data?.filter(
```
leekahung: Double `e` in variable name, should be `shareName`

src/components/Modals/SetAclPermissionsModal.jsx
```diff
@@ -125,28 +141,34 @@ const SetAclPermissionsModal = ({ showModal, setShowModal, dataset }) => {
...
+          <Autocomplete
+            id="set-acl-to"
+            name="setAclTo"
+            data-testid="newShareWith"
+            sx={{ marginBottom: '1rem' }}
+            freeSolo
+            fullWidth
+            required
+            autoFocus
+            value={shareeName?.person ?? permissionState.webIdToSetPermsTo}
```
leekahung: Double `e` in variable name, should be `shareName`

leekahung (commented):
Very minor issues, think it should be good to go after they're addressed

src/components/Modals/SetAclPermissionsModal.jsx
joshua-cornett: That was intentional, though I'm not sure if sharee is _actually_ a word haha. Meant it as the person being shared with. For readability though I'll make it shareName either way.

src/components/Modals/SetAclPermissionsModal.jsx
joshua-cornett: Ah, good catch

joshua-cornett:
> Very minor issues, think it should be good to go after they're addressed

Understood, taking care of it now - thanks for reviewing!

src/components/Modals/SetAclPermissionsModal.jsx
leekahung: I see, sharee like mentee, got it. No problem. 👍

src/components/Modals/SetAclPermissionsModal.jsx
```diff
@@ -125,28 +141,34 @@ const SetAclPermissionsModal = ({ showModal, setShowModal, dataset }) => {
...
+            sx={{ marginBottom: '1rem' }}
+            freeSolo
+            fullWidth
+            required
+            autoFocus
+            value={shareName?.person ?? permissionState.webIdToSetPermsTo}
+            disablePortal
+            autoSelect
+            options={contactListOptions}
+            onChange={(event, newValue) => {
```
leekahung: I don't think `event` is used here.

If that's the case, we could replace this with an `_`, `_event`, or `_e` to keep things consistent with the other event parameters that's used in event calls.

src/components/Modals/SetAclPermissionsModal.jsx
```diff
@@ -125,28 +141,34 @@ const SetAclPermissionsModal = ({ showModal, setShowModal, dataset }) => {
...
+              setPermissionState({
+                ...permissionState,
+                webIdToSetPermsTo: newValue.id ?? newValue
+              });
+            }}
+            placeholder={permissionState.podUrlToSetPermissionsTo}
+            error={permissionState.podUrlToSetPermissionsTo === podUrl}
```
leekahung: The state `permissionState` in line 47 don't have the values `podUrlToSetPermissionsTo` set to them. I believe it's supposed to be `webIdToSetPermsTo` now that podUrl got updated to webIds.

As for the placeholder, I don't think we need that anymore since we could just use the text from renderInput, which is "WebID to share with" and the text itself is pretty helpful.

leekahung: As for Line 164, the condition should be `permissionState.webIdToSetPermissionsTo === webId` since the webId would never equal podUrl, the error condition would never be triggered for the field.

leekahung (commented):
Seem to have missed a few other things the first time around (apologize for the slip), but this really should be it. In case, I've missed something, I've turned on the `@ts-check` flag locally just to make sure.

leekahung:
> > Very minor issues, think it should be good to go after they're addressed
> 
> Understood, taking care of it now - thanks for reviewing!

Hey Josh, thanks for bearing with me! Just a few more things, but I've made sure this time that it really should be good to go.

joshua-cornett:
Ah I see, not sure how I missed those. I'll take a look and get them sorted.

src/components/Modals/SetAclPermissionsModal.jsx
joshua-cornett: I opted for _

src/components/Modals/SetAclPermissionsModal.jsx
joshua-cornett: - renderInput isn't defined until the component is rendered, so I hardcoded `placeholder={"WebID to share with"}`

joshua-cornett: > The state `permissionState` in line 47 don't have the values `podUrlToSetPermissionsTo` set to them. I believe it's supposed to be `webIdToSetPermsTo` now that podUrl got updated to webIds.
> 
> As for the placeholder, I don't think we need that anymore since we could just use the text from renderInput, which is "WebID to share with" and the text itself is pretty helpful.

- renderInput isn't defined until the component is rendered, so I hardcoded `placeholder={"WebID to share with"}`

joshua-cornett:
Should be set, re-requesting review

leekahung (approved):
Nice changes! It looks good to me, approving.

leekahung:
I'll have this PR merged into the development branch. 👍

Issue #605, opened 2024-03-28
[Enhancement] - Site/Application Theme Updates


leekahung:
### Describe the Current Behavior/Feature:

With updates from the UX/UI team Figma board (see https://www.figma.com/file/sqbOHqUVtmuWvlN6KLujnt/Pass-Designs?type=design&node-id=1216-2986&mode=design&t=lc9UZaWY194NPspe-0), we could start implementing some of the changes from there. As a start, we could begin with the UI theming, specifically on the main font and color palettes using our existing `theme.js`.

As there are currently three separate groups of font-sizes (one for mobile, two for desktop), we can do size updates in a follow-up issue.

### Rationale:

Integrating design changes with existing UI

leekahung:
Had recently discovered during our in-person meeting the location of the new color palette on Figma, which would be beneficial as we change up the site theme.

![Screenshot 2024-04-25 at 7 10 27 PM (2)](https://github.com/codeforpdx/PASS/assets/14917816/081fc8db-27cb-4287-bfcc-b7055ffa64ca)

Also updated the issue description to narrow the scope.

JaeGif:
@leekahung I've nearly done with just the color theming and have been working on the Profile page CSS to get it closer to the Figma frames. I think what I'll do is finish just the coloring for this issue, and open another issue to finish the profile page theming. 

For the color scheme, I've updated theme.js with the new color codes and am incrementally swapping the old and new schemes. Some components CSS needs to be changed so they're accessing the correct theme as well since the themes were renamed. 

Wrangling MUI components takes a minute 😵‍💫 so I'm happy to move the component CSS to a separate issue. I can imagine tests needing to be rewritten with layout changes as well.

PR #612, opened 2024-04-09
607 enhancement update the profile card

### LINKED ISSUE #607: [Enhancement] - update the profile card
Update the profile card

### Rationale:
Implement the desired figma design with a few notable changes.

### Proposed Changes:
- maintains the existing fields on the profile card (Name, Nickname, Webite, edit, invite link) and profile image
- uses the figma design
- should be one card (rather than parent component having the box-shadow/border around it instead of its children components)
- should be responsive
   - vertical stack the image smaller screens
   - horizontal for large screens

image

For example, for desktop its orientation would be horizontal:

Vertial for mobile:

Further enhancement that would make UI cleaner would be to make the profile image/avatar the edit button as suggested by @andycwilliams instead of the standalone button below the image. (See example below)

Screenshot 2024-04-08 at 5 12 49 PM

Imagine "Edit Profile Image"/"Edit Avatar" instead of "View Profile".

The fields found in the Figma design, while valid, are not meant to be for the Profile page which act more like a public profile in social media (see images below).

Screenshot 2024-04-08 at 5 07 26 PM

Screenshot 2024-04-08 at 5 09 08 PM

Instead, those found in the Figma design should be personal fields used for the Civic Profile, which is on a separate page/route. For that, we could have a separate issue opened for it using the Figma design for the card as is or with other potential fields.

russfraze:
## This PR:
Resolves #607 

Screenshot 2024-04-09 at 3 42 00 PM

I made the separate profile image card and input fields into one card, which is still responsive. I also adjusted the input fields and the share and add buttons to be aligned left like the UX team's design. 

I could not find the exact icon used for the edit button, so I used one from the Material Icons that looks similar. If the original icon from the Figma file is desired, please link to the source. 

I also made the invite link align to the right, but another option would be to align it with the left of the input fields. 

This component will look more like the designs in the Figma files once issue #605 is complete. 
There are also additional comments in this issue around UI updates to the profile picture, which I will handle as next steps.


src/components/Profile/ProfileComponent.jsx
```diff
@@ -107,6 +108,11 @@ const ProfileComponent = ({ contactProfile, webId }) => {
             gap: '10px'
           }}
         >
+          <ProfileEditButtonGroup
+            edit={edit}
+            handleCancelEdit={handleCancelEdit}
+            handleEditInput={handleEditInput}
+          />
```
leekahung: The component that shows the buttons for updating profile information should only be available to the user. They should not have access to this button group in other people's profiles. That was the reason behind the component being inside the condition `!contactProfile`.

If we want to move this out of the box below, make it render with the same condition similar to the invite link.

src/components/Profile/ProfileInputField.jsx
leekahung, line 1: Currently the updates here cause the label and the text field to be misaligned. I'm also a bit confused as to why an additional Box component had been used for wrapping the Input component.

Won't it be better with how it was before with a single Box/flex box with two components vertically centered within it and using a space-between to space out the two components horizontally? (see the screenshot below using the old layout with styling added for justifyContent and fixed widths for the Input component)

Screenshot 2024-04-09 at 7 20 08 PM

src/components/Profile/ProfileEditButtonGroup.jsx
```diff
@@ -26,8 +27,7 @@ const ProfileEditButtonGroup = ({ edit, handleCancelEdit, handleEditInput }) =>
   <Box
     sx={{
       display: 'flex',
-      gap: '10px',
```
leekahung, line 29: While removing the `marginTop` seems good since we are moving the button group to a little icon in the corner, think we'll still need the gap between buttons. At the moment the buttons are sticking together without the gap.

leekahung (commented):
Hey Russ, just dropping some feedback on the updates. It looks like some of the unit tests are breaking for `ProfileComponents`, but I think it's mainly due to us reshuffling components around.

src/components/Profile/ProfileInputField.jsx
russfraze: Yes, that's much better. Thanks for the feedback.

leekahung (approved):
Okay, things seem to be in running order with the style changes and updates. I can approve it for merging.

russfraze:
With this PR, styling changes were made to the profile card to match the current UX design.

PR #613, opened 2024-04-11
571 edit contacts

### LINKED ISSUE #571: Enhancement: Allow editing of Contacts
Currently, a user is able to create, read, and delete a contact. But they are not able to easily edit an existing contact.

Add a way for users to edit the entries in their contacts list. This can be a button in the row of the contacts table with the [MUI edit icon](https://mui.com/material-ui/material-icons/?query=edit&selected=Edit). This button can then launch "Create Contact" modal with the title changed to "Update Contact." We can use this modal to modify the first name, last name, and web ID of an entry in the contact list.

### Additional Context
Since the Web ID is the identifier for the object, updating the web ID will likely require deleting the old contact entry, and then creating a new contact entry.

ofu997:
## This PR:
Resolves #571 

✅ prepopulate form when editing. 
✅ if webId is different, delete current contact and add contact. Having issues with re-render
✅ warning if webId already exists


ofu997:
now doing linting errors

src/components/Contacts/ContactListTable.jsx
```diff
@@ -151,8 +175,28 @@ const ContactListTable = ({ contacts, deleteContact }) => {
...
+    /> */}
     </Box>
   );
 };
 
+// const Console = props => {
+//   const contacts = props;
+//   console.table(contacts);
+//   return false;
+// }
```
leekahung: I think this is a temporary component used for debugging? We could remove it now that it's no longer in use.

src/components/Modals/AddContactModal.jsx
```diff
@@ -62,6 +75,28 @@ const AddContactModal = ({ addContact, showAddContactModal, setShowAddContactMod
...
+  // try to add new contact to pod A, which already has a contact
+  // -result: updates existing contact
+  // add new contact to pod B, which doesn't have a contact
+  //   -result: adds to grid
+  // current commit : Edit function will update contact names if webId remains the same.
+  // If user edits the webId, a new contact will be created.
+
+  // ✅ prepopulate form when editing.
+  // ✅ if webId is different, delete current contact and add contact. Having issues with re-render
+  // ✅ warning if webId already exists
```
leekahung: I don't think we need the whole implementation detail in here, maybe a one line comment for where the functionality is

src/components/Contacts/ContactListTable.jsx
```diff
@@ -94,6 +106,15 @@ const ContactListTable = ({ contacts, deleteContact }) => {
       headerAlign: 'center',
       align: 'center'
     },
+    {
```
leekahung, line 116: More of a styling thing than a functionality issue, but the new icon seems to have a different color when compared to the other ones. Would be nice to have the colors match up.

Screenshot 2024-04-16 at 6 19 31 PM

src/components/Modals/AddContactModal.jsx
leekahung, line 1: I believe there are additional changes made in this modal component so that it'll include an edit function. However, it doesn't seem to fill in the existing information of the user you wish to edit.

Screenshot 2024-04-16 at 6 23 22 PM

It'll probably be better if the form starts with values from the table row. I think it might be due to not being able to handle cases when the First Name/Last Name is null/undefined.

src/components/Modals/AddContactModal.jsx
```diff
@@ -104,6 +143,15 @@ const AddContactModal = ({ addContact, showAddContactModal, setShowAddContactMod
...
 
     try {
+      if (userObject.webId !== originalWebId && contactToEdit) {
+        if (await contactWebIds.includes(userObject.webId)) {
+          addNotification('error', 'Web ID exists. Edit appropriate contact');
+          return;
+        }
+        const toDelete = contacts.find((item) => item.webId === originalWebId);
+        await handleDeleteContact(toDelete);
+        setDeleteViaEdit(true); // attempt to re-render
```
leekahung: This function seems to delete the contact when editing and submitting without a username by mistake.

https://github.com/codeforpdx/PASS/assets/14917816/dd9e99de-a1b1-4644-af9e-3787a8f88cef

Seems to be related to a validation bug we found with the modal (see Issue #614), but we probably don't want the contact to be deleted if the information is incorrect, just a warning and error stating invalid information. The modal would stay open until the user exits or enters the right information.

ofu997:
@leekahung Can this issue be handed off to another contributor. I am stuck on the rendering issue.

leekahung:
> @leekahung Can this issue be handed off to another contributor. I am stuck on the rendering issue.
> 

No problem, I'll have this unassigned

JaeGif:
> @leekahung Can this issue be handed off to another contributor. I am stuck on the rendering issue.

Hi, I took up the issue to finish it. What rendering issue are you referring to?

ofu997:
@JaeGif 
[Discord link](https://discord.com/channels/1068260532806766733/1078526377378205736/1243079350161506376)

src/components/Modals/AddContactModal.jsx
JaeGif: Currently it fills First and Last name from the table. Getting the OIDC provider and the username is proving to be a little more convoluted. Do we want to try to autofill those as well or leave them for the user the enter?

leekahung: For OIDC provider, think we should try and use the same one as the original used by the WebID. Add for username, think we can use some sort of pattern or regex to capture it.

The username should either be right after the providerb (like with localhost and other CSS servers) or embedded (like with solidcommunity.net).

If proven to difficult, we could potentially include username as part of the contact user object

leekahung: Actually, got a better idea. Since the user webId and provider would not be the things that are edited, we could disable from editing. The provider and user name would just be passed in using parts from the existing webId.

PR #618, opened 2024-04-19
614/add contact validation

### LINKED ISSUE #614: [Bug Report] - Add Contact Validation
### Describe the bug:

Currently, validation for the `Add Contact` modal seems to be broken (see screenshot below). When clicking on "Add Contact", without the username for the new contact to be added, it breaks down and get stuck on the modal with the function for `onSubmit` getting triggered.

While users can escape with the cancel button or ESC key, the "Add Contact" button is no longer accessible.

### Expected behavior:

Form validation requires users to have a valid username before it can submit the form for "Add Contact".

### Screenshots:

Screenshot 2024-04-16 at 6 30 46 PM

### To Reproduce:
1. Go to Contacts
2. Click on "Add Contacts"
3. Insert information only in the optional fields
4. Click on "Add Contact" button
5. See error

### Possible Fix:

Include validation as part of the input field for username.

JaeGif:
## This PR:
Resolves #614 
 
- Added required tags to non-optional fields.
- Disabled submit button if mandatory fields not filled out
- Added tests to cover disabled submit

### DIFF
--- a/src/components/Modals/AddContactModal.jsx
+++ b/src/components/Modals/AddContactModal.jsx
@@ -1,5 +1,5 @@
 // React Imports
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 // Inrupt Imports
 import { getWebIdDataset } from '@inrupt/solid-client';
 // Material UI Imports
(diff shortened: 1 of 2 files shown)


leekahung (approved):
Nice! Was able to see validation and the disabled button when the mandatory fields are not filled out.

PR #623, opened 2024-04-29
605/theme updates

### LINKED ISSUE #605: [Enhancement] - Site/Application Theme Updates

JaeGif:
## This PR:
Resolves #605
 
- theme.js updated with new color scheme
- added 2 new fonts to theme.js
- color scheme applied to major site elements
- started adding/changing elements to match the Figma board on Profile page
- started adding/changing button elements to match Figma board on Contacts page

This PR is just the beginning of the work that needs to be done to overhaul the UI. With the addition of new themes and fonts, the next steps can be broken into multiple issues. I'd recommend making an issue per page to be re-themed since most pages require the addition of new elements/functionality and rewriting unit tests. Some components (like the DocumentList component) will need major overhauls and might warrant their own issue.

Also should contact the UI team and see if there's any more/other designs incoming before we get too far on that.


src/components/Modals/UploadDocumentModal.jsx
```diff
@@ -133,6 +139,7 @@ const UploadDocumentModal = ({ showModal, setShowModal }) => {
         onConfirm={confirmationModalType === 'add' ? handleDocAdd : handleDocReplace}
         uploadType={confirmationModalType}
       />
+      km
```
leekahung: There's an additional `km` string in the modal (see upper left)

Screenshot 2024-04-29 at 11 30 08 AM

src/components/Modals/UploadButtonGroup.jsx
```diff
@@ -91,27 +92,37 @@ const UploadButtonGroup = ({ file, setFile }) => {
...
+        startIcon={<FileUploadOutlinedIcon />}
       >
-        Choose file
-        <input
-          type="file"
-          hidden
-          accept=".pdf, .docx, .doc, .txt, .rtf, .gif, .png, .jpeg, .jpg, .webp"
-        />
+        Upload File
+        <input type="file" hidden accept=".pdf, .docx, .doc, .txt, .rtf" />
```
leekahung, line 105: There could be files users wish to update that are not text/pdf files, but also image files. That was why we had additional extensions like png and jpg accepted as acceptable file inputs.

JaeGif: I changed this one just because the Figma design has a separate button for media files

src/components/Modals/UploadDocumentModal.jsx
```diff
@@ -17,14 +17,19 @@ import { DatePicker } from '@mui/x-date-pickers/DatePicker';
...
+/* import Slide from '@mui/material/Slide';
+ */ // Context Imports
 import { DocumentListContext } from '@contexts';
 // Component Imports
 import { DocumentSelection, FormSection } from '../Form';
 import UploadButtonGroup from './UploadButtonGroup';
 import useNotification from '../../hooks/useNotification';
 import UploadDocumentConfirmationModal from './UploadDocumentConfirmationModal';
-
+/*
```
leekahung, line 28: Is this `Transition` component gonna be used for an up coming PR for modal/drawer animation?

src/components/Modals/UploadButtonGroup.jsx
```diff
@@ -91,27 +92,37 @@ const UploadButtonGroup = ({ file, setFile }) => {
...
       </Button>
 
+      {/*       <Button
```
leekahung: A bit confused as to why the buttons were split up in this PR?

src/pages/Contacts.jsx
```diff
@@ -76,15 +79,42 @@ const Contacts = () => {
...
-          color="secondary"
-          size="small"
-          startIcon={<AddIcon />}
-          onClick={() => setShowAddContactModal(true)}
-        >
-          Add Contact
-        </Button>
+        <Box sx={{ display: 'flex', justifyContent: 'space-between' }}>
+          <FormControl sx={{ minWidth: 120 }} size="small">
+            <Select
```
leekahung: I believe this `Select` field is here for mobile and not the desktop version?

Since we'll no longer have access to the MUI table's sorting capabilities when we switch to cards, would make sense to have that we'll need one for a custom sorting function in follow-up PRs.

Screenshot 2024-04-29 at 11 40 03 AM

src/components/Modals/UploadDocumentModal.jsx
JaeGif: Screenshot 2024-04-29 at 11 40 59 AM

Yeah, the button grouping, changing the file type, icons, transition etc are first steps toward achieving this ^ design. I started these before the scope of the issue changed.

src/components/Modals/UploadButtonGroup.jsx
leekahung: Think we should add the new button in too if that's the case since there would be times when people just upload images. It might look a little funky though since it's no longer 2 over 2, but 3 over 2.

src/pages/Contacts.jsx
JaeGif: Do you mean it's only supposed to be there for a mobile version?

leekahung:
Related to this PR, there still a bit of theming we could update from the Inbox/Outbox buttons, namely the custom styling from React Paginate (see `MessageFolder.jsx` for the component and `PaginationStyles.js` for the custom styles and classes used).

Screenshot 2024-04-29 at 11 46 36 AM (2)

I believe Ben at the time made it so that the styling and coloring were in line with what we had before. Think we could do the same thing here with the new theme.

leekahung (changes requested):
Alright, think that touches most of the things that still needed to be addressed. But outside of those, the changes made with the new theme looks great!

src/components/Modals/UploadButtonGroup.jsx
JaeGif: Gotcha, added the upload photos button back in and changed the layout to column for now so it doesn't look too funky. There's a unit test that counts the number of buttons rendered, so I updated it to count 3 now.

src/pages/Contacts.jsx
leekahung: Think we technically could have the sorting trigger on Desktop as well from the `Select` field, but that might be duplicate functionality?

src/components/Modals/UploadButtonGroup.jsx
leekahung: Sounds good, thanks!

src/pages/Contacts.jsx
JaeGif: > Think we technically could have the sorting trigger on Desktop as well from the `Select` field, but that might be duplicate functionality?

I'm not sure I understand what you mean. As far as I'm aware, there's only mobile wireframes on Figma. Do you mean that the Select input sorting should only be available in a mobile view?

leekahung: Yes, the `Select` sorting should only be available in Mobile.

JaeGif:
> Alright, think that touches most of the things that still needed to be addressed. But outside of those, the changes made with the new theme looks great!

Great, I think I've addressed most of the points in the last few commits. Please let me know if there's anything else or if something doesn't fit the requirements. Cheers!

leekahung (approved):
Looking great! Approving.

PR #624, opened 2024-04-30
Issue 582/add copy button function

### LINKED ISSUE #582: [Enhancement] - Refactor copy functionality as a reusable function

With our latest changes in Development, we now have a function that lets users copy long URLs associated with webIds and invite links. Given that it might be a function that could be used in other places outside of the Profile page, think it'll be a good idea to make it a reusable function.

### Rationale:

Makes copy functionality reusable across the app.

### Proposed Changes:

Store the clipboard and addNotification functions together as one function in perhaps utils and make it export from there.

andycwilliams:
## This PR:
Resolves #582 
 
**1.** Refactors copy to clipboard function as its own util
**2.** Adds it to AddContactModal as well, since it was missing

No visual change

**1.** The naming is arbitrary, both the util itself (`saveToClipboard`) and the file it's in (`copy`). There may be better alternatives
**2.** Doublecheck whether JSDocs for this util is fully accurate

### DIFF
--- a/src/components/Profile/ProfileComponent.jsx
+++ b/src/components/Profile/ProfileComponent.jsx
@@ -40,13 +40,11 @@ const ProfileComponent = ({ contactProfile, webId }) => {
   // Public Profile Data
   const [profileName, setProfileName] = useState(profileData?.profileInfo?.profileName);
   const [nickname, setNickname] = useState(profileData?.profileInfo?.nickname);
-
   const [edit, setEdit] = useState(false);
 
   const loadProfileData = async () => {
     const profileDataSolid = await fetchProfileInfo(session.info.webId);
     setProfileData(profileDataSolid);
-
     setProfileName(profileDataSolid.profileInfo?.profileName);
     setNickname(profileDataSolid.profileInfo?.nickname);
   };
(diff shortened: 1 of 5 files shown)


src/components/Profile/ProfileComponent.jsx
```diff
@@ -158,8 +159,7 @@ const ProfileComponent = ({ contactProfile, webId }) => {
                 aria-label="Copy Invite Link"
                 edge="end"
                 onClick={() => {
-                  navigator.clipboard.writeText(signupLink);
-                  addNotification('success', 'Invite link copied to clipboard');
+                  saveToClipboard(webId, 'Invite link copied to clipboard', addNotification);
```
leekahung: This should be the `signupLink` instead of `webId`

leekahung (commented):
Hey Andy! Just that one comment on the signup/invite link. Other than that, think it looks good 👍

src/components/Profile/ProfileComponent.jsx
```diff
@@ -158,8 +157,7 @@ const ProfileComponent = ({ contactProfile, webId }) => {
                 aria-label="Copy Invite Link"
                 edge="end"
                 onClick={() => {
-                  navigator.clipboard.writeText(signupLink);
-                  addNotification('success', 'Invite link copied to clipboard');
+                  saveToClipboard(signupLink, 'Invite link copied to clipboard', addNotification);
```
leekahung: Did a copy and paste of that new link and it doesn't seem to be matching the URL next to the button.

Looking that the Link above at Line 153, the link should really be `${signupLink}?webId=${encodeURIComponent(session.info.webId)}` I think. The `signupLink` might just be the initial portion of that link and doesn't actually return the full invite link.

leekahung (approved):
Alright, think that looks good. Was able to get to the right page using that new signup link on separate browser. 👍

PR #628, opened 2024-05-07
622 standardize all input fields

### LINKED ISSUE #622: [Enhancement] - Standardize all input fields in multiple modals

Currently, many of the margins between each input in the Add Contact, New Message, and Reply To modals are different sizes. They should be equidistant.

### Rationale:
Sizing should be more consistent and standardized. See the Upload Document modal for an example of the desired result:

### Proposed Changes:

Apply the same standardization of margin to the Add Contact and New Message modals.

DionSat:
## This PR:
Resolves #622 
 
**1.**   Standardize New Message Modal 
**2.**   Standardize Add Contact Modal
**3.**   Standardize Upload Document Modal

image

image

image

### DIFF
--- a/src/components/Modals/AddContactModal.jsx
+++ b/src/components/Modals/AddContactModal.jsx
@@ -164,6 +164,7 @@ const AddContactModal = ({ addContact, showAddContactModal, setShowAddContactMod
           />
           <Tooltip
             title="Select the server/website where your pod is located"
+            margin="normal"
             arrow
             placement="left"
           >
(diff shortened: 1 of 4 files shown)


andycwilliams:
Only just noticing this. When using different browsers I got different margins for the same modal.

Chromium:


Firefox:


Haven't looked too much into this, but it may have to do with that modal using `<br>` rather than setting the margins as "normal" the way other ones do. Add Contact uses `margin="normal"` rather than `<br>` and looks the same on both browsers.

DionSat:
Thanks for finding that. This is why I didn't want to finalize the PR yet. Because there are different modals I haven't checkout yet. I'll be sure to check that out and see if I can find any other issues once I'm back from work.

andycwilliams:
> Thanks for finding that. This is why I didn't want to finalize the PR yet. Because there are different modals I haven't checkout yet. I'll be sure to check that out and see if I can find any other issues once I'm back from work.

A long while back we were talking about not using `<br />`s and just use MUI's styling. But we never really settled on a plan so it's become a hodgepodge. This would be a good opportunity to standardize them.

DionSat:
I just decided to remove all the ```<br />``` tags from the modals. It seems to be running fine on my firefox and chrome. Hopefully thats everything I'll be moving this over to review.

leekahung:
I'm in favor of using styling/CSS to adjust for spacing over `<br />` since we'll have more control over margin and padding sizes.

andycwilliams (approved):
Nice work!

PR #630, opened 2024-05-10
627/resolve failing basic info tests

### LINKED ISSUE #627: [Bug Report] - Resolve failing Basic Info tests
### Describe the bug:
Two of the three tests for the Basic Info form, found in `BasicInfo.test.jsx`, are not passing. In order to retain complete test coverage, we want every test to pass.

### To Reproduce:
1. Un-comment out the tests
2. Run `npm run test` in the root directory

### Possible Fix:
Likely related to the data types being entered into the form. Strings are processing as intended, but `BasicInfo` has date as well as a select input. The test must account for these as well.

See `useCivicProfile.js` for more info about the data that the Basic Info form uses as a hook.

JaeGif:
## This PR:
Resolves #627
 
Fixes 2 broken tests.

# Test 1: Clear all returns everything to default state.
The functionality was correct, the test was failing because it was not selecting the correct element to view if the new state is rendered properly. I decided to check that /Decline to answer/ was in the DOM rather than burrowing looking for the value of 9. Changed DOB reset value to empty string.

# Test 2: submits a basic info profile update when you click the submit button

This one was harder to test. You can't type any specific date into this MUI component using RTL because it automatically opens the calendar view. What I decided to do was press enter to automatically select the current date, as the current date will always be default highlighted in the component. There's some variation in how Date.now() and the DatePicker element format the current date, so I converted Date.now() to the raw formatted yyyy-mm-dd date, and back into a days object to compare to the DatePicker's value.

### DIFF
--- a/test/components/CivicProfileForms/BasicInfo.test.jsx
+++ b/test/components/CivicProfileForms/BasicInfo.test.jsx
@@ -1,83 +1,114 @@
 import React from 'react';
-import { render } from '@testing-library/react';
+import { render, screen, within } from '@testing-library/react';
 import { vi, expect, it, describe } from 'vitest';
 import { BasicInfo } from '@components/CivicProfileForms';
-import { useCivicProfile } from '@hooks';
+import { useCivicProfile, useNotification } from '@hooks';
+import userEvent from '@testing-library/user-event';
+import '@testing-library/jest-dom';
+import dayjs from 'dayjs';
 
 vi.mock('@hooks', async () => {
   const actual = await vi.importActual('@hooks');
 
   return {
     ...actual,
-    useCivicProfile: vi.fn()
+    useCivicProfile: vi.fn(),
+    useNotification: vi.fn()
   };
 });
 
 describe('Basic info form', () => {
   it('renders', () => {
     useCivicProfile.mockReturnValue({ data: {}, isSuccess: true, refetch: vi.fn() });
+    useNotification.mockReturnValue({ addNotification: vi.fn() });
+
     const { getByRole } = render(<BasicInfo />);
     const legalFirstName = getByRole('textbox', { name: 'Legal first name' });
     expect(legalFirstName).not.toBeNull();
   });
 
   // TODO: Resolve test not passing
-  // it('clears all input fields when you click the clear button', async () => {
-  //   const mockClear = vi.fn();
-  //   const basicInfoProfile = {
-  //     legalFirstName: 'Jane',
-  //     legalLastName: 'Doe',
-  //     legalDOB: '1980-12-15',
-  //     legalGender: 1
-  //   };
-  //   useCivicProfile.mockReturnValue({
-  //     add: mockClear,
-  //     isSuccess: true,
-  //     storedDataset: {},
-  //     refetch: vi.fn()
-  //   });
-  //   const { getByRole } = render(<BasicInfo />);
-  //   const legalFirstName = getByRole('textbox', { name: 'Legal first name' });
-  //   const legalLastName = getByRole('textbox', { name: 'Legal last name' });
-  //   const legalDOB = getByRole('textbox', { name: 'Choose date' });
-  //   const legalGender = getByRole('combobox', { name: 'Gender' });
-  //   const clearButton = getByRole('button', { name: 'Clear button' });
 
-  //   await userEvent.type(legalFirstName, basicInfoProfile.legalFirstName);
-  //   await userEvent.type(legalLastName, basicInfoProfile.legalLastName);
-  //   await userEvent.type(legalDOB, basicInfoProfile.legalDOB);
-  //   await userEvent.type(legalGender, `${basicInfoProfile.legalGender}{enter}`);
+  /* Likely related to the data types being entered into the form. Strings are processing as intended, 
+  but BasicInfo has date as well as a select input. The test must account for these as well.
+  See useCivicProfile.js for more info about the data that the Basic Info form uses as a hook. 
+  */
 
-  //   await userEvent.click(clearButton);
+  it('clears all input fields when you click the clear button', async () => {
+    const user = userEvent.setup();
+    const mockClear = vi.fn();
+    const basicInfoProfile = {
+      legalFirstName: 'Jane',
+      legalLastName: 'Doe',
+      legalDOB: '1980-12-15',
+      legalGender: 1
+    };
+    useCivicProfile.mockReturnValue({
+      add: mockClear,
+      isSuccess: true,
+      storedDataset: {},
+      refetch: vi.fn()
+    });
+    useNotification.mockReturnValue({ addNotification: vi.fn() });
 
-  //   expect(legalFirstName.value).toBe('');
-  //   expect(legalLastName.value).toBe('');
-  //   expect(legalDOB.value).toBe(null);
-  //   expect(legalGender.value).toBe('');
-  // });
+    const { getByRole } = render(<BasicInfo />);
+    const legalFirstName = getByRole('textbox', { name: 'Legal first name' });
+    const legalLastName = getByRole('textbox', { name: 'Legal last name' });
+    const legalDOB = getByRole('textbox', { name: 'Choose date' });
+    const legalGender = getByRole('combobox', { name: 'Gender' });
+    const clearButton = getByRole('button', { name: 'Clear button' });
+
+    await user.type(legalFirstName, basicInfoProfile.legalFirstName);
+    await user.type(legalLastName, basicInfoProfile.legalLastName);
+    await user.type(legalDOB, basicInfoProfile.legalDOB);
+    await user.click(legalGender);
+    const listBox = within(getByRole('listbox'));
+    user.click(listBox.getByText('Female'));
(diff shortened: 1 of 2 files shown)


leekahung (approved):
The changes looks good. Not getting any errors when submitting new values for the form.

One thing I've noticed however is that once the user submits a date for DOB, they could no longer reset it to an empty string/null state. They have to select a different date. Was wondering if we would want to keep that given that the field was supposed to be optional.

leekahung:
A separate issue, not exactly related to this one, but I was wondering if we would want to make the clear and submit buttons visible only when the user wants to edit the information. This is similar to how we have the edit button in `Profile`.

JaeGif:
> The changes looks good. Not getting any errors when submitting new values for the form.
> 
> 
> 
> One thing I've noticed however is that once the user submits a date for DOB, they could no longer reset it to an empty string/null state. They have to select a different date. Was wondering if we would want to keep that given that the field was supposed to be optional.

Hmmm. I didn't notice this before. Will look into it after work.

JaeGif:
> A separate issue, not exactly related to this one, but I was wondering if we would want to make the clear and submit buttons visible only when the user wants to edit the information. This is similar to how we have the edit button in `Profile`.

I personally feel that having a clear button is more user friendly for initially filling out a form, while when editing a user would probably want to edit individual fields at a time. I don't see a UI design for it though so I'm willing to try either.

test/components/CivicProfileForms/BasicInfo.test.jsx
```diff
@@ -1,83 +1,113 @@
...
   it('renders', () => {
     useCivicProfile.mockReturnValue({ data: {}, isSuccess: true, refetch: vi.fn() });
+    useNotification.mockReturnValue({ addNotification: vi.fn() });
+
     const { getByRole } = render(<BasicInfo />);
     const legalFirstName = getByRole('textbox', { name: 'Legal first name' });
     expect(legalFirstName).not.toBeNull();
   });
 
   // TODO: Resolve test not passing
```
andycwilliams: Can remove this comment now

test/components/CivicProfileForms/BasicInfo.test.jsx
```diff
@@ -1,83 +1,113 @@
...
-  //   expect(legalDOB.value).toBe(null);
-  //   expect(legalGender.value).toBe('');
-  // });
+    expect(legalFirstName.value).toBe('');
+    expect(legalLastName.value).toBe('');
+    expect(legalDOB.value).toBe('');
+    expect(await screen.findByText(/Decline to answer/i)).toBeInTheDocument();
+  });
 
   // TODO: Resolve test not passing
```
andycwilliams: Can remove this one too

test/components/CivicProfileForms/BasicInfo.test.jsx
```diff
@@ -1,83 +1,113 @@
...
+import dayjs from 'dayjs';
 
 vi.mock('@hooks', async () => {
   const actual = await vi.importActual('@hooks');
 
   return {
     ...actual,
-    useCivicProfile: vi.fn()
+    useCivicProfile: vi.fn(),
+    useNotification: vi.fn()
```
andycwilliams, line 16: Good idea to test the notification as well.

test/components/CivicProfileForms/BasicInfo.test.jsx
```diff
@@ -1,83 +1,113 @@
...
-  //   await user.type(legalDOB, basicInfoProfile.legalDOB);
-  //   await user.type(legalGender, `${basicInfoProfile.legalGender}{enter}`);
-  //   await user.click(submitButton);
-  //   expect(mockAdd).toBeCalledWith(basicInfoProfile);
-  // });
+  it('submits a basic info profile update when you click the submit button', async () => {
+    const user = userEvent.setup();
+    const mockAdd = vi.fn();
+    const basicInfoProfile = {
+      firstName: 'Jane',
```
andycwilliams, line 73: The basicInfoProfile in this test is set up different than the one in the other. Is there a particular reason for this?

I'm not sure it really matters. Just curious. If the tests pass as they should then that's the important thing.

andycwilliams (approved):
Tests look good. Thanks for the fix!

As to @leekahung's point, I think the user should be able to fully clear the entire form and not have to select a different date. Personally, I would expect a clear button to treat all input the same way. It'd be a tad annoying to have to do an unexpected extra step. Mostly a quality of life kind of thing.

As for the other thing, it makes sense to make editing this work in the same way as the Profile (by the way, didn't we decide to rename that to "Account" and "Civic Profile" to simply "Profile"?). Plus it makes sense to prevent the user from accidentally erasing all of their info without some kind of warning (such as implementing the confirmation modal we have).

But I hesitate to change overly much because we're so close to MVP. The Housing and Financial forms also still need to be fully finished.

I don't think UX has weighed in on the Civic Profile yet. Hopefully they'll be there next Wednesday and we can get their feedback.

test/components/CivicProfileForms/BasicInfo.test.jsx
JaeGif: In the first test, the value of the DOB doesn't matter, it just needs to be _something_ so I didn't change it. The 2nd test needs the DOB value (after conversion) to match the value stored by the DatePicker so I had to do a little manipulation and specifically selected the current date. I think I'll change the first test to also select current date so it's not too confusing in the future.

Edit: Changed the first test to also use the current date, and the same input method (pressing enter to select the date)

JaeGif:
> The changes looks good. Not getting any errors when submitting new values for the form.
> 
> One thing I've noticed however is that once the user submits a date for DOB, they could no longer reset it to an empty string/null state. They have to select a different date. Was wondering if we would want to keep that given that the field was supposed to be optional.

@leekahung , I'm trying to replicate this but on my end everything appears to be fully cleared even after submitting previous information. Can you share the steps you took to replicate the behavior?

leekahung:
> @leekahung , I'm trying to replicate this but on my end everything appears to be fully cleared even after submitting previous information. Can you share the steps you took to replicate the behavior?

Hey @JaeGif, no problem.

For the steps, I simply went to the Basic Info form and first submitted some initial information.

From there I tried submitting the form after clearing it. While clearing the form itself with the button is fine, however, when I want to submit the new information, it was prevented from doing so even though the message says my form was updated with the cleared fields.

When I refresh the page or go to a different page and back, the value for DOB has not changed nor has my Legal Name (though I think it's more related DOB using an invalid input, see clip below). Only when I switched the DOB values to a new date did the values finally change.

From what I see from the errors, it seems to be related to the handler functions for input changes and form submission.

https://github.com/codeforpdx/PASS/assets/14917816/b3afd0fe-ca3f-45a9-91cc-406ee3570be3

Here's what it looks like from the storage standpoint:

https://github.com/codeforpdx/PASS/assets/14917816/33abe03a-c78d-494b-86df-6be0b91b0e31

JaeGif:
> > @leekahung , I'm trying to replicate this but on my end everything appears to be fully cleared even after submitting previous information. Can you share the steps you took to replicate the behavior?
> 
> Hey @JaeGif, no problem.
> 
> For the steps, I simply went to the Basic Info form and first submitted some initial information.
> 
> From there I tried submitting the form after clearing it. While clearing the form itself with the button is fine, however, when I want to submit the new information, it was prevented from doing so even though the message says my form was updated with the cleared fields.
> 
> When I refresh the page or go to a different page and back, the value for DOB has not changed nor has my Legal Name (though I think it's more related DOB using an invalid input, see clip below). Only when I switched the DOB values to a new date did the values finally change.
> 
> From what I see from the errors, it seems to be related to the handler functions for input changes and form submission.
> Screen.Recording.2024-05-11.at.12.24.06.PM.mov
> 
> Here's what it looks like from the storage standpoint:
> Screen.Recording.2024-05-11.at.12.33.45.PM.mov

I believe I've fixed the bug. It occurs because I was using '' as the default value of DatePicker. I changed the default value to null, however null is coerced by the DatePicker element to an empty string representation in the DatePicker's value, yet passed on as null to handler functions. The test therefore needs to check than the DatePicker's value is '' after clearing the form. If you check for null it will fail because the internal representation is ''.

JaeGif:
> For the steps, I simply went to the Basic Info form and first submitted some initial information.
> 
> From there I tried submitting the form after clearing it. While clearing the form itself with the button is fine, however, when I want to submit the new information, it was prevented from doing so even though the message says my form was updated with the cleared fields.
> 
> When I refresh the page or go to a different page and back, the value for DOB has not changed nor has my Legal Name (though I think it's more related DOB using an invalid input, see clip below). Only when I switched the DOB values to a new date did the values finally change.
> 
> From what I see from the errors, it seems to be related to the handler functions for input changes and form submission.
> 

@leekahung 
I believe I've fixed the bug. It occurs because I was using '' as the default value of DatePicker. I changed the default value to null, however null is coerced by the DatePicker element to an empty string representation in the DatePicker's value, yet passed on as null to handler functions. The test therefore needs to check that the DatePicker's value is '' after clearing the form. If you check for null it will fail because the internal representation is ''.

Issue #587, opened 2024-02-29
[Enhancement] - Restyle Sign-in component


fighawkins:
### Describe the Current Behavior/Feature:
I noticed the sign in component on the /signup page is quite UGLY. just kidding, but I think it could looks a bit better.
<img width="746" alt="Screenshot 2024-02-28 at 7 36 10 PM" src="https://github.com/codeforpdx/PASS/assets/143238486/963bb26f-3c58-4b67-8a8e-e2066e15d961">

### Proposed Changes:
Detail the changes you propose to make to the existing feature. Include:
Perhaps remove one card, make it all on one. make elements balanced. make card have even margins and padding on top and bottom. just make it look pretty.

### Rationale:
How does it improve the current functionality? Discuss any limitations or problems with the existing feature that your suggestion aims to resolve.
It currently comes across as a bit amateur, so it effects user experience by seeming untrustworthy.

### Possible Drawbacks:
Discuss any potential drawbacks or negative impacts this enhancement might have, including:
no drawbacks.

### Alternatives Considered:
Mention any alternative solutions or enhancements you've considered. Why were they not chosen or why are they less desirable than your proposal?

### Additional Context:
Add any other context or screenshots about the enhancement request here.

timbot1789:
The purpose of the current design is to later support an additional card that allows people to sign up with an existing pod[^1]. In that design, the Register for PASS paper would have 2 cards: Create a new pod OR sign up with an existing pod. There's a stub issue for [this here](https://github.com/codeforpdx/PASS/issues/586).

Any restyling of this page should be able to be extended in such a way.

[^1]: The first version of this page actually had both flows, but the second "signup with an existing pod" was really buggy, so we removed it and haven't put it back yet.

AJSterner:
Updated with a tab based design allowing the user to select between the two sign in flows, however I'm having some trouble with the navbar overflowing and moving things around when we switch to the "existing pod" flow.

Any ideas of how to handle this? It significantly shifts the sign in paper...

![new_pod](https://github.com/codeforpdx/PASS/assets/16945988/1a6545c9-11e4-4359-8a4e-3e31782de308)
![overflow](https://github.com/codeforpdx/PASS/assets/16945988/d82cfb57-7595-4813-88a3-d0905a737e1c)

AJSterner:
Mobile view! Still needs some cleanup.
[...]

PR #647, opened 2024-05-30
Fix links in pod registration card

### LINKED ISSUE #645: [Bug Report] - HTML incorrectly rendered in new pod page
### Describe the bug:
After creating a new pod, the user is redirected to a landing page where a few links are shown. Those links are incorrectly rendered so the <a></a> tags are visible to the user.

### Expected behavior:
The page should show clickable links rather than HTML as text on the screen.

### Actual Behavior:
See the screenshot.

### Screenshots:
Screenshot_20240521_184038

### To Reproduce:
1. Go to the main page
2. Click on sign up
3. Enter new credentials
4. See some HTML rendered as text

### Possible Fix:
Should be simple to reformat the HTML a bit.

AJSterner:
## This PR:
Resolves #645 

The pod registration page wasn't registering the pod links correctly! Now it does.

**Before**
Screenshot_20240521_184038
**After**
Screenshot_20240528_182502

### DIFF
--- a/src/components/Signup/ShowNewPod.jsx
+++ b/src/components/Signup/ShowNewPod.jsx
@@ -29,17 +29,29 @@ const ShowNewPod = ({ oidcIssuer, podUrl, webId }) => {
         <h1>You have logged in with your existing pod.</h1>
       )}
       <Typography>
-        {podUrl &&
-          `You can access your pod through your Pod Provider: <a href=${oidcIssuer}>${oidcIssuer}</a>
-        <br />
-        We set up this pod to operate with PASS: <a href=${podUrl}>${podUrl}</a>
-        <br />`}
-        Your Pod Provider has created this webId for you: <a href={webId}>{webId}</a>
-        <br />
+        {podUrl && (
+          <>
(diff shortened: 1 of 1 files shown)


leekahung (approved):
Don't want to be too picky on this since it looks like it's working, but was wondering if we should start replacing the existing `<br />` with `<span>`/`<p>` elements or include styling to the MUI `<Typography />` components so that we have more control over spacing/padding/margins in the future in case we want to update them.

Other than that, can approve this change.

andycwilliams:
> Don't want to be too picky on this since it looks like it's working, but was wondering if we should start replacing the existing `<br />` with `<span>`/`<p>` elements or include styling to the MUI `<Typography />` components so that we have more control over spacing/padding/margins in the future in case we want to update them.
> 
> Other than that, can approve this change.

That definitely should be done but I'm fine with leaving it for another PR as this one is pretty focused. And these might already be covered by #587. Otherwise we can create a new issue.

leekahung:
Think we can merge this for now then. 👍

AJSterner:
image
Updated to use MUI!

Issue #643, opened 2024-05-27
[Discovery] - Resize loading animations to fit spaces


andycwilliams:
Note: This issue is marked for discovery. It is open for discussion and not necessarily decided upon.

### Describe the Current Behavior/Feature:

Currently, when loading the Contacts and Profiles/Documents pages, the animation is much smaller than the final content size.

Profile:
<img width="798" alt="2024-05-26 (4)" src="https://github.com/codeforpdx/PASS/assets/83156697/cd3edb01-8c0a-48dd-b88f-26b5214e8bf6">
<img width="788" alt="2024-05-26 (5)" src="https://github.com/codeforpdx/PASS/assets/83156697/5d5cc723-5f5b-48da-b198-9a3692fe44f7">

Contacts:
<img width="801" alt="2024-05-26 (6)" src="https://github.com/codeforpdx/PASS/assets/83156697/77bb2eb2-0348-4b82-ad98-52c8038ae419">
<img width="791" alt="2024-05-26 (8)" src="https://github.com/codeforpdx/PASS/assets/83156697/f39d82af-26ec-458e-8e32-4d747b7c9d98">

Messages:
<img width="789" alt="2024-05-26 (2)" src="https://github.com/codeforpdx/PASS/assets/83156697/feb008fd-7c26-4542-813b-fb4487416d67">
<img width="804" alt="2024-05-26 (3)" src="https://github.com/codeforpdx/PASS/assets/83156697/11179d46-1255-4992-9928-10d7b4a04dda">

### Rationale:

The loading animations should more closely match the content that will eventually replace it. The difference in size between loading and loaded can be slightly strange. Using a "skeleton" design instead of a smaller linear box is a bit more professional in appearance and is what other apps use.

### Proposed Changes:

Using the [Material UI Skeleton component](https://mui.com/material-ui/react-skeleton/), refactor the loading animation to more closely match each page's size. For instance, the one for Documents would mimic a grid whereas the one for Profile would mimic both the profile card and grid beneath it.

Try refactoring using the existing `LoadingAnimation` component rather than creating a completely separate, new one.

Account for #637 if it ends up being added. That animation should remain linear and not a skeleton.

### Possible Drawbacks:

This applies primarily for the desktop version. The updated mobile versions have not yet been completed.

[...]

Issue #642, opened 2024-05-25
[Discovery] - Break down Upload Document into multiple steps


andycwilliams:
Note: This issue is marked for discovery. It is open for discussion and not necessarily decided upon.

### Describe the Current Behavior/Feature:

Currently, the Upload Document modal contains every input and interaction on a single screen. This is completely functional but potentially could be made smoother, at least for smaller screens.

<img width="197" alt="2024-05-24 (1)" src="https://github.com/codeforpdx/PASS/assets/83156697/0223545e-867b-4885-aa5a-63f8410266a1">

### Rationale:

Users may prefer a workflow where fewer things are presented all at once. It would create more sections, but they would be less crowded and perhaps overwhelming. This is particularly a concern for trauma-informed design. As well as accessibility, since a large amount of smaller-sized items could be burdensome for the vision-impaired, and multiple inputs so close to one another increase the possibility of misclicking.

### Proposed Changes:

Break down the existing modal into at least one additional step in order to more closely match the Figma design:

<img width="572" alt="UploadAndViewDocumentWorkflowFromFigma" src="https://github.com/codeforpdx/PASS/assets/83156697/db0b9d40-e03c-4ce4-a71b-50931db82d72">

### Possible Drawbacks:

Breaking up a working feature can be risky if not done carefully. Doing it may also prove unduly cumbersome.

### Alternatives Considered:

Consider if there are other ways to reduce screen "busyness". 

If we do decided to move forward with breaking down this modal, we may want to consider doing the same for other similar other modals.

PR #654, opened 2024-06-11
add loading animation to upload doc modal

### LINKED ISSUE #637: [Discovery] - Add loading animation to modals
Note: This issue is marked for discovery. It is open for discussion and not necessarily decided upon.

When the user uploads a file from the Upload Document modal, there is no indication about the progress. There is a notification upon completion or failure, but nothing until then.

### Rationale:

Adding this animation would provide real-time feedback to the user so they know if it's working, and whether the upload even began or not. For small files, this isn't a big problem. But for larger ones it would be good to give more information.

Example of how it currently appears once a file is selected and the Upload button pressed:
CurrentUploadDocumentProcess2

The only visible change is the Upload button becoming greyed out.

### Proposed Changes:

- Add a simple loading animation to modals that (namely the Upload Document modal).
- Try implementing `LoadingAnimation.jsx` rather than create a new, separate function or component.
- Style it based off of the proposed design from the Figma board:
AddUploadingAnimation

russfraze:
## This PR:
Resolves #(654) 

I've implemented this and you can see the loading animation for a split second when uploading a document. I uploaded a 12MB image and it still only takes a split second. I'm wondering if that is because I am using the local pod server? 

I think this loading animation makes for a better user experience if the file is large enough to take a couple of seconds, for sure. 

This needs further styling to match the Figma file. 

Currently, the loadingAnimation component delivers an indeterminant progress bar which is fine, but the Figma file shows a determinate progress bar with a percentage of the upload progress. Let me know if we intend to have a determinate progress bar here and I can look into how to make that work.


andycwilliams:
Just to make sure, is this still a draft or are you looking for a review already?

russfraze:
@andycwilliams this is still a draft, for sure. I should have just pinged you instead of requesting a review.

src/components/Modals/UploadDocumentModal.jsx
```diff
@@ -222,6 +223,7 @@ const UploadDocumentModal = ({ showModal, setShowModal }) => {
             </DialogActions>
           </FormControl>
         </form>
+        {processing && <LoadingAnimation loadingItem={file.name} />}
```
leekahung: Rather than it being condition within the FormSection component, I think it should be wrapping it as part of a ternary:

```
{ processing ? <LoadingAnimation /> : <FormSection>modal content here</FormSection> }
```

https://github.com/codeforpdx/PASS/assets/14917816/ce17ca2b-c7b3-4f9b-89bb-77294973f053

andycwilliams: So it would fully replace all of the content with the animation?

I prefer to stay with the Figma design. I think it looks a bit nicer and is less jarring than taking over the whole modal.

leekahung: The previous animation that's dangling at the bottom of the modal does not seem any better either. Might want to use a different kind of animation if not the full content

russfraze: Maybe we could look at a solution where the size of the modal does not change, which is jarring, and the loading animation displays over top of the modal with the content greyed out or removed. I don't mind the content being removed as long as the size of the modal doesn't change.

andycwilliams: > Maybe we could look at a solution where the size of the modal does not change, which is jarring, and the loading animation displays over top of the modal with the content greyed out or removed. I don't mind the content being removed as long as the size of the modal doesn't change.

We could try that. Or else the animation could temporarily replace one of the buttons (while retaining the size). Maybe "Upload File".

Thinking farther ahead, in #642 I mention possibly breaking down this whole process into several steps. That may or may not affect what we decide here.

russfraze: Hey @andycwilliams and @leekahung what do you think about something like this? 


Uploading Screen Recording 2024-06-28 at 3.59.39 PM.mov…

leekahung: The clip isn't showing. Think it might had not finished uploading

andycwilliams: Can't see your clip either but I just tested it locally. Not sure how it works for you, but this is what it's like for me:



At least for me, the animation stretches out the modal/drawer to the point where it's not all visible. My preferred approach is to temporarily replace one of the elements with the animation rather than add it anywhere.

Seeing as we already have a line that states "File to upload" under "Take Photos" maybe we could just have the animation replace that? It might be worthwhile make that text a tad larger anyway. So "File to upload: [filename]" would become "Loading: [filename]". Or just replace the text with the animation, which would probably make it easier to keep the modal from changing size.

---

Another option is to use the `useNotification` hook instead. Maybe create a separate state for loading (right now there are only two: add and delete). I haven't thought about this approach very much.

---

As an aside, I tested clicking "Cancel" before loading was complete. It just closes the modal while uploading continues in the background. Do we want to make that button also cancel the uploading process? If not, it may be good to disable that button until the uploading process is done. This would be a separate issue of course.

russfraze: Uploading Screen Recording 2024-06-28 at 3.59.39 PM.mov…

russfraze: @andycwilliams @leekahung  this is the clip I tried to post. I'm not sure why it wasn't working before. 

https://github.com/codeforpdx/PASS/assets/81979567/c4cda5ba-a8be-42a3-9e03-bd7c7669c738

andycwilliams: Ah, well that's pretty different! I'll take a closer look tomorrow.

leekahung: > @andycwilliams @leekahung  this is the clip I tried to post. I'm not sure why it wasn't working before. 
> 
> https://github.com/codeforpdx/PASS/assets/81979567/c4cda5ba-a8be-42a3-9e03-bd7c7669c738
> 
> 

I like this one 👍

andycwilliams: I don't know why, but it's still displaying the way it does in my screenshot. It's not appearing as an overlay in either desktop or mobile modes even after pulling the latest version of the branch.

russfraze: @andycwilliams Apologies - I did not push the changes I made to the code until just now.

src/components/Modals/UploadDocumentModal.jsx
```diff
@@ -222,7 +224,12 @@ const UploadDocumentModal = ({ showModal, setShowModal }) => {
             </DialogActions>
           </FormControl>
         </form>
+        {/* {processing && <LoadingAnimation loadingItem={file.name} />} */}
+        {/* <LoadingAnimation loadingItem={'testfilename'} /> */}
       </FormSection>
+      <Backdrop open={processing}>
+        <LoadingAnimation loadingItem="file name" />
```
andycwilliams: Shouldn't `loadingItem="file name"` still be `loadingItem={file.name}`?

src/components/Notification/LoadingAnimation.jsx
```diff
@@ -33,7 +33,15 @@ const LoadingAnimation = ({ loadingItem, children }) => (
...
-    <Paper elevation={2} sx={{ display: 'inline-block', mx: '2px', padding: '20px' }}>
+    <Paper
+      elevation={2}
+      sx={{
+        display: 'inline-block',
+        mx: '2px',
+        padding: '20px',
+        backgroundColor: 'background.tint'
+      }}
+    >
```
andycwilliams: While this works for this animation, it consequently affects other uses of this animation as well, such as when loading contacts or documents.



The box is now gray rather than white. Not a huge deal or anything, just an unintended side effect.

Best practice would be to carve out an exception for this. However, I'd like to do #643, which would refactor animations a tad anyway.

andycwilliams (commented):
Since this is still a draft I'm not doing a full on review, just adding a couple of thoughts.

Looking pretty nice though!

andycwilliams:
Any updates @russfraze?

andycwilliams:
Without any updates this PR might need to be closed or reassigned.

russfraze:
@andycwilliams I forgot about this, but back on it. I passed an isModal prop to the LoadIngAnimation component so that the background tint color is only if the LoadingAnimation is being used in a modal. The rest of the time it remains white.

andycwilliams:
Thanks for continuing work on this!

Looks like this branch has not been merged with the latest version of Development so it's behind a bit. Could you update it, please?

andycwilliams (commented):
Sorry for the delay.

Are you also getting this effect where "Enter Description" appears through the pop-up?



It's been a while since I last looked at this. Are we definitively decided to use a pop-up design rather than include the animation in the modal itself?

If we're to leave it as a pop-up, I think the width should not exceed the modal itself.

PR #655, opened 2024-06-11
Issue 648/contacts mobile view

### LINKED ISSUE #648: [Enhancement] - Add responsive mobile view for contacts table

#606 covers refactoring the Documents table. This covers the one for Contacts.

### Rationale:

Makes it responsive for small screens.

### Proposed Changes:

Refactor table to match Figma design:

andycwilliams:
## This PR:
Resolves #648 

Beginning this as a draft PR to get feedback, primarily about visuals and usability.
 
**1.** Switches from MUI Table to a new custom Card and Grid-based component on smaller viewports
**2.** Breaks up Contacts List Table into smaller parts (similar to Navbar)    
**3.** Adds the "copy web ID" function as an option

Smaller screen:

Smaller screen (open menu):

Smallest screen:

The styling is based on my own best judgment. I tried adhering to the Figma design while retaining as much information and action from the Table as possible.

Finish abstracting `ContactsListTable` into multiple, smaller components.

Write tests for each component.

Ensure clicking on the profile icon works as intended.

**1.** Feedback on styling. Is it intuitive and user-friendly?
**2.** Does everything perform as intended (e.g. adding or deleting a contact, clicking through to the contact's profile page)? Want to ensure it operates as expected under as many circumstances as possible.


russfraze:
Something I'm noticing is when you click on the dropdown for a contact, you can click the profile section and it seems like it should go to the profile but it doesn't unless the actual icon is clicked. 

Does it seem like clicking the section should work as well as the icon itself?


https://github.com/codeforpdx/PASS/assets/81979567/03dbaa81-2b4d-46a0-8d35-98960d5aa8fb

russfraze:
When checking out the smaller screens in the dev tools ( this is 375 X 667 ) it seems like the pod url overlaps the icon for the dropdown.  At larger sizes when there is space between the url and the icon, it reads that ACTIONS is the label for each of those icons, but I'm wondering if we should preserve that space to make it clear on mobile.
Screenshot 2024-06-25 at 1 06 37 PM

leekahung:
> When checking out the smaller screens in the dev tools ( this is 375 X 667 ) it seems like the pod url overlaps the icon for the dropdown.  At larger sizes when there is space between the url and the icon, it reads that ACTIONS is the label for each of those icons, but I'm wondering if we should preserve that space to make it clear on mobile.
> Screenshot 2024-06-25 at 1 06 37 PM
> 

Preserving the spacing would be good I think and let's users have a clear area to use for interacting with the actions button

Could make a size condition for it

src/components/Contacts/ContactListTable.jsx
```diff
@@ -42,110 +27,36 @@ const CustomToolbar = () => (
...
   const [showMessageModal, setShowMessageModal] = useState(false);
   const [messageToField, setMessageToField] = useState('');
+  const theme = useTheme();
+  const isSmallScreen = useMediaQuery(theme.breakpoints.down('sm'));
 
   const handleSendMessage = (contactId) => {
     setShowMessageModal(!showMessageModal);
-    setMessageToField(contactId.value.podUrl);
+    setMessageToField(isSmallScreen ? contactId.podUrl : contactId.value.podUrl);
   };
```
leekahung, line 36: I think the state to set would not need to change,  just the size of the input fields used needs to be corrected?

src/components/Contacts/ContactListTable.jsx
```diff
@@ -42,110 +27,36 @@ const CustomToolbar = () => (
...
-        <GridActionsCellItem
-          icon={<DeleteOutlineOutlinedIcon />}
-          onClick={() => deleteContact(contactData.row.Delete)}
-          label="Delete"
+      ) : (
+        <ContactListTableDesktop
+          contacts={contacts}
+          deleteContact={deleteContact}
+          handleSendMessage={handleSendMessage}
+          // handleProfileClick={handleProfileClick}
```
leekahung: Is this comment out line for profile click temporary while we transition these changes?

src/components/Contacts/ContactListTableMobile.jsx
```diff
@@ -0,0 +1,199 @@
...
+          <Card
+            sx={{
+              my: '5px',
+              position: 'relative'
+            }}
+          >
+            <CardContent>
+              <Box>
+                <Typography variant="body1" component="div" noWrap>
+                  {contact.givenName || ''} {contact.familyName || ''}
```
leekahung: I think adding a non-breaking space between the given and family name would let the line from collapsing if the user ends up not adding names to their profiles.

```
{contact.givenName || ''}&nbsp;{contact.familyName || ''}
```

Screenshot 2024-06-25 at 9 49 47 PM

src/components/Contacts/ContactListTableMobile.jsx
```diff
@@ -0,0 +1,199 @@
...
+                </Typography>
+                <Typography variant="body2" component="div" noWrap color="text.secondary">
```
leekahung: I think adding a width to this Typography component (Like 90%) might be good in solving what Russell has pointed out.

Screenshot 2024-06-25 at 9 53 19 PM

src/components/Contacts/ContactProfileIcon.jsx
```diff
@@ -38,7 +38,7 @@ const ContactProfileIcon = ({ contact }) => {
     try {
       await getWebIdDataset(contactInfo.webId);
       setContact(contactInfo);
-    } catch {
+    } catch (error) {
```
leekahung: Unless we're using something from the error object caught by the catch block like an error message or an error status, I don't think we need to call error. Just a catch would be enough.

leekahung:
See comments for feedback on draft 👍

andycwilliams:
> Something I'm noticing is when you click on the dropdown for a contact, you can click the profile section and it seems like it should go to the profile but it doesn't unless the actual icon is clicked.
> 
> Does it seem like clicking the section should work as well as the icon itself?
> 
>  Screen.Recording.2024-06-25.at.12.58.25.PM.mov

Yeah I note this in my draft post. It has to do with the way `ContactProfileIcon` is coded, which is different than the way the other actions.

src/components/Contacts/ContactListTable.jsx
andycwilliams: Could you elaborate please?

src/components/Contacts/ContactListTable.jsx
andycwilliams: That was just for testing. I already removed it in a change, just haven't pushed it up yet.

src/components/Contacts/ContactListTableMobile.jsx
andycwilliams: For me it's not very visually different (now it seems too large rather than too small) but I've added it anyway. 

Maybe we should have a default for when they don't add a name, like "No name provided" that is slightly greyed out. Not sure.

src/components/Contacts/ContactListTable.jsx
leekahung: Does contactId change when the screen size changes?

src/components/Contacts/ContactListTableMobile.jsx
leekahung: The size should be the same as if there were texts for the name, so it's fine

Adding no name provided is okay too

andycwilliams: For reference, here's what it looks like with `{contact.givenName || ''} {contact.familyName || ''}`:



And with `{contact.givenName || ''}&nbsp;{contact.familyName || ''}`:



Will now look into adding some kind of placeholder.

leekahung: A bit confused for the reference pic?

The gap with the non breaking space looks the same as the one I've screenshooted in an earlier comment

andycwilliams: Oh, I'm talking about the size of the card itself, not the space in between names.

src/components/Contacts/ContactListTable.jsx
andycwilliams: For whatever reason, this is what happens when using `contactId.value.podUrl` on mobile:



The "To" field does not populate and it renders this error.

leekahung: I think it supposed to just be `contactId.podUrl`. Use that as the main variable for the set function:

```
setMessageToField(contactId.podUrl);
```

andycwilliams: That has the opposite effect for me. It messes up sending a message on the desktop version

leekahung: I would check the state of the object being passed in and see why the changes occur when switching between mobile and desktop views.

leekahung: Just found the answer.

Screenshot 2024-06-26 at 9 29 12 PM

Since the object does change depending on what mode it's using, then this should be fine.

src/components/Contacts/ContactListTableMobile.jsx
```diff
@@ -0,0 +1,207 @@
...
+    setAnchorEl(null);
+    setOpenMenu(null);
+  };
+
+  const handleMenuItemClick = (action, contact) => () => {
+    action(contact);
+    handleClose();
+  };
+
+  const handleProfileClick = (contactData) => {
```
dbmzzo: if you change this function to the following, then clicking on the label will also load the profile:
```
  const handleProfileClick = (contact) => {
    navigate(`/contacts/${encodeURIComponent(contact.id)}`)
  };
```

andycwilliams:
Opening up for review now. 

Couple of things:
1. Needs confirmation that the mobile version of navigating to a contact's profile works correctly. I've tested but want to ensure it does for others as well.
2. The tests are broken. I'm creating another issue to fix those since there a few other components have failing tests anyway.
3. I added a "Copy WebId" action somewhat arbitrarily. Let me know if you think we even want or need it. Also, if we keep it, would we add it to the desktop table as well?
4. The sorting select for mobile still doesn't sort. I began writing a hook for it (`useTableSort`) but wasn't planning on finishing it with this PR specifically. But I can do that if we want.

src/components/Contacts/ContactListTableMobile.jsx
leekahung, line 1: While the navigation is correct, the profile displayed in contacts is not the correct one (see clip below).

https://github.com/codeforpdx/PASS/assets/14917816/f5949e7f-0b02-4150-a928-2413a1e6d8b2

The Desktop version still works as intended.

https://github.com/codeforpdx/PASS/assets/14917816/24f3ec1e-08df-4c55-861e-abe375270057

src/components/Contacts/ContactListTableMobile.jsx
```diff
@@ -0,0 +1,213 @@
...
+    setOpenMenu(null);
+  };
+
+  const handleMenuItemClick = (action, contact) => () => {
+    action(contact);
+    handleClose();
+  };
+
+  const handleProfileClick = (contact) => {
+    // TODO: Abstract this so ContactListTableDesktop uses the same function?
```
leekahung: While `navigate` helps re-route users to the right path, it's missing two things. The `contact` state and a handler function that helps processes `contact.value` (check out the component `ContactProfileIcon`).

Screenshot 2024-07-01 at 4 36 49 PM

leekahung (changes requested):
Think the new mobile view looks good. 👍

Just need to make sure the routing and state management is correct. See feedback.

leekahung:
For sorting, we could certainly leave it in a separate issue/PR.

src/components/Contacts/ContactListTableMobile.jsx
andycwilliams: What's the optimal way of implementing this? I'm trying not to mess with ContactProfileIcon if at all possible.

Also I'm still wondering about whether anyone wants to keep the Copy WebId option or not and, if so, whether it should then be added to the desktop version as well (though would it be worth its own column or perhaps something that happens when you simply click on the Web ID cell?).

dbmzzo: Adding the contact data to the state option of the `navigate` function appears to fix the issue of routing to the wrong contact:
```
navigate(`/contacts/${encodeURIComponent(contact.webId)}`, {state: {contact}});
```

leekahung: Yep, it's as @dbmzzo stated, there's an option to pass in the state in `navigate` with React Router (https://reactrouter.com/en/main/hooks/use-navigate#optionsstate).

In our case, we would pass in `{ state: { contact } }` after the path. As for the `onClick`, you would want to use the same handler function that ContactProfileIcon uses and have that run before `navigate`. Would need to check if `contact` give the correct state though since the Desktop version is passing `contact.value`, so it might be `{ state: { contact: contact.value } }`

andycwilliams: I was trying testing out `navigate(`/contacts/${encodeURIComponent(contact.webId)}`, {state: {contact}});` earlier but can for sure keep trying.

leekahung: Yeah try with `{ state: { contact: contact.value } }` instead. It might be looking for a state that's just part of contact and not the full contact object itself.

andycwilliams:
Alright, so I'm noticing some oddities but I think they've been there already.

This is from the deployed version, which is using the development branch:



"Samuel Foster" is the account I'm logged in with, and "Ava Townsend" is the profile I'm viewing. The breadcrumb and WebId are correct but name, nickname, and profile picture seem to populate with the logged-in account's data.

If I try a different profile, it does not do the same. Here name, nickname, and profile picture are not overwritten.



Since this seems to be an existing phenomenon, is anymore needed for this PR?

leekahung:
> Alright, so I'm noticing some oddities but I think they've been there already.
> 
> This is from the deployed version, which is using the development branch:
> 
> 
> 
> "Samuel Foster" is the account I'm logged in with, and "Ava Townsend" is the profile I'm viewing. The breadcrumb and WebId are correct but name, nickname, and profile picture seem to populate with the logged-in account's data.
> 
> If I try a different profile, it does not do the same. Here name, nickname, and profile picture are not overwritten.
> 
> 
> 
> Since this seems to be an existing phenomenon, is anymore needed for this PR?

No, something weird is going on. Not sure what's going on with your test accounts, but it is supposed to show only the contact info in contacts, not the user's own info. The object passed in is not correct.

https://github.com/codeforpdx/PASS/assets/14917816/a88d84b6-2f78-45a5-8b04-53ff92e0338e

src/components/Contacts/ContactListTableMobile.jsx
```diff
@@ -0,0 +1,214 @@
...
+
+  const handleMenuItemClick = (action, contact) => () => {
+    action(contact);
+    handleClose();
+  };
+
+  const handleProfileClick = (contact) => {
+    // TODO: Consider abstracting this into a reusable function
+    navigate(`/contacts/${encodeURIComponent(contact.webId)}`, {
+      state: { contact }
```
leekahung, line 85: This itself won't work. You're still missing the handler function which help sets up the correct profile.

leekahung: You need to use the `setContact` from `DocumentListContext` and create that new handler for the mobile view as well as importing `getWebIdDataset` to check if the webId is valid.

Screenshot 2024-07-03 at 12 54 34 PM

If you do that, we'll keep the same functionality as Desktop.

https://github.com/codeforpdx/PASS/assets/14917816/8b9c6fec-cb85-4a30-b1d5-f9887b781142

leekahung (changes requested):
See comment on Mobile view for Contacts

leekahung (approved):
Was able to get the mobile view working from the changes. Can approve this. 👍

Would honestly like the unit tests to be fixed soon though.

PR #653, opened 2024-06-07
587: Restyle Sign In

AJSterner:
## This PR:
Resolves #587, #660
 
**1.** Updates the styling of the sign-in page to be visually simple. Separates new and existing pod sign-up into tabs (#587)
**2.** Changes the show password buttons to be connected (clicking either one shows both the password and confirm password)
**3.** Changes the display property of `Layout.jsx` from grid to the default (#660)

## Screenshots (if applicable):
[More screenshots here](https://github.com/codeforpdx/PASS/pull/653#issuecomment-2212597616)
### Before
before_desktop
### After
after_new_desktop

PR #656, opened 2024-06-20
Issue 651/reroute civic profile

### LINKED ISSUE #634: [Discovery] - Refactor Civic Profile to use tabs
Note: This issue is marked for discovery. It is open for discussion and not necessarily decided upon.

Currently, the main page for the Civic Profile contains only a breadcrumb and links to the three forms.

### Rationale:

It seems like this is not the most efficient use of space. We can save the user a bit of time and clicking by consolidating the forms.

### Proposed Changes:

- Refactor the forms into Material UI Tabs and have Basic Info be the default when the user clicks on "Civic Profile"
- This could be retooled to work in the same way the Message page currently does (such as how, when clicked on, each tab will still change the breadcrumbs)

### Possible Drawbacks:

- Depending on how close to MVP we are, such a refactor could slow things down
- The Housing Information form is still underway whereas the Financial Information form has not quite yet started. It may not be fully clear what the Civic Profile will ultimately look like until they are done

### Alternatives Considered:

We could also potentially combine all the forms into a single page. Though this may depend on how many fields the forms will ultimately have combined.

### LINKED ISSUE #651: Reroute Civic Profile and disable Financial Info form

The first issue is that the Civic Profile landing page does not contain anything to distinguish it as its own page:

So there's little reason for the user to be directed to that page. 

Secondly, as opposed to the other forms in the Civic Profile, the functionality for Financial Info is not in place. As a result, it is empty:

### Rationale:

As we move towards MVP, we have to trim (if temporarily) any features that we don't expect to be fully ready in time.

Directing the user to the Civic Profile landing page right not just makes them perform an extra click to access any of the functionality (aka the two forms). It would be a bit more intuitive to just direct them to the Basic Information form itself.

As for the Financial Information form, the Basic and Housing forms cover all of the form functionality currently written. Thus Financial Information has no use at this time.

### Proposed Changes:

- [x] Reroute the "Civic Profile" link in the Navbar so that it lands on the Basic Info form rather than the empty parent page.
- [x] Disable (but do not remove) the link to the Financial Info form.
- [x] Comment out the unused code rather than delete it, since these may be implemented in the future.

### Additional Context:

These tasks may be simpler if #634 were to be done first but it is not necessary.

andycwilliams:
## This PR:
Resolves #651 and #634

Setting this as a draft for now. 
 
**1.** Disables linking to incomplete Financial Info form
**2.** Reroutes clicking "Civic Profile" in NavBar from main Civic Profile page to the Basic Info form
**3.** Refactors Civic Profile navigation from simple menu to Material UI Tabs

These are pretty janky, imperfect solutions. I'm sure I can optimize it but in the interest of time, I'm at least starting out with this.

Look into code optimization and/or abstraction.

Ensure it's responsive on all screens, browsers, and other user conditions.

### DIFF
--- a/src/components/CivicProfileForms/HousingInfo.jsx
+++ b/src/components/CivicProfileForms/HousingInfo.jsx
@@ -99,6 +99,7 @@ const HousingInfo = () => {
               label="Street:"
               onChange={handleChange}
               value={formData.lastPermanentStreet ?? ''}
+              autoComplete="street"
               fullWidth
               autoFocus
             />
@@ -110,6 +111,7 @@ const HousingInfo = () => {
               label="City:"
               onChange={handleChange}
               value={formData.lastPermanentCity ?? ''}
+              autoComplete="city"
               fullWidth
             />
           </Grid>
@@ -120,6 +122,7 @@ const HousingInfo = () => {
               label="State:"
               onChange={handleChange}
               value={formData.lastPermanentState ?? ''}
+              autoComplete="state"
               fullWidth
             />
           </Grid>
@@ -132,6 +135,7 @@ const HousingInfo = () => {
               value={formData.lastPermanentZIP ?? ''}
               error={zipError}
               helperText={zipError ? 'Invalid ZIP format. Expected: 12345 or 12345-6789' : ''}
+              autoComplete="postal-code"
               fullWidth
             />
           </Grid>
--- a/src/components/CivicProfileForms/HousingInfo.jsx
+++ b/src/components/CivicProfileForms/HousingInfo.jsx
@@ -22,7 +22,6 @@ import { FormSection } from '../Form';
  * @name HousingInfo
  * @returns {React.JSX.Element} The HousingInfo Component
  */
-
 const HousingInfo = () => {
   const { data, add, isSuccess, storedDataset, refetch } = useCivicProfile();
   const { addNotification } = useNotification();
@@ -92,14 +91,10 @@ const HousingInfo = () => {
   return (
     <FormSection title="Housing Information">
       <form aria-labelledby="hmis-basic-info-form" onSubmit={handleSubmit} autoComplete="off">
-        <Grid
-          container
-          spacing={2}
-          // columns={{ xs: 1, sm: 2 }}
-        >
+        <Grid container spacing={2} mt={0} mb={2}>
           <Grid item xs={12} md={6}>
             <TextField
-              id="street-input"
+              id="hmis-housing-info-last-street"
               name="lastPermanentStreet"
               label="Street:"
               onChange={handleChange}
@@ -110,7 +105,7 @@ const HousingInfo = () => {
           </Grid>
           <Grid item xs={12} md={6}>
             <TextField
-              id="city-input"
+              id="hmis-housing-info-last-city"
               name="lastPermanentCity"
               label="City:"
               onChange={handleChange}
@@ -120,7 +115,7 @@ const HousingInfo = () => {
           </Grid>
           <Grid item xs={12} md={6}>
             <TextField
-              id="state-input"
+              id="hmis-housing-info-last-state"
               name="lastPermanentState"
               label="State:"
               onChange={handleChange}
@@ -130,7 +125,7 @@ const HousingInfo = () => {
           </Grid>
           <Grid item xs={12} md={6}>
             <TextField
-              id="zip-input"
+              id="hmis-housing-info-last-zip"
               name="lastPermanentZIP"
               label="ZIP Code:"
               onChange={handleChange}
@@ -146,7 +141,7 @@ const HousingInfo = () => {
                 Months Houseless Past 3 Years:
               </InputLabel>
               <Select
-                id="months-homeless-input"
+                id="hmis-housing-info-months-homeless"
                 labelId="months-homeless-input-label"
                 name="monthsHomeless"
                 label="Months Houseless Past 3 Years:"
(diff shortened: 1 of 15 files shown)


leekahung (commented):
Looking at the visual changes, I'd much prefer the vertical tabs as oppose to horizontal ones since we already have horizontal tabs in the main navbar.

The second thing that might be an issue if certainly the content jumping which looks a bit clunky.

https://github.com/codeforpdx/PASS/assets/14917816/afc19882-18ac-48e6-b3f8-fb32ac60c0f0

Lastly, if we're cutting out Financial Information out for now and potentially adding it back in in the future, I would comment out the lines/tabs/components consisting of Financial Information like in the mobile view.

andycwilliams:
> Looking at the visual changes, I'd much prefer the vertical tabs as oppose to horizontal ones since we already have horizontal tabs in the main navbar.
> 
> The second thing that might be an issue if certainly the content jumping which looks a bit clunky.
> 
>  Screen.Recording.2024-06-26.at.12.45.06.AM.mov 
> Lastly, if we're cutting out Financial Information out for now and potentially adding it back in in the future, I would comment out the lines/tabs/components consisting of Financial Information like in the mobile view.

For mobile view, vertical looks nice but not for desktop, in my opinion. I can make it change based on screen size.

Do you mean completely commenting out all Financial Info code? Like everything `FinancialInfo.jsx` and its test?

leekahung:
> For mobile view, vertical looks nice but not for desktop, in my opinion. I can make it change based on screen size.
> 
> Do you mean completely commenting out all Financial Info code? Like everything `FinancialInfo.jsx` and its test?

Don't really like the Vertical tabs for mobile since there's only two and is not part of a drop down.

You mean horizontal tabs looks nicer in mobile than vertical since it takes up less space for the main content? I would prefer that. Other wise vertical tabs for desktop is good

And yes,  just comment it the unused components

dbmzzo:
The code looks good and works well for me. I tested on both Mac and iPhone, with a focus on narrower viewports. I find the tabs much better looking and functional than the menu.

The only issue I encountered was when I used autocomplete to fill in the form. Because of the presence of 'last' in the field names, autocomplete entered my last name in the ZIP and City fields. This can be easily fixed by adding the autoComplete prop to the inputs with the appropriate values. I can add inline comments for those potential changes.

src/components/CivicProfileForms/HousingInfo.jsx
```diff
@@ -90,121 +90,130 @@ const HousingInfo = () => {
...
               value={formData.lastPermanentStreet ?? ''}
+              fullWidth
+              autoFocus
             />
+          </Grid>
+          <Grid item xs={12} md={6}>
             <TextField
-              id="city-input"
+              id="hmis-housing-info-last-city"
               name="lastPermanentCity"
```
dbmzzo, line 110: adding `autoComplete="city"` helps avoid this auto-filling as last name.

src/components/CivicProfileForms/HousingInfo.jsx
```diff
@@ -90,121 +90,130 @@ const HousingInfo = () => {
...
               onChange={handleChange}
               value={formData.lastPermanentState ?? ''}
+              fullWidth
             />
+          </Grid>
+          <Grid item xs={12} md={6}>
             <TextField
-              id="zip-input"
+              id="hmis-housing-info-last-zip"
               name="lastPermanentZIP"
```
dbmzzo, line 132: adding `autoComplete="postal-code"` helps avoid this auto-filling as last name.

andycwilliams:
Alright, marking this as ready for review. The only lingering problem is the tests, but I plan on creating a new issue to address all of the failing ones anyway.

@leekahung It does look cleaner now. If and when the Financial form is ready, we may have to revisit tab arrangement but I don't think it's a limiting factor for now.

@dbmzzo Thanks for catching that. For some reason I can't quite replicate the issue but that's very likely because of the way I'm testing it. Also, since each field in the form has "last" in it I added an `autoComplete` value to each one rather than just those two.

dbmzzo (approved):
Looks and works great -- the rerouting works well, and the addition of the autoComplete props fixes the autocomplete errors I had in my browser.

leekahung (approved):
Yes, we'll probably want to re-visit the tab for Desktop.

https://github.com/codeforpdx/PASS/assets/14917816/88c6acf8-f1e7-4f04-a6e6-e5ffc77ec006

Don't really want Desktop users have to "chase" the tabs just to switch between views. I don't think it'll be a good user experience.

If we do want to keep it horizontal, we need to clamp down on the jumping. To do that, we'll want to mess around with the grid layout or make it so that the main content for civic profiles to be of fixed height.

I'll approve this for now.

test/pages/PersonalProfile.test.jsx
```diff
@@ -5,6 +5,8 @@ import { expect, it, describe } from 'vitest';
 import { CivicProfile } from '@pages';
 import { CIVIC_FORM_LIST } from '@components/CivicProfileForms';
 
+// TODO: Shouldn't this be testing Profile.jsx rather than CivicProfile.jsx?
```
leekahung, line 8: Also if there's a specific test file for CivicProfile, think we could drop the test here

andycwilliams:
@leekahung Since this seems to be caused by the grid in `Layout` (as noted in #653) is it really in this PR's scope?

Commented out CivicProfile tests and added a to-do note for a future PR to fix all tests.

If all looks good, I'll merge this soon.

leekahung:
> @leekahung Since this seems to be caused by the grid in `Layout` (as noted in #653) is it really in this PR's scope?
> 
> Commented out CivicProfile tests and added a to-do note for a future PR to fix all tests.
> 
> If all looks good, I'll merge this soon.

It's why I approved it

PR #657, opened 2024-06-26
[Bug fix] - Issue-650/bug fix mismatch pass pod creation

### LINKED ISSUE #650: [Bug Report] - Mismatching passwords not shown on pod creation
### Describe the bug:
The user is required to confirm their password when creating a new account, however the password in "Confirm Password" is not checked against the "Password" field. You can create a new account with mismatching passwords (Password, Confirm Password).

### Expected behavior:
I would expect client side validation to block submission until the passwords match.

### Actual Behavior:
You can successfully create an account with mismatching password (the primary?)

### Screenshots:
This would work:
image

### To Reproduce:
1. Go to /signup
2. Attempt to create an account with different passwords 
3. ...
4. Success?

### Possible Fix:
We should block submission and show an error, additionally password strength validation (length etc.) seems to be missing as well.

leekahung:
## This PR:

Resolves #650 by introducing validation and a disabled button until the error is corrected.

https://github.com/codeforpdx/PASS/assets/14917816/92d4d565-4136-4512-90b9-164a7867c7f3

### DIFF
--- a/src/components/Signup/PodRegistrationForm.jsx
+++ b/src/components/Signup/PodRegistrationForm.jsx
@@ -12,6 +12,8 @@ import InputLabel from '@mui/material/InputLabel';
 import TextField from '@mui/material/TextField';
 import Typography from '@mui/material/Typography';
 import { Visibility, VisibilityOff } from '@mui/icons-material';
+import { FormHelperText } from '@mui/material';
+import { useTheme } from '@mui/material/styles';
 
 /* Styles for MUI components */
 const cardStyle = {
@@ -47,6 +49,7 @@ const PodRegistrationForm = ({ register, caseManagerName, previousInfo = null })
   );
   const [showPassword, setShowPassword] = useState(false);
   const [showConfirmPassword, setShowConfirmPassword] = useState(false);
+  const theme = useTheme();
 
   const [searchParams] = useSearchParams();
   const handleSubmit = async (event) => {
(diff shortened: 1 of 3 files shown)


andycwilliams:
Looks good, only suggestion is to mark Password and Confirm Password as required, and to disable the button until everything is properly filled. Either that, or have text appear while the passwords don't match, like `      helperText="Passwords must match"`

leekahung:
> Looks good, only suggestion is to mark Password and Confirm Password as required, and to disable the button until everything is properly filled. Either that, or have text appear while the passwords don't match, like `      helperText="Passwords must match"`

I'll add it

leekahung:
Since it's a FilledInput, there are no attributes for helperText. Instead, I've used FormHelperText to help add the helper texts for the password mismatch

https://github.com/codeforpdx/PASS/assets/14917816/552cd60f-2cd2-400d-8b24-14ba312ebce8

The alignment issues and styling will be resolved once PR #653 is reviewed and merged, so that's been left alone here. For this PR, it's mainly focused on fixing the bug.

andycwilliams:
The helper does, well, help.

Come to think of it, I'm not sure why it uses `FilledInput` here. This is the only place we have it, otherwise we just use `TextField`.

And I'd still like to see the password fields marked as required and the "Set up your Pod" button disabled until all requirements are met. Same for "Login to Pod Provider".

leekahung:
> The helper does, well, help.
> 
> Come to think of it, I'm not sure why it uses `FilledInput` here. This is the only place we have it, otherwise we just use `TextField`.
> 
> And I'd still like to see the password fields marked as required and the "Set up your Pod" button disabled until all requirements are met. Same for "Login to Pod Provider".

We could open for a separate PR for those. This is mainly for the bug fix.

leekahung:
Would like to merge this in so that we could get the changes patched up before PR #653 (failing tests appears to be related to a previous merge, PR #656, as opposed to changes made in this PR)

andycwilliams (approved):

Issue #662, opened 2024-07-17
[Bug Report] - Fix Failing Tests


andycwilliams:
Work on this can begin right away. However, it should not be submitted until ensuring there are no additional failing tests from #571 and #606.

Component tests to fix:

- [x] BasicInfo.test.jsx - Andrew
- [x] ContactListTable.test.jsx
- [x] ContactListTableDesktop.test.jsx - Andy
- [x] ContactListTableMobile.test.jsx - Andy
- [x] DocumentCard.jsx - Josh
- [x] DocumentsDesktop.jsx - Josh
- [x] DocumentsMobile.jsx - Josh
- [x] HousingInfo.test.jsx - Andrew
- [x] Messages.test.jsx - Andy
- [x] PodRegistrationForm.test.jsx
- [x] Profile.test.jsx - Andrew
- [x] Signup.test.jsx

### Describe the bug:

Some of the tests are now failing.

### Expected behavior:

Every test should pass to ensure maximum test coverage possible.

### To Reproduce:

1. Run `npm run test` in the terminal

### Additional context:

The tests fail due to changes in the related components. So they should be updated or, when relevant, rewritten according to each component's new specifications. For example, some have been broken up into multiple smaller components and so will need additional tests.

`CivicProfile` and `FinancialInfo` are not yet implemented and therefore do not require tests at this time

andycwilliams:
I've written a few of the tests now but would appreciate some assistance so we can get this done sooner rather than later. I'm not the quickest at these.

AJSterner:
I'm fairly sure a bunch of these are timeouts which can be fixed by bumping up the global test timeout time or on a per-test basis.

AJSterner:
@andycwilliams I may have accidentally fixed them all on #666, take a look.

AJSterner:
@andycwilliams I don't see any obvious problems with BasicInfo or HousingInfo so I checked them off. Please let me know if there are others you'd like picked up (Profile?).

AJSterner:
Added my name next to Profile.

andycwilliams:
Several more tests have been written as of #667. Only a handful more to go.

andycwilliams:
@AJSterner Any news on tests for Profile? I'm aiming to get MVP wrapped up soon as possible.

AJSterner:
@andycwilliams Profile tests ready for an initial review:
https://github.com/codeforpdx/PASS/pull/673

AJSterner:
Closed out Profile tests!

PR #666, opened 2024-07-31
vitest: fix several tests

AJSterner:
This PR for #662:
- Updates the password queries for sign in page to select correct elements
- updates PersonalProfile tests to correctly select navigation and tab elements
- Makes the tests in CivicProfile and PersonalProfile `todo` instead of commenting out entire file

Resolves the following tests:
- test/components/CivicProfileForms/FinancialInfo.test.jsx
- test/components/Signup/PodRegistrationForm.test.jsx
- test/pages/CivicProfile.test.jsx
- test/pages/PersonalProfile.test.jsx
- test/pages/Signup.test.jsx
- test/components/Contacts/ContactListTableDesktop.test.jsx
- test/components/Contacts/ContactListTableMobile.test.jsx

I believe this fixes all the tests.


AJSterner:
😈😈😈😈😈😈😈😈😈😈😈😈😈😈😈

leekahung (approved):
Tests for PodRegistrationForm and Signup looks good (they passed the tests as intended with npm run test), can approve this for merging.

Although, now that I look at the two unit tests more, it feels like CivicProfile.test.jsx and PersonalProfile.test.jsx are supposed to be testing the same file (seeing that they both are rendering the CivicProfile component), but somehow the names diverged over time. Think we can kill one of the two test files.

The `.todo()`s are also a good idea until we get hard fixes on those tests. 👍

As for the unit test on Signup.test.jsx, think you and Andy could sort out which unit test update to keep from your two branches.

andycwilliams:
We don't have a component called `PersonalProfile` so `PersonalProfile.test.jsx` can probably be renamed `Profile.test.jsx`

I can look closer tomorrow. I haven't compared the our different tests for `Signup` yet

andycwilliams (commented):
I agree that `PersonalProfile.test` isn't really doing a whole lot. It seems like that should be testing the `Profile` page but it doesn't. 

There should be tests written for that instead of this. But I'm fine if it's not done with this specific PR, since we're already going to have multiple ones to fully finish the issue.

If not going to convert it into `Profile.test.jsx` and then make tests for it, I think `PersonalProfile.test` could be removed. Then we can create those tests later.

As for the `Signup` tests, we can just use yours. Mine are practically identical.

leekahung:
The unit test for PersonalProfile.test.jsx does not test the Profile component, but the CivicProfile component (see what the test is trying to render, there no imports for the Profile component).

We should merge the tests in there as part of CivicProfile.test.jsx.

andycwilliams:
@AJSterner Any updates?

I'm working on my own branch and would like to merge

AJSterner:
@andycwilliams @leekahung consolidated the CivicProfile tests that were in the `PersonalProfile.test.jsx` profile into `CivicProfile.test.jsx`.

https://github.com/codeforpdx/PASS/pull/666/commits/e119e7d2e9acf5ce1ccbb783159f48d57dadf389

andycwilliams (approved):

PR #673, opened 2024-08-28
662/asterner profile tests

AJSterner:
## This PR:
Resolves (part of) #662  
 
**1.** Adds basic render tests for the `Profile` page
**2.** Adds tests to check basic functionality for the `Profile` page 
**3.** Factors out a `MockSignedInUserContext` to allow multiple tests to access user profiles

## Future Steps:
We should figure out why there's a second timeout before loading the full profile page...
https://github.com/codeforpdx/PASS/blob/52c881fc278dfe0ebdc2c3c107aa4625db9c2793/src/pages/Profile.jsx#L114-L128

### DIFF
--- a/src/pages/Profile.jsx
+++ b/src/pages/Profile.jsx
@@ -21,7 +21,7 @@ import { LoadingAnimation } from '@components/Notification';
 // Util Imports
 import { truncateText } from '@utils';
 // Model Helpers
-import { fetchProfileInfo } from '@model-helpers/Profile';
+import { fetchProfileInfo } from '../model-helpers';
 
 /**
  * Profile Page - Page that displays the user's profile card information and
@@ -80,7 +80,7 @@ const Profile = () => {
   const contact = location.state?.contact;
   const [contactProfile, setContactProfile] = useState(null);
   const webIdUrl = contact?.webId ?? session.info.webId;
-  const [loadingProfile, setLoadingProfile] = useState(false);
+  const [loadingProfile, setLoadingProfile] = useState(true);
 
   // Handler for the SetAclPermissions Modal that
   // sets the appropriate version of the modal to load,
--- a/src/pages/Profile.jsx
+++ b/src/pages/Profile.jsx
@@ -21,7 +21,7 @@ import { LoadingAnimation } from '@components/Notification';
 // Util Imports
 import { truncateText } from '@utils';
 // Model Helpers
-import { fetchProfileInfo } from '@model-helpers';
+import { fetchProfileInfo } from '@model-helpers/Profile';
 
 /**
  * Profile Page - Page that displays the user's profile card information and
@@ -80,7 +80,7 @@ const Profile = () => {
   const contact = location.state?.contact;
   const [contactProfile, setContactProfile] = useState(null);
   const webIdUrl = contact?.webId ?? session.info.webId;
-  const [loadingProfile, setLoadingProfile] = useState(true);
+  const [loadingProfile, setLoadingProfile] = useState(false);
 
   // Handler for the SetAclPermissions Modal that
   // sets the appropriate version of the modal to load,
--- a/src/pages/Profile.jsx
+++ b/src/pages/Profile.jsx
@@ -21,7 +21,7 @@ import { LoadingAnimation } from '@components/Notification';
 // Util Imports
 import { truncateText } from '@utils';
 // Model Helpers
-import { fetchProfileInfo } from '../model-helpers';
+import { fetchProfileInfo } from '@model-helpers';
 
 /**
  * Profile Page - Page that displays the user's profile card information and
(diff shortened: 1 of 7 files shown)


src/pages/Profile.jsx
```diff
@@ -145,7 +145,9 @@ const Profile = () => {
           width: isSmallScreen ? '100%' : 'auto'
         }}
       >
-        <Typography sx={{ fontWeight: 'bold', fontSize: '18px' }}>My Profile</Typography>
+        <Typography variant="h1" sx={{ fontWeight: 'bold', fontSize: '18px' }}>
```
AJSterner, line 148: This change makes this text into a proper heading without modifying the visual rendering

andycwilliams: Looking over it, I'm not certain whether we need it to say "My Profile" at the top at all. The breadcrumb basically covers the same role already and we don't really apply this same concept to other pages (such as the Contacts page). But that's a different discussion

test/pages/Profile.test.jsx
```diff
@@ -0,0 +1,81 @@
+import React from 'react';
+import { BrowserRouter } from 'react-router-dom';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+
+import { act } from 'react-dom/test-utils';
+import { render, screen, cleanup, waitForElementToBeRemoved } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { describe, expect, it, beforeEach, afterEach } from 'vitest';
+
```
andycwilliams: Can delete these extra lines

test/mocks/contexts/MockSignedInUserContext.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { vi } from 'vitest';
+import { SignedInUserContext } from '@contexts';
+import * as profileHelper from '../../../src/model-helpers/Profile';
+
```
andycwilliams, line 5: If we're going to keep this, we can leave a TODO note here to import it elsewhere. I'm pretty sure there are other tests that can make use of this but not as part of this PR

andycwilliams (commented):
A couple of minor edits I'd like but it's looking good to me

test/mocks/contexts/MockSignedInUserContext.jsx
AJSterner: What would you like to add as a TODO?

I don't see the SignedInUserContext used in other tests (other than the tests for the context itself).

src/pages/Profile.jsx
AJSterner: Agreed. I attempted to make minimal changes to the page itself. This seems like something we can take a look at post MVP.

test/pages/Profile.test.jsx
AJSterner: Done.

test/mocks/contexts/MockSignedInUserContext.jsx
andycwilliams: I'll put a pin in that idea. I want to think it over more

test/components/Profile/ProfileComponent.test.jsx
```diff
@@ -4,37 +4,17 @@ import { cleanup, render, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 import { ProfileComponent } from '@components/Profile';
-import { SignedInUserContext } from '@contexts';
-import * as profileHelper from '../../../src/model-helpers/Profile';
 import '@testing-library/jest-dom/extend-expect';
 import createMatchMedia from '../../helpers/createMatchMedia';
```
andycwilliams: Seems like there's one more extra line that can be removed

andycwilliams (approved):
There's one extra line that can be removed but approving anyway.

Also would like @leekahung to give this a once-over just to be extra safe

test/components/Profile/ProfileComponent.test.jsx
AJSterner: Thanks! Removed

AJSterner:
> There's one extra line that can be removed but approving anyway.
> 
> Also would like @leekahung to give this a once-over just to be extra safe

Sweet! I'll wait for feedback from @leekahung

test/mocks/contexts/MockSignedInUserContext.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
```
leekahung, line 1: I see we're making mock contexts for the tests. I'm fine with this since it'll be useful in case multiple tests need the same context.

leekahung (approved):
Just a comment on the mock context, but it's a positive one. I can approve this. 👍

AJSterner:
Awesome! Thanks y'all :) merged!

PR #675, opened 2024-09-19
Hotfix contacts and messages

leekahung:
## This PR:

Does a hotfix to UI elements in Contacts and Messages. The `undefined` in New Messages modal is removed when name is not given. Only the pod URL will appear.

For the contacts list, I've included a `useEffect` from React and `refetch` function from React Query to grab a fresh list of contacts each time a user exits and comes back to the Contacts page, editing, or deleting contacts from the contacts list. The aforementioned interaction tends to make the app give a cryptic error messages related to null values when debugging.

The webId parser for adding new contacts does seem to be working in the development version when I logged into my solidcommunity.net account, so I'm not sure what's happening between what we have now and what's deployed on Vercel. That part remains untouched.

https://github.com/user-attachments/assets/3a377764-fe8e-4a60-a6e5-b9d0feb26980

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "pass",
-  "version": "1.0.0-alpha",
+  "version": "1.0.1-alpha",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "pass",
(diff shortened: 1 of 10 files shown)


andycwilliams (approved):
lgtm

PR #680, opened 2024-09-21
619/refactor image upload button

### LINKED ISSUE #619: [Enhancement] - Refactor image upload button onto profile picture
Refactor the CHOOSE IMG button to appear only when hovering over the profile picture itself.

### Rationale:
The profile is becoming quite visually crowded, with multiple input options all over the card. By limiting this input's visibility we can make it less "noisy".

ImageUploadRefactor

### Proposed Changes:
Refactor the upload function onto the profile picture, while adding CHOOSE IMG and a grayed-out background when hovering. It could look something like this:

Discord-Friend-03 width-500

andycwilliams:
## This PR:
Resolves #619
 
**1.** Removes "Choose/Remove Img" button on the Profile page and refactors it as a hover effect directly over the avatar
**2.** This approach more closely matches the way many services program the same functionality, so it ought to be intuitive enough for users

No photo uploaded:

No photo uploaded (plus hover):

With photo uploaded:

With photo uploaded (plus hover):

The goal is to make the Profile card less crowded-looking without sacrificing ease of use.

- Ensure this is accessible and intuitive.
- Can tweak it by adding text back over, if we decide that's necessary (rather than use the images alone).

### DIFF
--- a/src/components/Profile/ProfileImageField.jsx
+++ b/src/components/Profile/ProfileImageField.jsx
@@ -8,7 +8,6 @@ import Box from '@mui/material/Box';
 import Button from '@mui/material/Button';
 import HideImageIcon from '@mui/icons-material/HideImage';
 import ImageIcon from '@mui/icons-material/Image';
-import { useTheme } from '@mui/material/styles';
 // Contexts Imports
 import { SignedInUserContext } from '@contexts';
 // Component Imports
@@ -36,7 +35,6 @@ const ProfileImageField = ({ loadProfileData, contactProfile }) => {
   const [processing, setProcessing] = useState(false);
   const [showConfirmationModal, setShowConfirmationModal] = useState(false);
   const [hover, setHover] = useState(false);
-  const theme = useTheme();
 
   const handleProfileImage = async (event) => {
     if (event.target.files[0].size > 1 * 1000 * 1024) {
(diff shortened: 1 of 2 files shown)


src/components/Profile/ProfileImageField.jsx
leekahung, line 1: Small nitpick, could we have a small text right under the image icon (perhaps in white text) to let users know what it's for? For someone that's not tech-savvy, it's not immediately clear that I would need to click don't the image to change the image.

leekahung:
Think it's mostly just the test and linter at the moment, other than that, this looks good.

leekahung (changes requested):
The UI looks good, but I can't seem to upload any images as my profile image.

https://github.com/user-attachments/assets/a69f08e4-c68a-48a7-9d84-232c186759e9

andycwilliams:
> The UI looks good, but I can't seem to upload any images as my profile image.
> 
>  Screen.Recording.2024-10-09.at.2.11.26.PM.mov

Odd. It's working for me.

Jared-Krajewski:
> > The UI looks good, but I can't seem to upload any images as my profile image.
> > Screen.Recording.2024-10-09.at.2.11.26.PM.mov
> 
> Odd. It's working for me.

This works for me as well. I did a pair testing session with @leekahung and we were not able to figure out why it will not work. I think we should figure out why it works for some of us before merging in. The previous functionality in development works for all of us so we will have to dig in a little bit more and see what changed.

PR #681, opened 2024-09-25
663/add mobile contacts sort

### LINKED ISSUE #663: [Enhancement] - Implement search functionality for mobile contacts

A select option to sort the mobile version of the contacts table has been created. However, it does not yet work. 

However, the actual functionality has not yet been written. 

### Rationale:

Sorting is a basic feature and already exists for one version of the table. It makes no sense for it not to be available for the other. Besides, it's already been started.

### Proposed Changes:

Complete the function so that it properly sorts contacts by the existing values (First Name, Last Name, and Web ID), so that it operates as expected.

### Additional Context:

The desktop version works properly and does not need to be changed. This is because it uses an MUI component which includes sorting by default. But the mobile contacts table is written differently and so needs its own logic.

andycwilliams:
## This PR:
Resolves #663 
 
**1.** Adds functionality to the sorting menu in the mobile version of the Contacts page
**2.** Adds "Sort by:" as a placeholder, which is not a selectable sort option, as well as a "Default" option which returns the data to the original sort

This is a focused resolution for the issue which can still be improved upon in the future (see below). The reason for this is to prepare the app for demonstration.

Functionality is currently contained within Contacts. Could finish extrapolating it into custom hook (`useTableSort`) to use elsewhere, such as the Documents section. Tests are also rather preliminary.

### DIFF
--- a/src/components/Contacts/ContactListTable.jsx
+++ b/src/components/Contacts/ContactListTable.jsx
@@ -26,15 +26,15 @@ import ContactListTableMobile from './ContactListTableMobile';
  * @param {Function} Props.addContact - from Contacts page
  * @returns {React.JSX.Element} The ContactListTable Component
  */
-const ContactListTable = ({ contacts, deleteContact, handleDeleteContact, addContact }) => {
+const ContactListTable = ({ contacts = [], deleteContact, handleDeleteContact, addContact }) => {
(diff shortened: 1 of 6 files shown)


leekahung (approved):
Looks good

PR #683, opened 2024-10-10
[Enhancement] - Issue-682/create documents route

### LINKED ISSUE #682: [Enhancement] - Create My Documents View

Several changes can be made to make the design more user-friendly. For instance, accessing documents can be separated from the Profile page and into its own view, called My Documents

passdeisgn1001

### Rationale:

It is not the most intuitive for accessing documents to be inside of another page. When the user wants to see their documents, they would likely expect a more direct method than have to dig.

### Proposed Changes:

passdeisgn1001

Detail the changes you propose to make to the existing feature. Include:

- Create My Documents view, along with a route
- Move all relevant document-related actions/input over from Profile to the new view
- Additionally, rename Contacts to My Contacts to maintain convention

### Possible Drawbacks:

Should just be a matter of refactoring existing code. However, there might be a decent amount of it, so care is required.

leekahung:
## This PR:

Resolves #682 and migrates Documents list from Profile to new Documents page and re-purpose Documents list for contacts using state management.

https://github.com/user-attachments/assets/8cbb775d-0fcf-4769-9d53-054edf4d3e30


andycwilliams (changes requested):
Seeing some oddities with Shared Profile with resizing and overlapping elements.

Small screen (this happens when I resize the screen when on a shared profile. But doesn't seem to if I'm already on a mobile view on Contacts and then go to a shared profile):


Medium screen (placeholder documents table also cuts off at the bottom):


Large screen:


Also maybe the placeholder text for shared profiles should be something more like "No documents have been shared with you. Any that are shared will appear here". 

Other thoughts, likely for another PR. I think we should style the Documents and Contacts mobile versions more consistently. The Share/Add buttons are not spaced from the screen sides the same way. It also looks like the text in Documents is bolded.





Similarly, for the desktop versions, we should probably center the Add Contact button since we're centering the buttons for Documents now.

leekahung:
With the main problem with the issue resolved, think we can open up a new PR to handle all the styling corrections.

Jared-Krajewski:
> With the main problem with the issue resolved, think we can open up a new PR to handle all the styling corrections.

yeah, I think that sounds good. Is that good with you @andycwilliams ?

src/pages/Documents.jsx
```diff
@@ -0,0 +1,144 @@
...
+import { useTheme } from '@emotion/react';
+import { useMediaQuery } from '@mui/material';
+import { ConfirmationModal, SetAclPermissionsModal, UploadDocumentModal } from '@components/Modals';
+import { useNotification, useSession } from '@hooks';
+import { DocumentListContext } from '@contexts';
+import { truncateText } from '@utils';
+import { DocumentTable } from '@components/Documents';
+import { Container } from '@mui/system';
+
+const Documents = () => {
```
andycwilliams, line 20: No JSDocs?

src/pages/Documents.jsx
```diff
@@ -0,0 +1,144 @@
...
+import Button from '@mui/material/Button';
+import { useLocation } from 'react-router-dom';
+import { useTheme } from '@emotion/react';
+import { useMediaQuery } from '@mui/material';
+import { ConfirmationModal, SetAclPermissionsModal, UploadDocumentModal } from '@components/Modals';
+import { useNotification, useSession } from '@hooks';
+import { DocumentListContext } from '@contexts';
+import { truncateText } from '@utils';
+import { DocumentTable } from '@components/Documents';
+import { Container } from '@mui/system';
```
andycwilliams: Is there a reason Container is imported from '@mui/system' rather than '@mui/material' like it is elsewhere?

andycwilliams (changes requested):
375178745-a7b25bc1-12bc-4067-a461-36ca69dfaa2c

375178749-e4efa5f7-8bdf-4d67-a646-433f01f4ee81

This strange responsivity is still occurring. I'm not comfortable show this to a prospective investor. We want to put our best foot forward.

There's other tidying that could be done but I would really like this to be addressed.

src/pages/Documents.jsx
leekahung: No reason, it looks like the import path auto completed to the wrong path

I'll change it back

src/pages/Documents.jsx
leekahung: Given it's a page and not exactly a component, didn't think it'll be necessary to write one, but I can include a short documentation

leekahung:
> 375178745-a7b25bc1-12bc-4067-a461-36ca69dfaa2c
> 
> 375178749-e4efa5f7-8bdf-4d67-a646-433f01f4ee81
> 
> This strange responsivity is still occurring. I'm not comfortable show this to a prospective investor. We want to put our best foot forward.
> 
> There's other tidying that could be done but I would really like this to be addressed.

Which is why I recommend fixing this in a quick follow-up with this PR being the main push for component changes.

Currently working on those styling fixes in a local branch, as stated in an earlier comment about styling corrections. I've just pushed it up as a PR (see PR #684)

andycwilliams (approved):
Think it's good practice to address things like the responsivity issue. But since another PR is up now I'm more okay with approving this.

leekahung:
> Think it's good practice to address things like the responsivity issue. But since another PR is up now I'm more okay with approving this.

Don't think it's bad practice to have dedicated PRs for specific issues at hand, one with the main feature update, and then a follow-up to address bugs, unless the initial PR features breaks functionality. This is merging to the Development branch and not the Main/Master branch just yet where it's used for deployment.

PR #696, opened 2024-11-18
resize document columns for preview, share, and delete

### LINKED ISSUE #689: [Enhancement] - Resize Preview, Share, and Delete columns in Documents table

Currently, the Preview, Share, and Delete columns in the desktop version of Documents are wider than they need to be. As a result, less space is available for others such as Name and Description.

### Rationale:

It would be good for those to have more room so they're more immediately readable for the user.

### Proposed Changes:

Adjust the column sizes so the icon button ones are smaller and Name and Description are larger.

oddwili:
## This PR:
Resolves #689 
 
I updated the flex values to give preference to name and description. Preview, share, and delete have the smallest value

Screen Shot 2024-11-18 at 3 35 43 PM

### DIFF
--- a/src/components/Documents/DocumentsDesktop.jsx
+++ b/src/components/Documents/DocumentsDesktop.jsx
@@ -121,7 +121,6 @@ const DocumentsDesktop = ({ documents, handlers }) => {
       field: 'Delete',
       minWidth: 100,
       flex: 1,
-      flexShrink: 3,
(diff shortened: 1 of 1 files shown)


leekahung (approved):
Not much improvement when the screen width is below 1200-1300px, but I can see a big difference when it's used for a wider monitor when it prioritizes the earlier columns.

We can go with this.

PR #698, opened 2024-11-22
 Add notification for inaccessible documents

### LINKED ISSUE #694: [Enhancement] - Add notification for inaccessible data

User 1 gave User 2 access to one of their documents and later revoked access to it. However, User 2 still sees that document in their list. When User 2 then tries to access it, nothing appears to happen. With no feedback they might think something is broken.

### Rationale:

There is feedback provided in the console but not for the user themselves. This is neither accessible nor user-friendly design. The user ought to be informed when something does not work as they intended it.

### Proposed Changes:

There is already a custom hook called `useNotification.js`. Import that and alert the user whenever something is no longer accessible.

omarNaifer12:
## This PR:
Resolves #694 
**1.**    

screenShotNotification

I added a try-catch block to the handleShowDocumentLocal function in the DocumentTable component to handle the case where a user doesn't have access to a document. If a 403 error occurs (Unauthorized), the catch block triggers a notification informing the user they don't have permission to view the document.

@Jared-Krajewski @timbot1789 @xscottxbrownx @leekahung 

Please review the changes and let me know if there are any adjustments or improvements needed.

**1.**    

### DIFF
--- a/src/components/Documents/DocumentTable.jsx
+++ b/src/components/Documents/DocumentTable.jsx
@@ -9,7 +9,7 @@ import Box from '@mui/material/Box';
 import { useMediaQuery } from '@mui/material';
 // Context Imports
 import { DocumentListContext } from '@contexts';
-import { useSession } from '@hooks';
+import { useNotification, useSession } from '@hooks';
 // Utility Imports
 import { getBlobFromSolid } from '@utils';
 // Theme Imports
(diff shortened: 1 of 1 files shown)


andycwilliams:
This is working so far from my testing. However, I hesitate approving it just yet because it involves interaction with Solid and I have less experience on that aspect of the app. I'd like to wait for input from the development side and then can approve.

Thank you for your efforts!

omarNaifer12:
@andycwilliams  Thank you for your review! I’ll wait for the approval.

leekahung (approved):
This looks like a good try catch. Can approve for merging 👍

omarNaifer12:
@leekahung Yes, thank you. I was trying to find a way to do this without it since it's not used much in the project, but I couldn't find one because it gives an error and needs to be addressed.

PR #706, opened 2025-08-20
Make error messages user-friendly when adding contacts.

brianscherner:
## This PR:
Resolves #697

This PR updates the error handling in the Add Contact flow to ensure that users see clear, user-friendly error messages instead of raw exception output.
 
**1. Replaced raw exception messages with a getUserFriendlyError helper that maps common errors (401, 403, 404, network) into readable notifications for end users.**   

## Screenshots (if applicable):
issue-697-snip2
issue-697-snip

PR #707, opened 2025-08-22
Let's a user share their WebID via a QR code

### LINKED ISSUE #691: [Enhancement] - Add ability to generate QR code from WebId

Currently, if someone wants to provide their WebId to another, they have to tell them verbally or copy it to their clipboard and send it via another method.

We can add the ability to create a QR code for the user's WebId, which can then be scanned and shared with another person.

### Rationale:

Having a QR would be faster and simpler to provide WebId, than having to memorize the ID itself (especially when that ID may be long and cumbersome).

This ability is widespread by now and may even be expected by some users. It would be good to at least have this as an option.

### Proposed Changes:

Add a function on the user's Profile that will create the code.

Can check out free QR creation libraries, at least to start. One has not yet been decided upon.

brianscherner:
## This PR:
Resolves #691 
 
**1. Added a new button labeled "Show QR Code" on the user's Profile page, which opens a modal containing a scannable QR code of the user's webID using a library called _react-qr-code_. Their WebID is also displayed underneath the code for easy verification.**    

qr-code-btn
qr-code-modal

This has currently only been tested on a _localhost_ WebID but should work with other pods.

- Could add a _Download_ button so the QR code can be saved as a PNG for easy sharing (although they could just screenshot it as well - might not be necessary).
- Consider adding QR code support for other areas (ex: contact sharing).

### DIFF
--- a/src/components/Profile/CreateQRCode.jsx
+++ b/src/components/Profile/CreateQRCode.jsx
@@ -1,0 +1,49 @@
+import React, { useState } from 'react';
+import QRCode from 'react-qr-code';
+import QrCode2Icon from '@mui/icons-material/QrCode2';
+// import Box from '@mui/material/Box';
+// import ContentCopyIcon from '@mui/icons-material/ContentCopy';
+import IconButton from '@mui/material/IconButton';
+// import Typography from '@mui/material/Typography';
+// import useMediaQuery from '@mui/material/useMediaQuery';
+// import { useTheme } from '@mui/material/styles';
+
+/**
+ *
+ * @param root0
+ * @param root0.webId
+ */
+const CreateQRCode = ({ webId }) => {
+  console.log('WebId: ', webId);
+  const [qrIconClicked, setQrIconClicked] = useState(false);
+
+  const createQrCode = () => {
+    setQrIconClicked(!qrIconClicked);
+  };
+
+  return (
+    <IconButton
+      aria-label="Create QR Code"
+      edge="end"
+      onClick={() => {
+        createQrCode();
+      }}
+    >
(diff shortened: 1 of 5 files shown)


leekahung:
I see that this PR is based on #706.

I would merge the other PR first into development then rebase this branch to line up with Development.

Should be good to review after that.

src/components/Profile/CreateQRCode.jsx
```diff
@@ -0,0 +1,93 @@
...
+  Button,
+  Stack,
+  Typography
+} from '@mui/material';
+
+/**
+ *
+ * @param root0
+ * @param root0.webId
+ */
```
leekahung: Would update this to be a proper JSDoc comment, check other components for reference. The Modal components have very nice examples to follow.

leekahung:
The feature looks good, had just checked it on my local machine.

leekahung:
I've found a solution to the failing check in the CI/CD process, see #708. Think we can revert some of the changes you made to `ci.yml` here to avoid merge conflicts.

brianscherner:
I updated the component's documentation based on the examples from the other modal components and reverted the changes I had made to ci.yml.

I'm still getting a failed build check, but I think that's because this branch doesn't have the updated NPM package patches from #708 yet

leekahung:
> I'm still getting a failed build check, but I think that's because this branch doesn't have the updated NPM package patches from #708 yet

Yeah, let me merge the other PR. Think that when we update this branch with the updates from #708, we should pass the checks

leekahung (approved):
Sweet! I think we're in the clear.

